PR #3, opened 2023-03-02
docs: add a file with notes about codepush

eseidel:
For your review @felangel


eseidel:
Happy to put this other places, just wanted a place to write down all the known impls.

felangel (approved):

PR #26, opened 2023-03-06
feat(shorebird_cli): add build command

felangel:
## Description

- feat(shorebird_cli): add build command

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/build_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build_command.dart
@@ -75,12 +75,18 @@ class BuildCommand extends Command<int> {
     const executable = 'flutter';
     final arguments = [
       'build',
+      // This is temporary because the Shorebird engine currently
+      // only supports Android.
       'apk',
       '--release',
+      // This is temporary because the Shorebird engine currently
+      // does not support tree-shaking icons.
       '--no-tree-shake-icons',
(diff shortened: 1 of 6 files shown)


packages/shorebird_cli/lib/src/commands/build_command.dart
```diff
@@ -0,0 +1,102 @@
...
+    return ExitCode.success.code;
+  }
+
+  Future<void> _build(RunProcess runProcess, String shorebirdEnginePath) async {
+    const executable = 'flutter';
+    final arguments = [
+      'build',
+      'apk',
+      '--release',
+      '--no-tree-shake-icons',
```
eseidel, line 84: Should have a comment about why and that we're going to remove it.

packages/shorebird_cli/lib/src/commands/build_command.dart
```diff
@@ -0,0 +1,102 @@
...
+      '--local-engine-src-path',
+      shorebirdEnginePath,
+      '--local-engine',
+      'android_release_arm64',
```
eseidel, line 90: Probably also deserves a comment that this will need to change when we support more platform targets.

eseidel (approved):
Overall lgtm.

PR #27, opened 2023-03-06
chore: fix the rust build

eseidel:
This brings over all of the updater rust code from old_repo, except I deleted the "publisher" package which is now covered in the shorebird cli and I used the updated "shorebird_code_push_updater" package as "library".

I restored building with the rust and cdylib variants and kept the UPDATER_DEMO docs (we'll delete them soon but they still have some build instructions).  I edited down the README a bit since some of the instructions were old.

Once this is checked in, I'll work with felix to organize the code more aligned with his vision for the repo layout.


updater/dart_cli/pubspec.lock
```diff
@@ -0,0 +1,381 @@
+# Generated by pub
```
felangel, line 1: Thoughts on adding the `pubspec.lock` to the `.gitignore` since I believe the lock file is not taken into consideration when the CLI is activated via pub.

updater/UPDATER_DEMO.md
```diff
@@ -0,0 +1,128 @@
+To replicate the updater demo, you'll need a copy of the Flutter Engine.
```
felangel, line 1: Most of this should hopefully be able to go away very soon :)

felangel (approved):
Overall LGTM

updater/dart_cli/pubspec.lock
eseidel: I suspect it doesn't matter here much.  Technically this should commit today, but won't once we move it to be a library? https://dart.dev/guides/libraries/private-files

updater/UPDATER_DEMO.md
eseidel: Correct.

updater/dart_cli/pubspec.lock
felangel: I think Dart CLIs are treated like libraries.

> For library packages, don’t commit the pubspec.lock file. Regenerating the pubspec.lock file lets you test your package against the latest compatible versions of its dependencies.

Other CLIs like `pub` don't commit the lock file either https://github.com/dart-lang/pub but yeah it's not a big deal either way.

PR #29, opened 2023-03-06
feat(shorebird_cli): add run command

felangel:
## Description

- feat(shorebird_cli): add run command

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/run_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/run_command.dart
@@ -108,10 +108,13 @@ class RunCommand extends Command<int> {
       'flutter',
       [
         'run',
+        // Eventually we should support running in both debug and release mode.
         '--release',
         '--local-engine-src-path',
         shorebirdEnginePath,
         '--local-engine',
+        // This is temporary because the Shorebird engine currently
+        // only supports Android arm64.
         'android_release_arm64',
         if (argResults?.rest != null) ...argResults!.rest
       ],
--- a/packages/shorebird_cli/lib/src/commands/run_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/run_command.dart
@@ -55,7 +55,7 @@ class RunCommand extends Command<int> {
     }
 
     // This will likely change in the future as each Flutter application
-    // will not need to cache it's own copy of the Shorebird engine.
+    // will not need to cache its own copy of the Shorebird engine.
     final shorebirdEnginePath = p.join(
       Directory.current.path,
       '.shorebird',
--- a/packages/shorebird_cli/lib/src/commands/run_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/run_command.dart
@@ -1,0 +1,163 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:archive/archive_io.dart';
+import 'package:args/command_runner.dart';
+import 'package:mason_logger/mason_logger.dart';
+import 'package:path/path.dart' as p;
+import 'package:shorebird_cli/src/auth/auth.dart';
+import 'package:shorebird_cli/src/command_runner.dart';
+import 'package:shorebird_cli/src/compute.dart';
+import 'package:shorebird_code_push_api_client/shorebird_code_push_api_client.dart';
+
+typedef StartProcess = Future<Process> Function(
+  String executable,
+  List<String> arguments, {
+  bool runInShell,
+});
+
+/// {@template run_command}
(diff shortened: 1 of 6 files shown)


packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -0,0 +1,166 @@
...
+    if (session == null) {
+      _logger
+        ..err('You must be logged in to run.')
+        ..err("Run 'shorebird login' to log in and try again.");
+      return ExitCode.noUser.code;
+    }
+
+    final shorebirdEnginePath = p.join(
+      Directory.current.path,
+      '.shorebird',
```
eseidel, line 61: I might leave a comment that this ends up as part of some shorebird install eventually (each flutter checkout probably doesn't cache its own engine).

felangel: yup good call 👍

packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -0,0 +1,166 @@
...
+    return process.exitCode;
+  }
+}
+
+Future<void> _downloadShorebirdEngine(
+  ShorebirdCodePushApiClient codePushClient,
+  String path,
+) async {
+  final engine = await codePushClient.downloadEngine('latest');
+  final targetFile = File(path);
```
eseidel, line 142: It's pretty lame that dart doesn't have a "deleteIfExistsSync" or similar...

packages/shorebird_cli/lib/src/compute.dart
```diff
@@ -0,0 +1,41 @@
+import 'dart:async';
+import 'dart:isolate';
+
+/// Perform [computation] with [input] in an [Isolate].
+Future<R> compute<R, M>(FutureOr<R> Function(M) computation, M input) async {
```
eseidel, line 5: I thought there was some fancy new dart support for this?  I guess that's coming in Dart 3 and not yet in Dart 2.x.

packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -0,0 +1,166 @@
...
+  String archivePath,
+  String targetPath,
+) async {
+  final targetDirectory = Directory(targetPath);
+
+  if (targetDirectory.existsSync()) targetDirectory.deleteSync(recursive: true);
+
+  targetDirectory.createSync(recursive: true);
+
+  await compute(
```
eseidel, line 160: Could this function end up getting called twice and causing racing computes?  I wonder if we shouldn't just do this sync?

eseidel (approved):
lgtm, the compute stuff might be overkill?  Maybe that's how you do the fancy progress bars though?

packages/shorebird_cli/lib/src/compute.dart
felangel: Not that I'm aware of but it would be great to delete all this code. There's https://pub.dev/packages/compute but I didn't want to pull in another dependency just for this 🤷

packages/shorebird_cli/lib/src/commands/run_command.dart
felangel: If we do it sync then it'll block and the progress reporting will freeze up (and right now it takes me ~20s) which is long enough to where it might look like we locked up your shell 😛 

Lmk what you think.

felangel:
> lgtm, the compute stuff might be overkill? Maybe that's how you do the fancy progress bars though?

Yeah I added the compute so that the progress indicator keeps animating since the engine download takes ~20s on my network. Happy to simplify things though if you prefer to remove it and keep everything sync. Lmk what you prefer 👍

felangel:
Since you approved will merge the current changes but happy to revisit any of this in subsequent PRs 👍

eseidel:
At this point speed is the goal rather than correctness (or in this case bikeshed color), glad you biased towards action and just committed. :)  Yeah, we can always redo the compute stuff later if needed. 🤷

packages/shorebird_cli/lib/src/compute.dart
a14n: [Isolate.run](https://api.flutter.dev/flutter/dart-isolate/Isolate/run.html) has been [introduced in 2.19](https://api.flutter.dev/flutter/dart-isolate/Isolate/run.html).

PR #38, opened 2023-03-07
refactor(shorebird_code_push_api): use public shorebird engine

felangel:
## Description

- refactor(shorebird_code_push_api): use public shorebird engine

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_code_push_api/lib/src/middleware/http_client_provider.dart
```diff
@@ -1,12 +1,9 @@
...
 
-Middleware httpClientProvider(String key) {
-  return provider<Future<http.Client>>(
-    (_) async => _httpClient ??= createClient(key),
-  );
-}
+http.Client? _httpClient;
 
-Future<http.Client>? _httpClient;
+Middleware httpClientProvider() {
```
eseidel, line 7: Was the async removal automatic?  Feels a bit odd to leave the provider but remove the async.

eseidel (approved):

packages/shorebird_code_push_api/lib/src/middleware/http_client_provider.dart
felangel: it's because the `createHttpClient` from `pkg:googleapis_auth` was async but since we're no longer using it we can synchronously create a new http client if needed.

PR #47, opened 2023-03-08
refactor: rewrite the updater library into layers

eseidel:
## Description
I had to squash my commit history down, so this one commit has a lot:
* Moved to cbindgen instead of a manually created updater.h
* Moved to using shorebird_init to set config instead of passing each time.  Most symbols now take no arguments.
* Namspaced symbols with shorebird_
* Added more config options (and passed along to server) including channel, product_id, base_version, original_libapp_path, vm_path.
* Renamed "version" to "patch_version" and added "base_version"
* Split dart_bindings out from dart_cli (so it can be used from flutter apps).  dart_bindings is a placeholder name.
* Updated dart_bindings to the new apis.  dart_bindings should be generated by ffigen, but the setup looked awkward.
* Added a Updater.loadFlutterLibrary() path which tries to load symbols from the current process.  Doesn't work yet (example coming soon) due to linker commands in engine hiding all symbols, will fix.
* Simplified dart_bindings code now that it doesn't have to pass config each time.
* Fixed updater/library to compile on non-Android platforms.
* Split updater/library into many layers (hopefully for clarity)
* Documented AppParameters (cbindgen is fancy enough to forward the docs to updater.h too!)
* Changed internal names from "version" to "patch" where I could, since we believe "patches" will need to be based on a specific "version" to work.
* Made config (the params set by shorebird_init) thread-local and moved handling into config.rs

## Type of Change

- [X] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [X] 🧹 Code refactor
- [X] 📝 Documentation
- [ ] 🗑️ Chore


bootstrap.sh
```diff
@@ -0,0 +1,28 @@
...
+dart pub get
+cd ../..
+
+cd updater/dart_bindings
+dart pub get
+cd ../..
+
+cd updater/dart_cli
+dart pub get
+cd ../..
```
felangel, line 24: FYI

`very_good_cli` supports a recursive packages get so this could theoretically all be replaced with:

`very_good packages get -r ./packages`

eseidel: Happy to try it. 🤷

eseidel: Eventually we have to offer a solution for this problem to our customers (I think?).  Whether that's vending vg_cli or melos or whatever I don't know.

example/counter/pubspec.yaml
```diff
@@ -0,0 +1,91 @@
...
+# Read more about Android versioning at https://developer.android.com/studio/publish/versioning
+# In iOS, build-name is used as CFBundleShortVersionString while build-number is used as CFBundleVersion.
+# Read more about iOS versioning at
+# https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
+# In Windows, build-name is used as the major, minor, and patch parts
+# of the product and file versions while build-number is used as the build suffix.
+version: 1.0.0+1
+
+environment:
+  sdk: '>=3.0.0-266.0.dev <4.0.0'
```
felangel: nit but does this need to be 3.0.0 or can we use stable `>=2.19.0`?

example/counter/lib/main.dart
```diff
@@ -0,0 +1,130 @@
+import 'package:dart_bindings/updater.dart';
+import 'package:flutter/material.dart';
+
+void main() {
+  Updater.loadFlutterLibrary();
```
eseidel, line 5: These are the only two lines changed in the entire "counter" example, btw.

updater/library/include/updater.h
```diff
@@ -1,20 +1,104 @@
...
 
+/* Warning, this file is autogenerated by cbindgen. Don't modify this manually. */
+
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#ifdef _WIN32
+#define SHOREBIRD_EXPORT __declspec(dllexport)
```
eseidel: This was manually added (as part of my trying to export these symbols) and will go away next time I generate the header.

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
+// This file handles the global config for the updater library.
+
+use std::cell::RefCell;
+
+use crate::updater::AppConfig;
+
+const DEFAULT_BASE_URL: &'static str = "https://shorebird-code-push-api-cypqazu4da-uc.a.run.app";
```
eseidel, line 7: 'static just means the "borrow" lifetime here is static.  This is a borrowed string reference that lives as long as the program does.

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+const DEFAULT_CHANNEL: &'static str = "stable";
+
+thread_local!(static CONFIG: RefCell<Option<ResolvedConfig>> = RefCell::new(None));
```
eseidel, line 10: RefCell is just an object which knows how to track when it's borrowed.  So instead of copying out of this global I can borrow its contents and modify them.

updater/dart_bindings/lib/updater.dart
```diff
@@ -0,0 +1,117 @@
...
+  String? activePath() {
+    ffi.Pointer<Utf8> cVersion = ffi.Pointer<Utf8>.fromAddress(0);
+    try {
+      cVersion = bindings.activePath();
+      if (cVersion.address == 0) {
+        return null;
+      }
+      return cVersion.toDartString();
+    } finally {
+      // Can toDartString ever throw an exception, such that this finally
```
felangel: Looks like it could only throw if length is specified or if it's a null pointer

```dart
String toDartString({int? length}) {
  _ensureNotNullptr('toDartString');
  final codeUnits = cast<Uint8>();
  if (length != null) {
    RangeError.checkNotNegative(length, 'length');
  } else {
    length = _length(codeUnits);
  }
  return utf8.decode(codeUnits.asTypedList(length));
}
```

updater/dart_bindings/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: dart_bindings
+description: A starting point for Dart libraries or applications.
+version: 1.0.0
+# repository: https://github.com/my_org/my_repo
+
+environment:
+  sdk: '>=3.0.0-259.0.dev <4.0.0'
```
felangel: same comment here about whether we can use stable (>=2.19.0)

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+pub fn with_config<F, R>(f: F) -> R
+where
+    F: FnOnce(&ResolvedConfig) -> R,
```
eseidel, line 14: This is a type constraint.  Its saying that "F" is a kind of function/lambda (FnOnce, they're ephemeral) that takes this specific signature of a reference/borrow to a ResolveConfig and returns R (another type).  Where clauses like this are common for dealing with long types in the signature which may need to be repeated (as this one is).

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+{
+    CONFIG
+        .try_with(|config| {
```
eseidel, line 17: "with" or "try_with" is how you grab at thread locals in rust. 🤷   I'm using try_with here so I can have a custom crash message.  .expect says "if this was not OK, panic e.g. crash with this message".

updater/library/include/updater.h
```diff
@@ -1,20 +1,104 @@
 #ifndef updater_h
 #define updater_h
 
+/* Warning, this file is autogenerated by cbindgen. Don't modify this manually. */
```
felangel, line 4: 🚀

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+            let config = config.borrow();
+            let config = config
+                .as_ref()
+                .expect("Must call updater_init before using the updater library.");
```
eseidel, line 21: expect is like "unwrap" but with a message.  Unwrap takes the insides out of an Option<T> or Result<T> and hands you back the T.

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+fn resolve_config(config: AppConfig) -> ResolvedConfig {
+    // Resolve the config
+    // If there is no base_url, use the default.
+    // If there is no channel, use the default.
+    return ResolvedConfig {
+        client_id: config.client_id.to_string(),
+        base_url: config
+            .base_url
+            .as_deref()
```
eseidel, line 54: as_ref() above and as_deref() here are just ways of getting a flavor of the Option<T> as Option<&T>.  I never remember which to use, but thankfully copilot tells me. 🤣

updater/library/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+            .unwrap_or(DEFAULT_BASE_URL)
+            .to_owned(),
```
eseidel, line 56: to_owned() is the same as to_string() on str.  It's just a way of making a copy of a string reference for yourself (as a String).  to_owned is the name other versions of this "give me my own copy" work, to_string() just also happens to be a string version (give me my own String, representing this object) which happens to be the same on str.

example/counter/pubspec.yaml
eseidel: I just have the wrong Dart installed.  You and I should pick a version of flutter/dart to use consistently (e.g. stable).  But yes, this should be 2.19 to < 4.0.0

updater/dart_bindings/lib/updater.dart
eseidel: Ok, so I could remove that finally.

updater/library/src/updater.rs
```diff
@@ -1,23 +1,14 @@
-use std::collections::HashMap;
-use std::fmt::{Display, Formatter};
-use std::fs::File;
-use std::io::{BufReader, BufWriter, Write};
-use std::path::{Path, PathBuf};
-use std::string::ToString;
-
-use android_logger::Config;
-use log::LevelFilter;
+// This file's job is to be the Rust
```
felangel: Was this comment intentional? 😆

felangel (approved):
overall lgtm

example/counter/pubspec.yaml
eseidel: Fixed.

updater/dart_bindings/pubspec.yaml
eseidel: Fixed.

updater/library/src/updater.rs
eseidel: Fixed.

updater/dart_bindings/lib/updater.dart
eseidel: Refactored instead.

felangel (approved):

PR #54, opened 2023-03-10
feat: Teach the rust updater library about shorebird.yaml

eseidel:
The shorebird_init call now expects a string containing yaml contents in addition to the AppParameters struct.  AppParameters no longer contains product_id, client_id, update_url, or channel.  client_id is generated internally for the rust library and cached in the state cache.  product_id, base_url and channel are all read from shorebird.yaml which is passed in alongside the AppParameters.

## Type of Change

- [x] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore


updater/cli/src/main.rs
```diff
@@ -20,16 +20,17 @@ fn main() {
     let cli = Cli::parse();
 
     let config = updater::AppConfig {
-        client_id: "demo".to_string(),
```
felangel, line 23: Was it intentional to remove `client_id` completely?

updater/library/src/config.rs
```diff
@@ -59,27 +59,27 @@ impl ResolvedConfig {
...
         .to_owned();
     lock.cache_dir = config.cache_dir.to_string();
-    lock.client_id = config.client_id.to_string();
-    lock.product_id = config.product_id.to_string();
+    lock.product_id = yaml.product_id.to_string();
     lock.base_version = config.base_version.to_string();
     lock.original_libapp_path = config.original_libapp_path.to_string();
     lock.vm_path = config.vm_path.to_string();
     lock.is_initialized = true;
+    info!("Updater configured with: {:?}", lock);
```
felangel, line 84: do we want to keep this log in?

felangel (approved):
LGTM 🚀

updater/library/src/config.rs
eseidel: Yes, that one was intentional.  Felt good to see what config the updater actually got.

PR #60, opened 2023-03-13
refactor: remove package:shorebird_code_push_api_client

felangel:
## Description

- refactor: remove package:shorebird_code_push_api_client
  - the source code was moved to the private code_push_server in order to share models across the server and the client libraries.

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/pubspec.yaml
```diff
@@ -14,14 +14,16 @@ dependencies:
   checked_yaml: ^2.0.2
   cli_completion: ^0.3.0
   cli_util: ^0.4.0
+  code_push_server:
```
eseidel, line 17: Hmm.  I think we need these dependencies to work the other way (eventually at least).  It should be possible to use the public shorebird repos without shorebird_private (or other private repos), even if not all the functionality is there.

eseidel (approved):
lgtm, but we need to figure out how to keep the public shorebird repos compiling at least w/o access to the internal ones.

packages/shorebird_cli/pubspec.yaml
felangel: yeah open to ideas, I'm working on getting the CI to run properly

felangel:
> lgtm, but we need to figure out how to keep the public shorebird repos compiling at least w/o access to the internal ones.

Yeah this shouldn't be necessary if we move to using a proper DB on the backend and can make the repos public again...maybe I'll hold off on this for now even though it is annoying not being able to easily share the models with the current setup.

PR #77, opened 2023-03-15
chore: rename `product_id` to `app_id`

felangel:
## Description

- chore: rename product_id to app_id

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [X] 🗑️ Chore


packages/shorebird_code_push_protocol/test/src/models/registry_test.dart
```diff
@@ -44,14 +44,14 @@ void main() {
...
-              App(productId: 'app2'),
+              App(id: 'app2'),
             ],
           ),
           Account(
             apiKey: 'api_key2',
             apps: [
               App(
-                productId: 'app2',
+                id: 'app2',
```
eseidel: So this may not actually be the "id" in the database.  Since this is a per-account id, not a global-unique id.

eseidel (approved):

RodolfoSilva (approved):
LGTM!

eseidel:
Yeah, the only trouble you will have with this is what if you chose to store all apps in one table, you will need to use an globally unique id, rather than account-unique.  Which could just be account_id+app_id to make it unique. 🤷

felangel:
> Yeah, the only trouble you will have with this is what if you chose to store all apps in one table, you will need to use an globally unique id, rather than account-unique. Which could just be account_id+app_id to make it unique. 🤷

yeah good call, I suppose we can rename the field to `app_id` for clarity

PR #81, opened 2023-03-16
feat: Start adding rust tests.

eseidel:
We need to remove most/all of the unwrap() calls in the library since we should never crash (just fail to update and log).

Adding testing while I do this.

Updated updater::init and updater::report_failed_launch to now return Result<(), UpdaterError> instead () as before.
Added tempdir as a dev-only/test-only dependency.
Added UpdaterError so we have our own error type for our API, rather than exposing different error types for different functions.  I'm not sure if this is the right approach, but trying it.

Added two very simple tests which confirm that if you pass invalid yaml to init, or you don't have a current patch when calling report_failed_launch, we don't crash but we do return error (previously we would have crashed in both).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/updater/library/src/updater.rs
+++ b/updater/library/src/updater.rs
@@ -1,5 +1,6 @@
 // This file's job is to be the Rust API for the updater.
 
+use std::fmt;
 use std::fmt::{Display, Formatter};
 
 use crate::cache::{download_into_unused_slot, PatchInfo, UpdaterState};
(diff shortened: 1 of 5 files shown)


updater/library/src/updater.rs
```diff
@@ -1,5 +1,6 @@
 // This file's job is to be the Rust API for the updater.
 
+use std::fmt;
```
felangel, line 3: why is this necessary if we're using `std::fmt` below?

updater/library/src/updater.rs
```diff
@@ -130,3 +155,55 @@ pub fn update() -> UpdateStatus {
...
+            ),
+            Err(crate::UpdateError::InvalidArgument(
+                "yaml".to_string(),
+                "missing field `app_id`".to_string()
+            ))
+        );
+    }
+
+    #[test]
+    fn report_failure_with_no_current() {
```
felangel, line 200:
```suggestion
    fn report_failure_with_no_current_patch() {
```
nit 🤷

updater/library/src/updater.rs
eseidel: That exposes the 'fmt' name, which can then be used in fmt::Foo to get Foo from it.  I could have just added "Foo" to the line below and avoided this.  It's a nit I plan to clean up later.

eseidel: In rust you don't ever have to import anything, you can always just fully qualify the name.  So you can just always type std::fmt::Foo if you want, but often you instead "use std::fmt::Foo" to be able to just type "Foo". But similarly you can do "use std::fmt" and then be able to type "fmd::Foo" everywhere.

felangel (approved):
overall lgtm just left a few comments/questions

updater/library/src/updater.rs
eseidel: Will add.

PR #88, opened 2023-03-17
docs: add GLOSSARY.md

felangel:
## Description

- docs: add GLOSSARY.md

## Type of Change

- [ ] 🧹 Code refactor
- [X] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/GLOSSARY.md
+++ b/GLOSSARY.md
@@ -10,7 +10,7 @@ The application downloaded and run on various devices/platforms. All application
 
 **Example**
 
-The [`time_shift`](https://github.com/shorebirdtech/time_shift) app is an example of an application. See the `shorebird.yaml` for the `app_id`.
+The [`time_shift`](https://github.com/shorebirdtech/time_shift) app is an example of an application. The corresponding `app_id` can be found in the `shorebird.yaml` file at the root of the project.
 
 ### Release
 
@@ -36,7 +36,7 @@ If we had an existing `1.0.0` release of `time_shift` available, we could publis
 
 **Definition**
 
-A tag used to manage the subset of applications that receive a patch. By default, a "stable" channel is created and used when .
+A tag used to manage the subset of applications that receive a patch. By default, a "stable" channel is created and used by devices to query for available patches.
 
 **Example**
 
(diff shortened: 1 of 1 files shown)


GLOSSARY.md
```diff
@@ -0,0 +1,53 @@
+## Glossary
+
+Before getting started, we recommend familiarizing yourslef with the following terms and their definitions within the Shorebird ecosystem.
```
eseidel: yourself

eseidel: Although, honestly I'd remove this.  I'm not sure folks read these things first. 🤣  Some do I'm sure. 🤷

GLOSSARY.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Patch
+
+**Definition**
+
+An over the air update which is applied to a specific release.
+
+**Example**
+
+If we had an existing `1.0.0` release of `time_shift` available, we could publish a patch to `version 1.0.0`. All patches have a patch number (auto-incrementing integer) and multiple patches can be published for a given app release version.
```
eseidel: ... that would result in "patch 1", there can be many patches to a given release, typically the latest is active and applied to devices in the field.

GLOSSARY.md
```diff
@@ -0,0 +1,53 @@
...
+
+**Example**
+
+We can create a "development" and "staging" channels in addition to the "stable" channel and promote patches to those internal channels for testing before rolling them out to the general public.
+
+### Artifact
+
+**Definition**
+
+An artifact contains metadata about the contents of a specific patch for a specific platform and architecture.
```
eseidel, line 47: Oh, the Artifact is the blob + the metadata imo.

eseidel (approved):

GLOSSARY.md
felangel: I considered the download_url for the blob part of the artifact's metadata. Is that what you had in mind as well or did you imagine an artifact containing the blob itself? Maybe we need a different name for this 🤷

GLOSSARY.md
felangel: Yeah good catch just removed it for now as you suggested

GLOSSARY.md
felangel: Will merge this for now so we have a starting point but we can adjust this as needed in subsequent PRs 👍

PR #135, opened 2023-03-22
fix(shorebird_cli): init always prompts for display name

felangel:
## Description

- fix(shorebird_cli): init always prompts for display name

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


eseidel (approved):
I think this is fine for now, but I think you want it to always prompt *and* have a default?

e.g. what happens if you run init twice?

felangel:
> I think this is fine for now, but I think you want it to always prompt _and_ have a default?
> 
> e.g. what happens if you run init twice?

I updated to avoid creating an app if a `shorebird.yaml` already exists.

PR #163, opened 2023-03-24
feat: Add support for diff patches

eseidel:
This adds support for zstd-compressed patch files.

This reduces our update sizes by two orders of magnitude.  In testing changing the default flutter counter app from "blue" to "red' background it was a 3mb update before, now it's a 31k update.  That's still larger than needed, and I think we can go smaller, but this is good enough for now.

I ended up needing to change the engine code as well (it was sending the wrong path for the base libapp.so), so I also changed the signature of shorebird_active_version to shorebird_active_patch_number to match our modern wording.  It's still a char* for now.

The compression comes from zstd, through comde (compression/decompression) which is a simple wrapper library around a few rust compression libraries.  We could drop the intermediate layer once we're sure which compression algorithm we're using.

Fixed dont to do_not to appease cspell.

Added a new (optional) bool in the server Patch response "is_diff" which controls whether this behavior is turned on or not.

I've tested by manually submitting my own diff that this flow does work correctly, but we'll need to change the cli and server to support it by default.

Fixed a couple places we were using PathBuf instead of Path.  Turns out PathBuf is to String where Path& is to str&, so using Path& instead of PathBuf& where possible.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/updater/library/include/updater.h
+++ b/updater/library/include/updater.h
@@ -59,13 +59,13 @@ void shorebird_init(const struct AppParameters *c_params,
                     const char *c_yaml);
 
 /**
- * Return the active version of the app, or NULL if there is no active version.
+ * Return the active patch number, or NULL if there is no active patch.
  */
 SHOREBIRD_EXPORT char *shorebird_active_patch_number(void);
 
 /**
- * Return the path to the active version of the app, or NULL if there is no
- * active version.
+ * Return the path to the active patch for the app, or NULL if there is no
+ * active patch.
  */
 SHOREBIRD_EXPORT char *shorebird_active_path(void);
 
--- a/updater/library/include/updater.h
+++ b/updater/library/include/updater.h
@@ -61,7 +61,7 @@ void shorebird_init(const struct AppParameters *c_params,
 /**
  * Return the active version of the app, or NULL if there is no active version.
  */
-SHOREBIRD_EXPORT char *shorebird_active_version(void);
+SHOREBIRD_EXPORT char *shorebird_active_patch_number(void);
 
 /**
  * Return the path to the active version of the app, or NULL if there is no
@@ -84,6 +84,11 @@ SHOREBIRD_EXPORT bool shorebird_check_for_update(void);
  */
 SHOREBIRD_EXPORT void shorebird_update(void);
 
+/**
+ * Report that the app failed to launch.  This will cause the updater to
+ * attempt to roll back to the previous version if this version has not
+ * been launched successfully before.
+ */
 SHOREBIRD_EXPORT void shorebird_report_failed_launch(void);
 
 #ifdef __cplusplus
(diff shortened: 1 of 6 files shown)


updater/library/include/updater.h
```diff
@@ -61,7 +61,7 @@ void shorebird_init(const struct AppParameters *c_params,
 /**
  * Return the active version of the app, or NULL if there is no active version.
  */
-SHOREBIRD_EXPORT char *shorebird_active_version(void);
+SHOREBIRD_EXPORT char *shorebird_active_patch_number(void);
```
felangel, line 64: 💯

updater/library/include/updater.h
```diff
@@ -61,7 +61,7 @@ void shorebird_init(const struct AppParameters *c_params,
 /**
  * Return the active version of the app, or NULL if there is no active version.
```
felangel:
```suggestion
 * Return the active patch number of the app, or NULL if there is no active patch number.
```

updater/library/src/c_api.rs
```diff
@@ -67,9 +67,8 @@ pub extern "C" fn shorebird_init(c_params: *const AppParameters, c_yaml: *const
 }
 
 /// Return the active version of the app, or NULL if there is no active version.
```
felangel:
```suggestion
/// Return the active patch number of the app, or NULL if there is no active patch number.
```

updater/library/src/updater.rs
```diff
@@ -125,6 +134,39 @@ fn update_internal(config: &ResolvedConfig) -> anyhow::Result<UpdateStatus> {
...
+
+    let (patch_r, patch_w) = pipe::pipe();
+
+    let decompress = ZstdDecompressor::new();
+    std::thread::spawn(move || {
+        let result = decompress.copy(compressed_patch_r, patch_w);
+        // If this thread fails, undoubtedly the main thread will fail too.
+        // Most important is to not crash.
+        if let Err(err) = result {
+            error!("Decompression thread failed: {err}");
```
felangel, line 164: Rust noob question:
What does this `error!(...)` do? Does it propagate as the returned `Result` or are we swallowing it and always returning `Ok`?

eseidel:
Also requires https://github.com/shorebirdtech/engine/pull/4

felangel (approved):
overall lgtm just had some rust-specific questions

updater/library/src/updater.rs
eseidel: error! just logs with the "error" tag.

spawn() takes a closure here, the "||", which does not return anything (if it did it would have a -> Type saying what it returned).  This spawns a new, short-lived thread, which just runs until the decompression is done.

eseidel:
Updated the patch with answers to your questions.  Thank you!

eseidel:
Thank you to @bbqsrc for providing the helpful rust library!

PR #179, opened 2023-03-27
feat: Add patch command for use by cli

eseidel:
I created a `patch` command that knows how to create a patch from a base and new libapp.so.  This uses the same rust crate the updater library does to minimize compatibility risk.

Also improved build_engine.sh, to also build the rust side, both the updater library and the new patch command.

build_engine.sh now errors in output is relative (previously it just dumped output in temp dir silently. :joy:)

I made the rust side build for both arm64 and arm32 even though we don't yet include the arm32 bits in the build.

I also fixed the updater build to support arm32/armv7 (currently called "arm" to the server).

Added --no-goma flag to gn to silence the later warning from ninja that goma was configured but not present.  GOMA is the (google-internal?) compile service Chrome uses.

Included the `patch` binary in the engine.zip for now.  It belongs in some separate artifact system eventually, but this is good enough for now.

Removed dart2wasm_product.snapshot for now since it's not present in our current build.

Made build_engine exit on any error with -e, previously it would just continue happily.

### DIFF
--- a/scripts/build_engine.sh
+++ b/scripts/build_engine.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
 # Given a path to a Flutter engine, generate a shorebird version of it.
 # This is done by removing all the files that are not needed for the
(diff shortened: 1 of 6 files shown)


scripts/build_engine.sh
```diff
@@ -8,21 +8,67 @@
...
 
 # Build the the host_release output.
-./src/flutter/tools/gn --runtime-mode=release
+./src/flutter/tools/gn --runtime-mode=release --no-goma
 ninja -C ./src/out/host_release
+# Arm64 version for when we need it:
+# ./src/flutter/tools/gn --runtime-mode=release --mac-cpu=arm64
+# ninja -C ./src/out/host_release_arm64
+
+# Hack around host_release builds being broken on 3.7.8 stable.
```
felangel, line 69: I thought they were working on 3.7.8

eseidel: Yeah, sorry will remove that bit.

felangel (approved):

PR #195, opened 2023-03-29
fix(shorebird_cli): upgrade use origin/stable as upstream

felangel:
## Description

- fix(shorebird_cli): upgrade use origin/stable as upstream

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/upgrade_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/upgrade_command.dart
@@ -83,8 +83,8 @@ class UpgradeCommand extends ShorebirdCommand {
       ['fetch', '--tags'],
       workingDirectory: workingDirectory,
     );
-    // Get the latest commit revision of the upstream
-    return _gitRevParse('@{upstream}', workingDirectory: workingDirectory);
+    // Get the latest commit revision of the stable branch
+    return _gitRevParse('origin/stable', workingDirectory: workingDirectory);
   }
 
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/upgrade_command.dart
```diff
@@ -83,8 +83,8 @@ class UpgradeCommand extends ShorebirdCommand {
       ['fetch', '--tags'],
       workingDirectory: workingDirectory,
     );
-    // Get the latest commit revision of the upstream
-    return _gitRevParse('@{upstream}', workingDirectory: workingDirectory);
+    // Get the latest commit revision of the stable branch
+    return _gitRevParse('origin/stable', workingDirectory: workingDirectory);
```
eseidel, line 87: This won't be right once we have channels.  Are you sure you want this?

eseidel (approved):

packages/shorebird_cli/lib/src/commands/upgrade_command.dart
felangel: It's currently breaking because `@{upstream}` doesn't point to a branch

eseidel: Go for it then.  We can always change it later. :)

PR #198, opened 2023-03-29
feat(shorebird_cli): `shorebird init` creates app when `shorebird.yaml` exists

felangel:
## Description

- feat(shorebird_cli): `shorebird init` creates app when `shorebird.yaml` exists

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -72,10 +72,7 @@ Please make sure you are running "shorebird init" from the root of your Flutter 
         return ExitCode.software.code;
       }
 
-      final app = apps.firstWhereOrNull(
-        (a) => a.id == getShorebirdYaml()!.appId,
-      );
-
+      final app = apps.firstWhereOrNull((a) => a.id == shorebirdYaml!.appId);
       appId = app?.id;
     }
 
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -44,7 +44,7 @@ Please make sure you are running "shorebird init" from the root of your Flutter 
       return ExitCode.software.code;
     }
 
-    late final ShorebirdYaml? shorebirdYaml;
+    final ShorebirdYaml? shorebirdYaml;
     try {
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -41,35 +44,59 @@ Please make sure you are running "shorebird init" from the root of your Flutter
       return ExitCode.software.code;
     }
 
-    late final bool shorebirdYamlExists;
+    late final ShorebirdYaml? shorebirdYaml;
```
bryanoltman: Does this need to be late since it's already optional?

packages/shorebird_cli/test/src/commands/init_command_test.dart
```diff
@@ -134,24 +138,64 @@ Please make sure you are running "shorebird init" from the root of your Flutter
...
+      ).writeAsStringSync('app_id: $appId');
+      final exitCode = await IOOverrides.runZoned(
+        command.run,
+        getCurrentDirectory: () => tempDir,
+      );
+      verify(() => progress.fail(error)).called(1);
+      expect(exitCode, ExitCode.software.code);
+    });
+
+    test('detects existing shorebird.yaml (existing app)', () async {
```
bryanoltman: Not blocking, but putting this test and the following in a `group('existing Shorebird app')` might make the test names a bit cleaner.

bryanoltman (approved):
I've been here for an hour and am not confident in my review of the domain logic quite yet, but otherwise looks good!

packages/shorebird_cli/lib/src/commands/init_command.dart
felangel: Yup good call it's not needed 👍

packages/shorebird_cli/test/src/commands/init_command_test.dart
felangel: I adjusted the test names to be a bit clearer lmk what you think. One of the two tests is for an existing shorebird app and the other is for a non-existing shorebird app.

bryanoltman (approved):

Issue #110, opened 2023-03-21
feat: Channels (some promotion mechanism)


felangel:
## Problem

As a developer, I want to be able to publish a patch to a sub-set of customers to validate the patch is applied correctly before promoting the patch to my production user base.

## Assumptions

Developers use existing test tracks to promote releases within the Play Store/TestFlight etc. so a single release artifact is propagated through multiple "test tracks" before ultimately landing in production. In addition, we've observed that of existing customers that are requesting this, all have applications that are already using flavors.

## Proposal

This can be achieved today without any changes to shorebird by using flavors.

Create a release for each flavor:

```
shorebird release --flavor dev
shorebird release --flavor stg
shorebird release --flavor prd
```

Upload the artifacts to the store and roll them out to your internal testing tracks:

<img width=500 src='https://github.com/shorebirdtech/shorebird/assets/8855632/ea138d52-3ffc-4956-a3bd-efa06acf13a6'/>

Once you've rolled out to production, if you find a bug and want to test a patch, publish the patch to the dev variant:

`shorebird patch --flavor dev`

<img width=500 src='https://github.com/shorebirdtech/shorebird/assets/8855632/12f1cc22-5c2f-4934-a4f0-2e6cde1491ec'/>

If you're happy with the dev patch you can apply the patch to staging:

`shorebird patch --flavor stg`

<img width=500 src='https://github.com/shorebirdtech/shorebird/assets/8855632/5afbe08d-31bd-4520-80fc-a0a800310c15'/>

Finally when the patch has been verified in staging you can push to production:

`shorebird patch --flavor prd`

<img width=500 src='https://github.com/shorebirdtech/shorebird/assets/8855632/cfb59d4f-54d8-4d95-bbc8-1a3d46288b85'/>

felangel:
Closing this for now since it seems the proposal addresses the needs of our current customers.

PR #201, opened 2023-03-29
feat(shorebird_cli): `shorebird channels list`

felangel:
## Description

- feat(shorebird_cli): `shorebird channels list` (part of #110)

See all available channels for your application using the `shorebird channels list` command:

```bash
shorebird channels list
```

**Sample**

```
shorebird channels list
📱 App ID: 61fc9c16-3c4a-4825-a155-9765993614aa
📺 Channels
┌─────────────┐
│ Name        │
├─────────────┤
│ stable      │
├─────────────┤
│ development │
└─────────────┘
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/channels/list_channels_command.dart
```diff
@@ -0,0 +1,114 @@
...
+/// {@endtemplate}
+class ListChannelsCommand extends ShorebirdCommand with ShorebirdConfigMixin {
+  /// {@macro list_channels_command}
+  ListChannelsCommand({
+    required super.logger,
+    super.buildCodePushClient,
+    super.auth,
+  }) {
+    argParser.addOption(
+      'app-id',
```
bryanoltman: It might be good to pull the `app-id` string out into a constant.

packages/shorebird_cli/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ environment:
 dependencies:
   archive: ^3.3.6
   args: ^2.3.1
+  barbecue: ^0.4.0
```
eseidel, line 14: Reminds me that we will eventually care about providence of our dependencies.  Not yet, but as an SDK that will produce binaries that run on millions (billions!) of devices we will need to care about our supply chain so we're not an eventual EventStream/Solarwinds.

packages/shorebird_cli/lib/src/commands/channels/list_channels_command.dart
```diff
@@ -0,0 +1,114 @@
...
+      logger.err('You must be logged in to view channels.');
+      return ExitCode.noUser.code;
+    }
+
+    final client = buildCodePushClient(
+      apiKey: session.apiKey,
+      hostedUri: hostedUri,
+    );
+
+    final appId = results['app-id'] as String? ?? getShorebirdYaml()?.appId;
```
bryanoltman: …and reference it here

bryanoltman (approved):

packages/shorebird_cli/README.md
```diff
@@ -239,6 +239,29 @@ Build a new release of your application using the `shorebird build` command:
...
+**Sample**
+
+```
+shorebird channels list
+📱 App ID: 61fc9c16-3c4a-4825-a155-9765993614aa
+📺 Channels
+┌────┬─────────────┐
+│ ID │ Name        │
+├────┼─────────────┤
+│ 0  │ stable      │
```
eseidel: Why does the user need to care about the id?  What if the names are unique?

eseidel (approved):

packages/shorebird_cli/pubspec.yaml
felangel: Yup it's Apache 2.0 but we can always roll our own or use something else.

packages/shorebird_cli/README.md
felangel: We could omit the id and do the lookup based on (name, app-id) 👍

packages/shorebird_cli/lib/src/commands/channels/list_channels_command.dart
felangel: good call, done 👍

PR #204, opened 2023-03-30
chore(shorebird_cli): adjust shorebird engine revision

felangel:
## Description

- chore(shorebird_cli): adjust shorebird engine revision

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [X] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/engine_revision.dart
+++ b/packages/shorebird_cli/lib/src/engine_revision.dart
@@ -3,4 +3,4 @@
 const requiredFlutterEngineRevision = '9aa7816315';
 
 /// The hash of the Shorebird engine used by this version of Shorebird CLI.
-const shorebirdEngineRevision = 'b6e8c4ecd973b39b7ca4fe71ca6a12b7bd361814';
+const shorebirdEngineRevision = 'b6e8c4ecd9';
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/engine_revision.dart
```diff
@@ -3,4 +3,4 @@
 const requiredFlutterEngineRevision = '9aa7816315';
 
 /// The hash of the Shorebird engine used by this version of Shorebird CLI.
-const shorebirdEngineRevision = 'b6e8c4ecd973b39b7ca4fe71ca6a12b7bd361814';
+const shorebirdEngineRevision = 'b6e8c4ecd9';
```
eseidel, line 6: 🤷  This feels like the kind of thing that you want the computers to check precisely even if the humans see imprecisely.

eseidel (approved):

packages/shorebird_cli/lib/src/engine_revision.dart
felangel: Yeah out of curiosity I looked and apparently 10 characters is plenty to uniquely identify the revision 🤷 

> Generally, eight to ten characters are more than enough to be unique within a project. One of the largest Git projects, the Linux kernel, is beginning to need 12 characters out of the possible 40 to stay unique.

https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1

We can always go back to using the full (especially once we create the new engine repo with submodules for updater and engine) 👍

PR #206, opened 2023-03-30
fix(updater): use all base library paths

felangel:
## Description

- fix(updater): use all base library paths rather than assuming the last path is the correct one

Looks like we have an issue because we're incorrectly assuming the last path in `settings.application_library_path` is the correct path when the docs say:

```
// Path to a library containing the application's compiled Dart code.
// This is a vector so that the embedder can provide fallback paths in
// case the primary path to the library can not be loaded.
```

This change passes the entire list of paths to the rust updater and the rust updater uses the first path that exists.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/updater/dart_bindings/lib/src/bindings.dart
+++ b/updater/dart_bindings/lib/src/bindings.dart
@@ -16,7 +16,7 @@ class AppParameters extends ffi.Struct {
   // ignore: non_constant_identifier_names
   external ffi.Pointer<Utf8> update_url;
   // ignore: non_constant_identifier_names
-  external ffi.Pointer<ffi.Pointer<Utf8>> original_libapp_path;
+  external ffi.Pointer<ffi.Pointer<Utf8>> original_libapp_paths;
   // ignore: non_constant_identifier_names
   external ffi.Pointer<Utf8> vm_path;
   // ignore: non_constant_identifier_names
@@ -38,9 +38,9 @@ class AppParameters extends ffi.Struct {
     if (updateUrl != null) {
       config.ref.update_url = updateUrl.toNativeUtf8();
     }
-    config.ref.original_libapp_path = calloc<ffi.Pointer<Utf8>>();
+    config.ref.original_libapp_paths = calloc<ffi.Pointer<Utf8>>();
     for (var i = 0; i < libappPaths.length; i++) {
-      config.ref.original_libapp_path[i] = libappPaths[i].toNativeUtf8();
+      config.ref.original_libapp_paths[i] = libappPaths[i].toNativeUtf8();
     }
     config.ref.vm_path = libflutterPath.toNativeUtf8();
     config.ref.cache_dir = cacheDir.toNativeUtf8();
@@ -54,11 +54,11 @@ class AppParameters extends ffi.Struct {
     calloc.free(config.ref.update_url);
     // Free all paths in original_libapp_path.
     var i = 0;
-    while (config.ref.original_libapp_path[i].address != 0) {
-      calloc.free(config.ref.original_libapp_path[i]);
+    while (config.ref.original_libapp_paths[i].address != 0) {
+      calloc.free(config.ref.original_libapp_paths[i]);
       i++;
     }
-    calloc.free(config.ref.original_libapp_path);
+    calloc.free(config.ref.original_libapp_paths);
     calloc.free(config.ref.vm_path);
     calloc.free(config.ref.cache_dir);
     calloc.free(config);
--- a/updater/dart_bindings/lib/src/bindings.dart
+++ b/updater/dart_bindings/lib/src/bindings.dart
@@ -42,7 +42,7 @@ class AppParameters extends ffi.Struct {
       config.ref.update_url = updateUrl.toNativeUtf8();
     }
     config.ref.original_libapp_paths = calloc<ffi.Pointer<Utf8>>(
-      libappPaths.length + 1, // +1 for the null terminator.
+      libappPaths.length,
     );
 
     for (var i = 0; i < libappPaths.length; i++) {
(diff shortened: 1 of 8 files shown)


updater/dart_bindings/lib/src/bindings.dart
```diff
@@ -49,6 +52,12 @@ class AppParameters extends ffi.Struct {
     calloc.free(config.ref.base_version);
     calloc.free(config.ref.channel);
     calloc.free(config.ref.update_url);
+    // Free all paths in original_libapp_path.
+    var i = 0;
+    while (config.ref.original_libapp_path[i].address != 0) {
```
bryanoltman: Is it guaranteed that the last entry will be 0?

updater/library/src/updater.rs
```diff
@@ -145,8 +145,7 @@ fn update_internal(config: &ResolvedConfig) -> anyhow::Result<UpdateStatus> {
     let download_path = download_dir.join(patch.number.to_string());
     download_to_path(&patch.download_url, &download_path)?;
 
-    // Inflate the patch from a diff.
-    let base_path = PathBuf::from(&config.original_libapp_path);
+    let base_path = get_base_path(&config.original_libapp_paths)?;
```
bryanoltman, line 148: Do we need to check whether this result is Err?

bryanoltman (commented):
Seems fine to me, but I'm also still getting up to speed and am not well-versed in Rust. Comments are mostly for my learning.

updater/dart_bindings/lib/src/bindings.dart
felangel: Was reading about it and it looks like lists are null-terminated and the way to determine this is by checking `pointer.address == 0`.

updater/dart_bindings/lib/src/bindings.dart
```diff
@@ -38,7 +38,10 @@ class AppParameters extends ffi.Struct {
     if (updateUrl != null) {
       config.ref.update_url = updateUrl.toNativeUtf8();
     }
-    config.ref.original_libapp_path = libappPath.toNativeUtf8();
+    config.ref.original_libapp_paths = calloc<ffi.Pointer<Utf8>>();
```
eseidel: This seems wrong?  It looks like it's allocating a pointer to utf8.  It needs to allocate an array of pointers (at a fixed size), no?

updater/dart_bindings/lib/src/bindings.dart
```diff
@@ -49,7 +52,13 @@ class AppParameters extends ffi.Struct {
     calloc.free(config.ref.base_version);
     calloc.free(config.ref.channel);
     calloc.free(config.ref.update_url);
-    calloc.free(config.ref.original_libapp_path);
+    // Free all paths in original_libapp_path.
+    var i = 0;
+    while (config.ref.original_libapp_paths[i].address != 0) {
```
eseidel: There are various ways to represent lists in C.  It looks like you're assuming that they're an array that's N+1 in length and the last entry is 0?

updater/library/src/updater.rs
felangel: My understanding is by using ? here we propagate the error up to the caller if base_path is an error.

eseidel: ? propagates the Err if needed.

updater/dart_bindings/lib/src/bindings.dart
eseidel: I think this is the only real problem in this patch. I think this code is writing out of bounds and we're just not catching it because we're not using libgmalloc or similar.

eseidel (changes requested):

updater/dart_bindings/lib/src/bindings.dart
eseidel: I think if you passed the size + 1 it would "just work".
https://api.dart.dev/stable/2.19.4/dart-ffi/Allocator/allocate.html

eseidel:
The Dart side is still wrong in the one place noted.

updater/dart_bindings/lib/src/bindings.dart
felangel: I think I addressed this in the latest commit, lmk what you think.

updater/dart_bindings/lib/src/bindings.dart
```diff
@@ -38,7 +41,13 @@ class AppParameters extends ffi.Struct {
     if (updateUrl != null) {
       config.ref.update_url = updateUrl.toNativeUtf8();
     }
-    config.ref.original_libapp_path = libappPath.toNativeUtf8();
+    config.ref.original_libapp_paths = calloc<ffi.Pointer<Utf8>>(
+      libappPaths.length + 1, // +1 for the null terminator.
```
eseidel: You don't need the null terminator anymore with your new method of passing the size.

updater/dart_bindings/lib/src/bindings.dart
```diff
@@ -49,7 +58,14 @@ class AppParameters extends ffi.Struct {
     calloc.free(config.ref.base_version);
     calloc.free(config.ref.channel);
     calloc.free(config.ref.update_url);
-    calloc.free(config.ref.original_libapp_path);
+    // Free all paths in original_libapp_path.
+    var i = 0;
+    for (var i = 0; i < config.ref.original_libapp_paths_size; i++) {}
```
eseidel: ?

updater/dart_bindings/lib/src/bindings.dart
felangel: oh yeah good point 👍

updater/library/src/c_api.rs
```diff
@@ -38,13 +40,30 @@ fn to_rust(c_string: *const libc::c_char) -> String {
     unsafe { CStr::from_ptr(c_string).to_str().unwrap() }.to_string()
 }
 
+fn to_rust_vector(c_array: *const *const libc::c_char, size: libc::c_int) -> Vec<String> {
+    let mut result = Vec::new();
+    let mut i = 0;
+    loop {
+        if i >= size {
```
eseidel: I think a for should work. :)

updater/dart_bindings/lib/src/bindings.dart
felangel: Ah yeah good call.

updater/dart_bindings/lib/src/bindings.dart
felangel: Oops must've accidentally changed this, will fix 👍

eseidel (approved):

PR #207, opened 2023-03-31
feat(shorebird_cli): add `--dry-run` and `--force` to `shorebird patch`

felangel:
## Description

- feat(shorebird_cli): add --dry-run and --force to patch

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch_command.dart
@@ -153,7 +153,7 @@ class PatchCommand extends ShorebirdCommand
     final version = pubspecYaml.version!;
     final versionString = '${version.major}.${version.minor}.${version.patch}';
 
-    late final List<App> apps;
+    final List<App> apps;
     final fetchAppsProgress = logger.progress('Fetching apps');
     try {
       apps = (await codePushClient.getApps())
@@ -222,7 +222,7 @@ ${styleBold.wrap(lightGreen.wrap('🚀 Ready to publish a new patch!'))}
       }
     }
 
-    late final List<Release> releases;
+    final List<Release> releases;
     final fetchReleasesProgress = logger.progress('Fetching releases');
     try {
       releases = await codePushClient.getReleases(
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/patch_command.dart
```diff
@@ -94,6 +106,14 @@ class PatchCommand extends ShorebirdCommand
       return ExitCode.software.code;
     }
 
+    final force = results['force'] == true;
```
eseidel, line 109: Looks fine, but isn't this also solved by adding default: false to the addFlag?

eseidel: Then it's just results['force']?  Donno which is clearer. 🤷

packages/shorebird_cli/test/src/commands/patch_command_test.dart
```diff
@@ -433,9 +452,6 @@ Please create a release using "shorebird release" and try again.
 
     test('throws error when release artifact cannot be retrieved.', () async {
       const error = 'something went wrong';
-      when(
```
eseidel, line 436: Why are these no longer needed?

eseidel (approved):

bryanoltman (approved):

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: I tried this – `results['force']` is still null if no arg is provided.

felangel: `default: false` is the default but accessing arg flags is `dynamic` which is why I wrote it this way.

packages/shorebird_cli/test/src/commands/patch_command_test.dart
felangel: They are the default stubs so they were unnecessary

PR #218, opened 2023-04-03
feat(updater): use `try_exists` instead of `exists`

felangel:
## Description

> As opposed to the `exists()` method, this one doesn't silently ignore errors unrelated to the path not existing. (E.g. it will return `Err(_)` in case of permission denied on some of the parent directories.)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/updater/library/src/updater.rs
+++ b/updater/library/src/updater.rs
@@ -172,8 +172,18 @@ fn get_base_path(original_lib_app_paths: &Vec<String>) -> anyhow::Result<PathBuf
     // Iterate through the paths and find the first one that exists.
     for path in original_lib_app_paths {
         let path = PathBuf::from(path);
-        if path.try_exists()? {
-            return Ok(path);
+        match path.try_exists() {
+            Ok(true) => {
+                return Ok(path);
+            }
+            Ok(false) => {
+                info!("File does not exist: {:?}", path);
+                continue;
+            }
+            Err(err) => {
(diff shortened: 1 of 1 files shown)


updater/library/src/updater.rs
```diff
@@ -172,7 +172,7 @@ fn get_base_path(original_lib_app_paths: &Vec<String>) -> anyhow::Result<PathBuf
     // Iterate through the paths and find the first one that exists.
     for path in original_lib_app_paths {
         let path = PathBuf::from(path);
-        if path.exists() {
+        if path.try_exists()? {
```
eseidel: This changes the behavior of the method so that if there is an earlier path with permissions errors we will now error out of get_base_path, rather than previously we would have returned a valid later path.

Is that what you want?

eseidel: https://doc.rust-lang.org/std/fs/fn.try_exists.html

eseidel (approved):
Anyway, feel free to land.  You just might want to document the behavior change.

updater/library/src/updater.rs
felangel: yeah good point I'll adjust the behavior

felangel: @eseidel I adjusted the behavior lmk what you think, thanks!

eseidel (approved):

Issue #133, opened 2023-03-22
Move to OAuth


eseidel:
Goals here are two:
* Make self-signup possible (which is more than just auth)
* Make non-API-key auth possible (so we can safely add superuser controls, we can move towards supporting teams, etc.)

3p options include:
* https://clerk.com/startups
* https://firebase.google.com/docs/auth

eseidel:
#215 and #216 are related.

PR #223, opened 2023-04-04
feat(shorebird_cli): use Google OAuth 2.0

felangel:
## Description

- feat: use Google OAuth 2.0 on Shorebird CLI (part of #133)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -1,54 +1,104 @@
...
+export 'package:googleapis_auth/googleapis_auth.dart' show AccessCredentials;
+
+final _clientId = ClientId(
+  /// Shorebird CLI's OAuth 2.0 identifier.
+  '30552215580-6kgm623s9l4te0rd63r18c5u6b6gkts5.apps.googleusercontent.com',
+
+  /// Shorebird CLI's OAuth 2.0 secret.
+  ///
+  /// This isn't actually meant to be kept secret.
+  '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',
```
eseidel: I'm confused.  Is this not supposed to be secret? https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/

eseidel: Looks like the "secret" is optional? https://pub.dev/documentation/googleapis_auth/latest/googleapis_auth/ClientId-class.html

eseidel: https://docs.gitguardian.com/secrets-detection/detectors/specifics/google_oauth2_keys

felangel: > The Client ID must be created with a client_secret here, however there is no way to properly secure a client_secret for installed/console applications. Fortunately the OAuth2 flow used in this case [assumes that the app cannot keep secrets](https://developers.google.com/identity/protocols/oauth2/native-app) so this particular client_secret does not need to be kept secret. You should however make sure not to re-use the same client_secret anywhere secrecy is required.

https://pub.dev/packages/googleapis_auth 

Without a client secret we would get back:

```
Failed to obtain access credentials. Error: invalid_request client_secret is missing. Status code: 400
```

`pub` follows the same approach: https://github.com/dart-lang/pub/blob/196e89e5716d64b2954021e6ac9bb5a20d975fc6/lib/src/oauth2.dart#L37

packages/shorebird_cli/lib/src/commands/login_command.dart
```diff
@@ -18,32 +18,37 @@ class LoginCommand extends ShorebirdCommand {
...
     } catch (error) {
-      loginProgress.fail();
       logger.err(error.toString());
       return ExitCode.software.code;
     }
   }
+
+  void _prompt(String url) {
+    logger.info('''
+Shorebird needs your authorization to manage apps, releases, and patches on your behalf.
```
eseidel: Is "Shorebird" here the command or the service or both?

eseidel (approved):
Other than the client_secret question lgtm.

packages/shorebird_cli/lib/src/commands/login_command.dart
felangel: I was referring to the Shorebird CLI tool in this context. Happy to reword this.

packages/shorebird_cli/lib/src/auth/auth.dart
eseidel: OK.  Do we need to leave ourselves more breadcrumbs here about why this isn't secret or what we can/can't do with it?  Thanks.

packages/shorebird_cli/lib/src/commands/login_command.dart
eseidel: I think if it's the `shorebird` command we should say something like:
"The `shorebird` command needs your authorization to manage your Shorebird services on your behalf."

Close enough regardless.  Thanks.

felangel: Updated ✅

PR #224, opened 2023-04-04
feat(shorebird_cli): expose user

felangel:
## Description

- feat(shorebird_cli): expose user

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -121,7 +121,7 @@ class Auth {
 
 extension on AccessCredentials {
   User toUser() {
-    final claims = Jwt.decodeClaims(idToken!);
+    final claims = Jwt.decodeClaims(idToken ?? '');
 
     if (claims == null) throw Exception('Invalid JWT');
 
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -4,9 +4,11 @@ import 'dart:io';
 import 'package:googleapis_auth/auth_io.dart';
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as p;
+import 'package:shorebird_cli/src/auth/jwt.dart';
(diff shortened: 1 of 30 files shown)


packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -75,7 +78,11 @@ class Auth {
 
   AccessCredentials? _credentials;
 
-  AccessCredentials? get credentials => _credentials;
```
felangel, line 78: The `user` is exposed instead of `credentials`

eseidel (approved):

packages/shorebird_cli/lib/src/commands/login_command.dart
```diff
@@ -30,7 +29,7 @@ class LoginCommand extends ShorebirdCommand {
       await auth.login(prompt);
       logger.info('''
 
-🎉 ${lightGreen.wrap('Welcome to Shorebird! You are now logged in.')}
+🎉 ${lightGreen.wrap('Welcome to Shorebird! You are now logged in as <${auth.user!.email}>.')}
```
bryanoltman, line 32: Nice! Making a TODO to add this to `doctor`.

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -109,3 +118,17 @@ class Auth {
     _httpClient.close();
   }
 }
+
+extension on AccessCredentials {
+  User toUser() {
+    final claims = Jwt.decodeClaims(idToken ?? '');
```
bryanoltman: Can we throw if `idToken` is null? It's an entirely different issue than an invalid token, even if we ultimately handle it the same way.

bryanoltman (approved):
Love it

packages/shorebird_cli/lib/src/auth/auth.dart
felangel: Yup good call, updated ✅

PR #243, opened 2023-04-06
feat: Set FLUTTER_STORAGE_BASE_URL when using Shorebird-vended Flutter, warn user when path flutter version does not match shorebird flutter version

bryanoltman:
## Description

Adds warning to `run`, `build`, `patch`, `release`, and `doctor` commands if the version of Flutter on the user's path does not match the version of Flutter we vend.

Fixes https://github.com/shorebirdtech/shorebird/issues/220
Fixes https://github.com/shorebirdtech/shorebird/issues/238
Fixed https://github.com/shorebirdtech/shorebird/issues/237

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [x] 🗑️ Chore

PR #248, opened 2023-04-06
feat: Add FLUTTER_STORAGE_BASE_URL when using shorebird vended flutter

bryanoltman:
## Description

Sets the `FLUTTER_STORAGE_BASE_URL` to `"https://download.shorebird.dev/"` when using shorebird-vended Flutter.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_process.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_process.dart
@@ -50,7 +50,8 @@ abstract class ShorebirdProcess {
 
   static String _resolveExecutable(String executable) {
     if (executable == 'flutter') {
-      return ShorebirdPaths.flutterBinaryFile.path;
+      return 'FLUTTER_STORAGE_BASE_URL="https://download.shorebird.dev/" '
+          '${ShorebirdPaths.flutterBinaryFile.path}';
     }
 
     return executable;


packages/shorebird_cli/lib/src/shorebird_process.dart
```diff
@@ -50,7 +50,8 @@ abstract class ShorebirdProcess {
 
   static String _resolveExecutable(String executable) {
     if (executable == 'flutter') {
-      return ShorebirdPaths.flutterBinaryFile.path;
+      return 'FLUTTER_STORAGE_BASE_URL="https://download.shorebird.dev/" '
```
eseidel, line 53: You don't want to do this this way, because this depends on runInShell=true, which you may not want in all environments.

Instead, you will want to modify the env passed to the call.

eseidel: https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/shorebird_process.dart#L25

https://api.dart.dev/stable/2.10.5/dart-io/Process/run.html
says you can just set the one env and it will inherit by default.

bryanoltman:
This fix has been rolled into https://github.com/shorebirdtech/shorebird/pull/243

Issue #261, opened 2023-04-07
fix(shorebird_cli): release, build, etc commands don't identify serious issues


bryanoltman:
**Description**

As a user, I would expect the `release`, `build`, etc. commands to inform me of serious issues that would prevent code push from working (e.g., missing permissions in the android manifest). It currently does not.

bryanoltman:
Fixed by https://github.com/shorebirdtech/shorebird/pull/264

PR #264, opened 2023-04-07
chore(shorebird_cli): Update ShorebirdCommand to accept a list of validators instead of a single ShorebirdFlutterValidator

bryanoltman:
## Description

Replaces the `flutterValidator` constructor parameter of `ShorebirdCommand` with `validators` to facilitate more thorough validation for shorebird commands. This is pre-work for https://github.com/shorebirdtech/shorebird/issues/261.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/validators.dart
+++ b/packages/shorebird_cli/lib/src/validators/validators.dart
@@ -65,8 +65,8 @@ class ValidationIssue {
 /// Checks for a specific issue with either the Shorebird installation or the
 /// current Shorebird project.
 abstract class Validator {
-  /// The short name of this validator.
-  String get name => '$runtimeType';
+  /// A unique identifer for this class.
+  String get id => '$runtimeType';
 
   /// A one-sentence explanation of what this validator is checking.
   String get description;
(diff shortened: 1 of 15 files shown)


packages/shorebird_cli/lib/src/validators/validators.dart
```diff
@@ -65,6 +65,9 @@ class ValidationIssue {
 /// Checks for a specific issue with either the Shorebird installation or the
 /// current Shorebird project.
 abstract class Validator {
+  /// The short name of this validator.
+  String get name => '$runtimeType';
```
felangel: minor but this probably won't be a human-readable name in a snapshot so do we want to give this a different name like `id` or something to indicate that you shouldn't use this in logs/stdout?

felangel (approved):

packages/shorebird_cli/lib/src/validators/validators.dart
bryanoltman: Done.

PR #267, opened 2023-04-07
chore(shorebird_cli): bump shorebird engine revision

felangel:
## Description

- chore(shorebird_cli): bump shorebird engine revision

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [X] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/engine_revision.dart
+++ b/packages/shorebird_cli/lib/src/engine_revision.dart
@@ -1,2 +1,2 @@
 /// The hash of the Shorebird engine used by this version of Shorebird CLI.
-const shorebirdEngineRevision = '978a56f2d97f9ce24a2b6bc22c9bbceaaba0343c';
+const shorebirdEngineRevision = '7aa5c44764e10722d188ece75819f7d10f5269a3';


packages/shorebird_cli/lib/src/engine_revision.dart
```diff
@@ -1,2 +1,2 @@
 /// The hash of the Shorebird engine used by this version of Shorebird CLI.
-const shorebirdEngineRevision = '978a56f2d97f9ce24a2b6bc22c9bbceaaba0343c';
+const shorebirdEngineRevision = '7aa5c44764e10722d188ece75819f7d10f5269a3';
```
eseidel, line 2: I would expect we'd bump the version of flutter/flutter somewhere instead?

eseidel (approved):

packages/shorebird_cli/lib/src/engine_revision.dart
felangel: Ah okay and then we'd get the engine revision from `flutter --version`?

eseidel: I'm not sure what we need the engine revision for anymore?  Previously we wanted to make sure Flutter had compatible bits?  Now "flutter" is our direct dependency and "engine" is its dependency, and "updater" its, etc.  each responsible for validating their own.

PR #268, opened 2023-04-07
chore(artifact_proxy): add shorebirdtech/engine@7aa5c44764e to mappings

eseidel:
I also deleted the rust calls from bootstrap.sh (since they failed for me) and removed build_engine.sh since it's now done differently in shorebirdtech/build_engine


packages/artifact_proxy/lib/config.dart
```diff
@@ -8,62 +8,50 @@ const config = ProxyConfig(
   },
 );
 
+// On the assumption artifact layouts don't change very often for Flutter.
+class _EngineMapping37 extends EngineMapping {
```
felangel: Is the 37 supposed to represent Flutter 3.7.x?

eseidel: Yeah. I tried 3_7 and the linter got mad at me.

bryanoltman: Maybe `_EngineMapping3_7` instead?

eseidel: Will change and update the linter. :)

bryanoltman: Sniped 🙂 

I'd vote to override the linter on this one.

eseidel: Overriden. 🤣  PTAL.

bryanoltman (approved):

PR #274, opened 2023-04-08
fix(shorebird_cli): `shorebirdConfigDir` doesn't point to `testApplicationConfigHome` in tests

erikas-taroza:
## Description
`shorebirdConfigDir` points to `testApplicationConfigHome` if it is not null. However with the current implementation, it never points to the test directory.

**Example:**
Current (real and test):
```
/home/erikas/.config/shorebird
```

Fixed:
```
Real: /home/erikas/.config/shorebird
Test: /tmp/DNLFWF
```

The solution makes `shorebirdConfigDir` a getter instead of being a result from an invoked function.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


erikas-taroza:
Oops didn't create a branch for my other PR. I'll mark as draft until the other PR changes are merged so that the history is clean.

felangel:
Thanks for the contribution! This should be addressed by #294

PR #278, opened 2023-04-12
feat(cloud_logger): create `package:cloud_logger`

felangel:
## Description

- feat(cloud_logger): create `package:cloud_logger`
  - Based on the `package:gcp` logging implementation but mainly simplified for our use-case.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/cloud_logger/lib/src/log_entry.dart
+++ b/packages/cloud_logger/lib/src/log_entry.dart
@@ -1,0 +1,63 @@
+import 'dart:convert';
+
+import 'package:cloud_logger/cloud_logger.dart';
+import 'package:collection/collection.dart';
+import 'package:stack_trace/stack_trace.dart';
+
+/// Create a formatted log entry.
+String createLogEntry(
+  String? traceValue,
+  String message,
+  LogSeverity severity, {
+  Frame? stackFrame,
+}) {
+  // https://cloud.google.com/logging/docs/agent/logging/configuration#special-fields
+  final logContent = {
+    'message': message,
+    'severity': severity,
+    if (traceValue != null) 'logging.googleapis.com/trace': traceValue,
+    if (stackFrame != null)
+      'logging.googleapis.com/sourceLocation': _sourceLocation(stackFrame),
+  };
+  return json.encode(logContent);
+}
+
+/// Create a log from error information.
+/// https://cloud.google.com/functions/docs/monitoring/logging#writing_structured_logs
+String createErrorLogEntry(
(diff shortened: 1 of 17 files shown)


packages/cloud_logger/lib/src/logger.dart
```diff
@@ -0,0 +1,103 @@
...
+
+  @override
+  String toString() => 'LogSeverity $name ($value)';
+
+  /// Convert a [LogSeverity] to a JSON string.
+  String toJson() => name;
+}
+
+/// Returns the [Logger] for the current [Zone].
+Logger get logger => Zone.current[loggerKey] as Logger? ?? const _BasicLogger();
```
eseidel, line 91: Is it "typical" for a logging package to just expose a global getter like this?

packages/cloud_logger/lib/src/log_entry.dart
```diff
@@ -0,0 +1,63 @@
...
+}
+
+Chain _fromStackTrace(StackTrace? stackTrace) {
+  return (stackTrace == null ? Chain.current() : Chain.forTrace(stackTrace))
+      .foldFrames(
+    (f) =>
+        f.isCore ||
+        f.package == 'cloud_logging' ||
+        f.package == 'shelf' ||
+        f.package == 'dart_frog',
```
eseidel, line 60: I suspect eventually callers will want to extend this?

eseidel (approved):
Bryan may have helpful thoughts, but lgtm!

packages/cloud_logger/lib/src/log_entry.dart
felangel: Yeah this should eventually be configurable I'll add a todo.

eseidel:
FYI @kevmoo (no action expected, just thought you might like to see this.

packages/cloud_logger/lib/src/logger.dart
felangel: Imo this is the simplest approach because it'll allow us to access the "right" logger instance from anywhere in our backend code via `logger`. I thought about also using `RequestContext` but that would make it `Dart Frog` specific which didn't seem ideal.

felangel:
Actually after thinking about it a bit more I think just using `package:gcp` directly is the right way to go.

kevmoo:
PRs WELCOME for pkg:gcp !!

PR #297, opened 2023-04-17
feat(shorebird_cli): `shorebird login` verifies user

felangel:
## Description

- feat(shorebird_cli): `shorebird login` verifies user

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -6,7 +6,9 @@ import 'package:googleapis_auth/auth_io.dart' as oauth2;
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/auth/jwt.dart';
+import 'package:shorebird_cli/src/command.dart';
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -170,19 +180,15 @@ class Auth {
 }
 
 extension on oauth2.AccessCredentials {
-  String get email {
+  String? get email {
```
eseidel, line 183: I don't have the context to know, but should be be obscuring the fact that this is an email here? e.g. just treating it like an opaque identifier?

eseidel (approved):

packages/shorebird_cli/lib/src/auth/auth.dart
felangel: This is currently used to extract the email from the claims so that we can include the user's email in the output of commands like `shorebird account` or `shorebird login` (when they're already logged in). This code will have to be refactored once we support multiple OAuth providers 👍

PR #301, opened 2023-04-17
feat(shorebird_code_push_protocol): Add stripeCustomerId to User

bryanoltman:
## Description

Adds the stripeCustomerId field to the User model.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/models/user.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/models/user.dart
@@ -12,6 +12,7 @@ class User {
     required this.id,
     required this.email,
     this.hasActiveSubscription = false,
+    this.stripeCustomerId,
   });
 
   /// Converts a Map<String, dynamic> to a [User]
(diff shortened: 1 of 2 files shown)


packages/shorebird_code_push_protocol/lib/src/models/user.dart
```diff
@@ -28,4 +29,9 @@ class User {
 
   /// Whether the user is currently a paying customer.
   final bool hasActiveSubscription;
+
+  /// The user's Stripe customer id.
+  ///
+  /// This will be null for users that have never paid.
+  final String? stripeCustomerId;
```
felangel, line 36: Why is this necessary? I don't think we should expose this publicly imo.

bryanoltman: We need this on the server side, but I agree about not exposing it to the client – maybe we should create a different User model for the server?

felangel: Yeah I think we should create a UserEntity on the DB side which strictly corresponds to the DB schema and keep that decoupled from the public models.

PR #307, opened 2023-04-18
feat(shorebird_cli): use specific Flutter revision

eseidel:
Move to pulling a specific version of Flutter rather than a branch.

The idea is that `shorebird` is the thing pulled via a branch, and `flutter` is just a versioned dependency.  Lets you know that when you pull a specific version of `shorebird` you get the correct dependencies (where as as it works now you'd end up with some future flutter when pulling an old version of shorebird).


eseidel:
This depends on https://github.com/shorebirdtech/shorebird/pull/306 landing first, and also needs fixes to the `flutter doctor` commands to not fail.

packages/shorebird_cli/test/src/validators/shorebird_flutter_validator_test.dart
```diff
@@ -14,16 +14,16 @@ class _MockPlatform extends Mock implements Platform {}
 
 void main() {
   group('ShorebirdFlutterValidator', () {
+    const flutterRevision = '45fc514f1a9c347a3af76b02baf980a4d88b7879';
     const gitStatusMessage = """
 On branch stable
```
eseidel: Did you mean to keep this?

eseidel (commented):
lgtm

packages/shorebird_cli/test/src/validators/shorebird_flutter_validator_test.dart
felangel: updated, good catch

Issue #281, opened 2023-04-12
feat: Make sign-up self-serve


eseidel:
Ideally we would radically simplify our sign-up flow before we add another round of trusted testers.  Currently it takes about ~10 mins of manual work per tester.

Current flow is documented in our internal repo:
https://github.com/shorebirdtech/_shorebird/blob/main/TRUSTED_TESTER_ONBOARDING.md

eseidel:
When we open signups to the public (ideally in the next ~month or so?) we will also need signups to be fully self-serve.

PR #321, opened 2023-04-19
chore(shorebird_code_push_protocol): add CreateUserRequest message

bryanoltman:
## Description

Adds a `CreateUserRequest` message to support self-serve signup.

Part of https://github.com/shorebirdtech/shorebird/issues/281

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


packages/shorebird_code_push_protocol/lib/src/messages/create_user/create_user_request.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:shorebird_code_push_protocol/shorebird_code_push_protocol.dart';
+
+part 'create_user_request.g.dart';
+
+/// {@template create_user_request}
+/// The request body for POST /api/v1/users, which creates a new User.
+///
+/// Email is retrieved from the user's auth token.
```
felangel, line 9: Seems a bit weird to require an access token for user creation imo. I would have expected the endpoint would be open and we'd provide email and display name explicitly in the request body.

bryanoltman: I originally had designed it that way and then realized I could (potentially maliciously) create accounts for anyone I wanted to. This approach adds some level of verification that the person creating the user is who they say they are, although it will probably need to be tweaked a bit if/when we expand our auth offerings.

felangel: Yeah I see makes sense to do that for now since that's simpler/safer. We can change it in the future 👍

felangel (approved):

PR #334, opened 2023-04-20
feat(shorebird_cli): Add account create command

bryanoltman:
## Description

Adds the `shorebird account create` command to allow users to sign up for Shorebird.

Example of a user attempting to log in without an account and subsequently running `shorebird account create`:

Screenshot 2023-04-20 at 3 33 24 PM

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -111,7 +111,13 @@ class Auth {
     );
   }
 
-  Future<void> login(void Function(String) prompt) async {
+  /// Logs the string returned by [prompt] to the console and obtains auth
+  /// credentials. If [verifyEmail] is true, check that there is a Shorebird
+  /// user with the same email address as the newly-authenticated user.
+  Future<void> login(
+    void Function(String) prompt, {
+    bool verifyEmail = true,
```
felangel: Thoughts on exposing a `signUp` instead of having `login` do both `login` and `signUp` based on a `bool`?

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
```diff
@@ -0,0 +1,98 @@
...
+  String get name => 'create';
+
+  @override
+  Future<int> run() async {
+    final CodePushClient client;
+
+    if (!auth.isAuthenticated) {
+      try {
+        await auth.login(prompt, verifyEmail: false);
+        client = buildCodePushClient(
```
felangel: why are we creating the client in this block? We're already creating it on line 47 so can we delete this?

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
```diff
@@ -0,0 +1,98 @@
...
+      return ExitCode.success.code;
+    } catch (error) {
+      progress.fail(error.toString());
+      return ExitCode.software.code;
+    }
+  }
+
+  void prompt(String url) {
+    logger.info('''
+Shorebird is currently only open to trusted testers. To participate, you will need a Google account for authentication.
```
eseidel: Shorebird currently requires a Google account for authentication.  If you'd like to use a different kind of auth, please let us know:
https://github.com/shorebirdtech/shorebird/issues/335

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
```diff
@@ -0,0 +1,98 @@
...
+
+The first step is to sign in with a Google account. Please follow the sign-in link below:
```
eseidel: Please follow the link to sign-in with your Google account:

eseidel (approved):

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
eseidel: My message is I don't think we need the "trusted testers" warning, we can enforce that in other ways.

bryanoltman: Done

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
bryanoltman: Went with "Follow the link below to authenticate:" because "your Google account" felt redundant after "Shorebird currently requires a Google account for authentication". lmk what you think

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: I refactored auth a bit to make this somewhat more ergonomic in the commands. lmk what you think, happy to iterate

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
bryanoltman: This has been deleted, resolving

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -111,7 +106,7 @@ class Auth {
     );
   }
 
-  Future<void> login(void Function(String) prompt) async {
+  Future<void> getCredentials(void Function(String) prompt) async {
```
felangel: Should this be called `loadCredentials`? It's strange that it's not returning the credentials based on the naming.

felangel: I still think it's strange for `auth` to not expose a `login` and `signup` since that's what most auth abstractions do (e.g. FirebaseAuth). Currently we also still have `logout` as a public method on `Auth` which makes things confusing imo. Is the main benefit of this change to avoid having to pass the two prompts to the signup call? I think the flow is a bit confusing in this state.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -45,7 +53,9 @@ class CodePushClient {
     final uri = Uri.parse('$hostedUri/api/v1/users/me');
     final response = await _httpClient.get(uri);
 
-    if (response.statusCode != HttpStatus.ok) {
+    if (response.statusCode == HttpStatus.notFound) {
+      throw UserNotFoundException();
```
felangel: rather than throwing thoughts on making this return `Future<User?>`? I think that would be cleaner/more explicit than us having to know to `catch UserNotFoundException`.

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: Re: naming, totally agreed.  `loadCredentials` is a better name (or `fetchCredentials`), and `logout` should probably be `clearCredentials`.

Let me take a stab at the login/signup approach. I was avoiding this since the flow is somewhat interactive, but it might not be so bad.

felangel: I think overall it'll be easier to reason about because I feel Auth shouldn't expose low-level implementation details about loading/clearing credentials and should instead expose higher abstractions like login, logout, signup, etc.

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -192,3 +233,36 @@ extension on oauth2.AccessCredentials {
...
+}
+
+/// {@template user_already_exists_exception}
+/// Thrown when an attempt to create a User object results in a 409.
+/// {@endtemplate}
+class UserAlreadyExistsException implements Exception {
+  /// {@macro user_already_exists_exception}
+  UserAlreadyExistsException(this.user);
+
+  /// The extant user.
```
felangel:
```suggestion
  /// The existing user.
```

packages/shorebird_cli/lib/src/commands/login_command.dart
```diff
@@ -18,26 +18,35 @@ class LoginCommand extends ShorebirdCommand {
...
         ..info("Run 'shorebird logout' to log out and try again.");
       return ExitCode.success.code;
+    } on UserNotFoundException catch (error) {
+      logger
+        ..err(
+          '''
+We could not find a Shorebird account for ${error.email}.''',
+        )
+        ..info(
+          """If you have not yet created an account, you can do so by running "${green.wrap('shorebird account create')}". If you believe this is an error, please reach out to us via Discord, we're happy to help!""",
```
felangel: minor but we've been using `lightCyan` to wrap commands in other places so might want to align on a single style for commands

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -126,6 +128,9 @@ class Auth {
       final codePushClient = _buildCodePushClient(httpClient: this.client);
 
       final user = await codePushClient.getCurrentUser();
```
felangel, line 130: Thoughts on caching the user instead of the email so that we don't need to call `getCurrentUser()` in other places like cancel_subscription?

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
```diff
@@ -0,0 +1,75 @@
...
+      logger
+        ..info('You are already logged in as <${error.email}>.')
+        ..info("Run 'shorebird logout' to log out and try again.");
+      return ExitCode.success.code;
+    } on UserAlreadyExistsException catch (error) {
+      // TODO(bryanoltman): change this message based on the user's subscription
+      // status.
+      logger.info('''
+You already have an account, ${error.user.displayName}!
+To subscribe, run ${green.wrap('shorebird account subscribe')}.
```
felangel: same comment here about using `lightCyan` to wrap commands for consistency.

packages/shorebird_cli/lib/src/commands/subscription/cancel_subscription_command.dart
```diff
@@ -34,7 +34,7 @@ class CancelSubscriptionCommand extends ShorebirdCommand
 
     final User user;
     try {
-      user = await client.getCurrentUser();
+      user = (await client.getCurrentUser())!;
```
felangel: If you're already logged in we could cache the user and avoid having to make this extra request

packages/shorebird_cli/lib/src/commands/account_command.dart
```diff
@@ -1,29 +0,0 @@
-import 'dart:async';
```
felangel, line 1: 💯

packages/shorebird_cli/lib/src/commands/login_command.dart
bryanoltman: Fixed

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
bryanoltman: Fixed

packages/shorebird_cli/lib/src/commands/subscription/cancel_subscription_command.dart
bryanoltman: Not sure I follow – we won't be running login or signUp in this command, and we'll definitely want to get the latest subscription info for this user before dealing with subscription info.

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: Updated

packages/shorebird_code_push_client/lib/src/code_push_client.dart
bryanoltman: Done.

packages/shorebird_cli/lib/src/commands/subscription/cancel_subscription_command.dart
felangel: Oh sorry I think I misread this. We might want to remove the `auth.isAuthenticated` check here and instead handle the case where `getCurrentUser` is null rather than force unwrapping though 🤔

packages/shorebird_cli/test/src/commands/account/create_account_command_test.dart
```diff
@@ -0,0 +1,150 @@
...
+
+class _MockLogger extends Mock implements Logger {}
+
+class _MockUser extends Mock implements User {}
+
+void main() {
+  const userName = 'John Doe';
+  const email = 'tester@shorebird.dev';
+
+  late Auth auth;
```
felangel: minor but I think in other tests we keep these scoped to the group they correspond to we should just be consistent if possible 🤷

packages/shorebird_cli/lib/src/commands/subscription/cancel_subscription_command.dart
bryanoltman: I think this should still fail in the same way as before – if the user is null or if getCurrentUser fails in some other way, we log the error and exit. What's your concern?

packages/shorebird_cli/lib/src/commands/account/create_account_command.dart
```diff
@@ -0,0 +1,75 @@
...
+    final User newUser;
+    try {
+      newUser = await auth.signUp(
+        authPrompt: authPrompt,
+        namePrompt: namePrompt,
+      );
+    } on UserAlreadyLoggedInException catch (error) {
+      logger
+        ..info('You are already logged in as <${error.email}>.')
+        ..info("Run 'shorebird logout' to log out and try again.");
```
felangel:
```suggestion
        ..info("Run ${lightCyan.wrap('shorebird logout')} to log out and try again.");
```

minor

packages/shorebird_cli/lib/src/commands/login_command.dart
```diff
@@ -18,26 +18,35 @@ class LoginCommand extends ShorebirdCommand {
...
   @override
   Future<int> run() async {
-    if (auth.isAuthenticated) {
+    try {
+      await auth.login(prompt);
+    } on UserAlreadyLoggedInException catch (error) {
       logger
-        ..info('You are already logged in as <${auth.email}>.')
+        ..info('You are already logged in as <${error.email}>.')
         ..info("Run 'shorebird logout' to log out and try again.");
```
felangel:
```suggestion
        ..info("Run ${lightCyan.wrap('shorebird logout')} to log out and try again.");
```

nit

packages/shorebird_cli/lib/src/commands/subscription/cancel_subscription_command.dart
felangel: My concern is mainly that the error from force unwrapping a null value won't be helpful to the end user

packages/shorebird_cli/test/src/commands/account/create_account_command_test.dart
bryanoltman: Fixed

packages/shorebird_cli/lib/src/commands/subscription/cancel_subscription_command.dart
bryanoltman: Ah! That makes sense. Updated.

felangel (approved):
LGTM ✅

PR #339, opened 2023-04-20
feat(code_push_protocol): add subscription object to user

bryanoltman:
## Description

Adds the `Subscription` object to our API contract and adds a `Subscription? subscription` to `User`.

The info provided by this object will allow us to provide richer information to the user about their subscription (price, when/if it expires, when they will be charged, etc.)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// When the subscription will be renewed or expire.
+  @TimestampConverter()
+  final DateTime paidThroughDate;
+
+  /// Whether this subscription will renew on [paidThroughDate].
+  final bool willRenew;
+
+  /// Whether this subscription is currently active.
+  bool get isActive => paidThroughDate.isAfter(DateTime.now());
```
felangel, line 35: will this potentially conflict with `User.hasActiveSubscription`?

bryanoltman: Yes. My goal is to keep this around until we update the CLI to no longer expect it and then change it to a computed property or delete it entirely.

packages/shorebird_code_push_protocol/lib/src/models/user.dart
```diff
@@ -32,4 +34,7 @@ class User {
 
   /// Whether the user is currently a paying customer.
   final bool hasActiveSubscription;
+
+  /// The user's current subscription, if any.
+  final Subscription? subscription;
```
felangel: This seems like it will cause confusion because we would be exposing two ways to determine whether a user has an active subscription which aren't guaranteed to be in sync.

felangel: Also do we want to tightly couple `User` and `Subscription` so that querying for a user also queries for the user's subscription? What if the user has more than one subscription in the future?

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
```diff
@@ -0,0 +1,36 @@
...
+    required this.willRenew,
+  });
+
+  /// Creates a [Subscription] from a JSON object.
+  factory Subscription.fromJson(Json json) => _$SubscriptionFromJson(json);
+
+  /// Converts a [Subscription] to a JSON object.
+  Json toJson() => _$SubscriptionToJson(this);
+
+  /// Billing rate, in cents.
```
eseidel, line 24: Oh man.  US Cents I assume. :) Someone is going to write an amazing typed "money" object someday. 🤣

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
```diff
@@ -0,0 +1,36 @@
...
+  final int cost;
+
+  /// When the subscription will be renewed or expire.
+  @TimestampConverter()
+  final DateTime paidThroughDate;
```
eseidel, line 29: I assume this is midnight UTC on that date?

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
eseidel: (Might want to leave a TODO with those plans.)

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
felangel: https://pub.dev/packages/money2

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
bryanoltman: Discussed offline: we'll keep Subscription and User separate. Resolving.

packages/shorebird_code_push_protocol/lib/src/models/user.dart
bryanoltman: Removing Subscription from User. Resolving.

bryanoltman: Same here, resolving

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
bryanoltman: Assuming usd for now. 

Stripe specifies `currency` on the individual price objects (https://stripe.com/docs/api/prices/object#price_object-currency), so we can get fancier if/when we need to.

packages/shorebird_code_push_protocol/lib/src/models/subscription.dart
bryanoltman: [Not specified](https://stripe.com/docs/api/subscriptions/object#subscription_object-current_period_end), but looking at some of the live subscriptions, it seems to be `purchase_time + Duration(days: 30)`

packages/shorebird_code_push_protocol/lib/src/models/user.dart
```diff
@@ -1,4 +1,5 @@
 import 'package:json_annotation/json_annotation.dart';
+import 'package:shorebird_code_push_protocol/src/models/subscription.dart';
```
felangel: Can this import be deleted?

packages/shorebird_code_push_protocol/test/src/models/subscription_test.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:shorebird_code_push_protocol/src/models/models.dart';
```
felangel:
```suggestion
import 'package:shorebird_code_push_protocol/shorebird_code_push_protocol.dart';
```
nit

bryanoltman: Done

packages/shorebird_code_push_protocol/lib/src/models/user.dart
bryanoltman: Yep, done

felangel (approved):

PR #340, opened 2023-04-20
refactor(artifact_proxy): use per-revision artifacts manifest

### LINKED ISSUE #285: Re-write proxy server to use per-revision config
We could delete the central config for the proxy server and instead make it per-revision config.

This would just mean we upload download.shorebird.dev/$revision/engine.version or similar that gives us an id that we're mapping back to.

The proxy server would then cache these mappings in memory.

felangel:
## Description

- refactor(artifact_proxy): use per-revision artifacts manifest (closes #285)

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/artifact_proxy/lib/config.dart
```diff
@@ -1,70 +1,56 @@
...
-  '978a56f2d97f9ce24a2b6bc22c9bbceaaba0343c': _EngineMapping3_7(
-    flutterEngineRevision: 'ec975089acb540fc60752606a3d3ba809dd1528b',
-  ),
-  '7aa5c44764e10722d188ece75819f7d10f5269a3': _EngineMapping3_7(
-    flutterEngineRevision: 'ec975089acb540fc60752606a3d3ba809dd1528b',
-  ),
+/// Patterns for all artifacts paths that Flutter
+/// requests which contain an engine revision.
+final engineArtifactPatterns = {
+  r'flutter_infra_release\/flutter\/(.*)\/android-arm-64-release\/artifacts\.zip',
```
eseidel: This is clearly not everything.  Did you try `precache -a`?  I don't see windows for example in this list.

packages/artifact_proxy/lib/src/artifact_manifest_client.dart
```diff
@@ -0,0 +1,48 @@
...
+
+  final http.Client _httpClient;
+
+  final _cache = <String, ArtifactsManifest>{};
+
+  /// Fetches the [ArtifactsManifest] for the provided [revision] from the
+  /// shorebird storage bucket.
+  Future<ArtifactsManifest> getManifest(String revision) async {
+    if (_cache.containsKey(revision)) return _cache[revision]!;
+    final manifest = await _fetchManifest(revision);
```
eseidel, line 26: I think you will want to cache negative results with a TTL?  I guess as an MVP doesn't matter.

packages/artifact_proxy/lib/src/artifact_manifest_client.dart
```diff
@@ -0,0 +1,48 @@
...
+/// and caches them in memory.
+/// {@endtemplate}
+class ArtifactManifestClient {
+  /// {@macro artifact_manifest_client}
+  ArtifactManifestClient({http.Client? httpClient})
+      : _httpClient = httpClient ?? http.Client();
+
+  final http.Client _httpClient;
+
+  final _cache = <String, ArtifactsManifest>{};
```
eseidel: You'll want this to be a fixed-size LRU at some point.

packages/artifact_proxy/lib/config.dart
felangel: Updated sorry about that

packages/artifact_proxy/lib/src/artifact_manifest_client.dart
felangel: done 👍

eseidel (approved):

PR #343, opened 2023-04-21
chore(shorebird_cli): remove unnecessary trailing `/` from storage url

felangel:
## Description

- chore(shorebird_cli): remove unnecessary trailing `/` from storage url

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [X] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_process.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_process.dart
@@ -106,7 +106,7 @@ class ShorebirdProcess {
     if (executable == 'flutter') {
       // If this ever changes we also need to update the `shorebird` shell
       // wrapper which downloads runs Flutter to fetch artifacts the first time.
-      return {'FLUTTER_STORAGE_BASE_URL': 'https://download.shorebird.dev/'};
+      return {'FLUTTER_STORAGE_BASE_URL': 'https://download.shorebird.dev'};
     }
 
     return {};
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/shorebird_process.dart
```diff
@@ -106,7 +106,7 @@ class ShorebirdProcess {
     if (executable == 'flutter') {
       // If this ever changes we also need to update the `shorebird` shell
       // wrapper which downloads runs Flutter to fetch artifacts the first time.
-      return {'FLUTTER_STORAGE_BASE_URL': 'https://download.shorebird.dev/'};
+      return {'FLUTTER_STORAGE_BASE_URL': 'https://download.shorebird.dev'};
```
felangel, line 109: The trailing slash results in `//` since flutter_tool already includes a `/` as part of the path.

eseidel: Is the system sensitive to this slash existing?  (If so we might want to fix that part of the system.).

eseidel (approved):

packages/shorebird_cli/lib/src/shorebird_process.dart
felangel: No I just noticed it while I was updating the proxy so I figured we'd remove it.

PR #350, opened 2023-04-21
fix: Speculative fix for missing Linux x64 assets

eseidel:
Turns out `flutter precache -a` doesn't download host architecture variants (at least for Linux, it seems to for Mac).

Filed a bug with Flutter: https://github.com/flutter/flutter/issues/125344


felangel (approved):
Do you know why this was not encountered during our testing? I did multiple runs with a clear cache and never hit this 🤔

eseidel:
I believe it's simply a bug in `precache --all`.  I think precache all is a dream, but not actually tested.

Issue #351, opened 2023-04-21
feat: `shorebird --version` should print flutter version too


eseidel:
Sorta like how `flutter --version` prints its dependent versions:

```
eseidel@erics-mbp GitHub % flutter --version
Flutter 3.7.12 • channel stable • https://github.com/flutter/flutter.git
Framework • revision 4d9e56e694 (4 days ago) • 2023-04-17 21:47:46 -0400
Engine • revision 1a65d409c7
Tools • Dart 2.19.6 • DevTools 2.20.1
```

Toavina23:
so those informations would just be added to the actual info on shorebird --version command or just the flutter version line ?

felangel:
This is done by #1075

PR #352, opened 2023-04-22
feat: add flutter version to --version tools list

Toavina23:
## Description

I have added a tool list when using shorebird --version command, as asked by @eseidel at issue #351 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/test/src/command_runner_test.dart
```diff
@@ -67,13 +68,21 @@ void main() {
 
     group('--version', () {
       test('outputs current version and engine revisions', () async {
+        final versionResult = await Process.run('flutter', ['--version']);
```
bryanoltman, line 70: I don't think we want to spawn a process from our test – it would make this test a fair bit slower and makes this test depend on a correctly configured environment (which I believe is why it's failing in CI). We use `ShorebirdProcess` as a wrapper for `Process` as a way to get around this elsewhere in tests. See the `testProcess` parameter we supply to our Commands in other tests, such as `shorebird/packages/shorebird_cli/test/src/commands/release_command_test.dart`.

Toavina23: Ok, I'll fix that. thx for the review

Toavina23: > I don't think we want to spawn a process from our test – it would make this test a fair bit slower and makes this test depend on a correctly configured environment (which I believe is why it's failing in CI). We use `ShorebirdProcess` as a wrapper for `Process` as a way to get around this elsewhere in tests. See the `testProcess` parameter we supply to our Commands in other tests, such as `shorebird/packages/shorebird_cli/test/src/commands/release_command_test.dart`.

I a bit confused on how it should be used on my test here, because here I am actually just testing an option of the 
`ShorebirdCliCommandRunner` rather than a `ShorebirdCommand` which in that case, I understand that I can pass it the shorebird process but in the case of a `CompletionCommandRunner` I can't manage to find a solution. 
It would be great if can give me a bit more indication on they way to implement this. 

Thank you, in advance

eseidel:
I definitely still want this change. :) Just looks like it needs some edits.  Let us know on Discord (https://discord.gg/shorebird) if we can help!  Thank you for proposing this!

eseidel:
I started on this in https://github.com/shorebirdtech/shorebird/pull/598 but didn't finish.

felangel:
This is done now as part of #1075

PR #353, opened 2023-04-23
First attempt at `shorebird` command for windows

eseidel:
Writing from scratch since Flutter still uses `.bat` files and I hope that's not still the right choice in 2023.


eseidel:
@bryanoltman this is yours to continue should you wish (you're also welcome to start from scratch).

eseidel:
Bryan solved this for real.

PR #362, opened 2023-04-25
fix(shorebird_cli): use path.join to construct paths

bryanoltman:
## Description

Updates path construction logic to use [`path.join`](https://api.flutter.dev/flutter/package-path_path/join.html) instead of hardcoding the path separator character as `/`, which doesn't work on Windows. Also changes a few invocations of `flutter` to spawn in a shell to fix `shorebird doctor`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
+++ b/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
@@ -159,6 +159,7 @@ This can cause unexpected behavior if you are switching between the tools and th
       'flutter',
       ['--version'],
       useVendedFlutter: !checkPathFlutter,
+      runInShell: true,
     );
 
     if (result.exitCode != 0) {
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
```diff
@@ -159,6 +159,7 @@ This can cause unexpected behavior if you are switching between the tools and th
       'flutter',
       ['--version'],
       useVendedFlutter: !checkPathFlutter,
+      runInShell: true,
```
eseidel, line 162: Does windows require runInShell (I seem to recall it does)

bryanoltman: Yep, although seemingly not always. The git commands run find without it. I'm sure I'll learn a lot more about this in the days to come.

felangel (approved):

PR #378, opened 2023-04-26
chore(shorebird_cli): don't use FlutterVersionValidator for all commands

### LINKED ISSUE #341: flutter validator could be a range of allowed flutters?
"we don't know" is the scary part.  "Isn't that your responsibility?  I certainly don't know" -- tt

bryanoltman:
## Description

Don't run `ShorebirdFlutterValidator` for all commands. Also only compare major and minor versions (i.e., ignore patch version) when comparing Shorebird and path Flutter versions.

Fixes https://github.com/shorebirdtech/shorebird/issues/341

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
```diff
@@ -82,21 +82,29 @@ class ShorebirdFlutterValidator extends Validator {
       );
     }
 
-    if (shorebirdFlutterVersion != null &&
-        pathFlutterVersion != null &&
-        shorebirdFlutterVersion != pathFlutterVersion) {
-      final message = '''
+    if (shorebirdFlutterVersion != null && pathFlutterVersion != null) {
+      final shorebirdVersionComponents = shorebirdFlutterVersion.split('.');
```
felangel: minor but we can also use `Version.parse` to access major, minor, patch segments

felangel (approved):

packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
bryanoltman: Saw that, didn't know if we wanted to take on a new dependency just for this. wdyt?

bryanoltman: It looks like their parsing is a bit more robust, so I decided to pull that in

PR #405, opened 2023-04-28
feat(shorebird_cli): enable autofixing issues with `shorebird doctor --fix`

bryanoltman:
## Description

- Adds the `--fix` flag to `shorebird doctor`.
- Adds an optional `fix` function to `ValidationIssue`.
- Implements `fix` for `AndroidInterentPermissionValidator`.

Screenshot 2023-05-01 at 10 43 58 PM

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel:
I wonder if we shouldn't expose a separate command like `shorebird fix` for this to keep doctor focused on just diagnosing the health of your installation/setup.

bryanoltman:
@felangel Good call. I put this behind a flag and (I think) improved the copy. Let me know what you think!

packages/shorebird_cli/lib/src/commands/doctor_command.dart
```diff
@@ -36,33 +43,74 @@ class DoctorCommand extends ShorebirdCommand with ShorebirdVersionMixin {
...
+        if (shouldFix) {
+          // If --fix flag was used and there are fixable issues, fix them.
+          progress.update('Fixing');
+          for (final issue in fixableIssues) {
+            await issue.fix!();
+          }
+
+          // Re-run the validator to see if there are any remaining issues that
+          // we couldn't fix.
+          unresolvedIssues = await validator.validate(process);
```
felangel, line 80: Won't this result in potentially lots of validations running twice? In what circumstances would fix complete successfully without actually fixing the issue?

packages/shorebird_cli/lib/src/validators/android_internet_permission_validator.dart
```diff
@@ -90,4 +91,20 @@ class AndroidInternetPermissionValidator extends Validator {
     return attribute.qualifiedName == 'android:name' &&
         attribute.value == 'android.permission.INTERNET';
   }
+
+  void _addInternetPermissionToFile(String path) {
+    final xmlDocument = XmlDocument.parse(File(path).readAsStringSync());
```
felangel, line 96: what if this throws because the file at `path` does not exist?

packages/shorebird_cli/lib/src/validators/android_internet_permission_validator.dart
```diff
@@ -67,6 +67,7 @@ class AndroidInternetPermissionValidator extends Validator {
             (String manifestPath) => ValidationIssue(
               severity: ValidationIssueSeverity.error,
               message: '$manifestPath is missing the INTERNET permission.',
+              fix: () async => _addInternetPermissionToFile(manifestPath),
```
felangel:
```suggestion
              fix: () => _addInternetPermissionToFile(manifestPath),
```
does it need to be async?

packages/shorebird_cli/lib/src/validators/validators.dart
```diff
@@ -30,14 +30,21 @@ extension Display on ValidationIssueSeverity {
...
+  });
 
   /// How important it is to fix this issue.
   final ValidationIssueSeverity severity;
 
   /// A description of the issue.
   final String message;
 
+  /// Fixes this issue.
+  final Future<void> Function()? fix;
```
felangel:
```suggestion
  final FutureOr<void> Function()? fix;
```
minor but couldn't this be `FutureOr` since not all fixes will need to be async?

packages/shorebird_cli/lib/src/commands/doctor_command.dart
bryanoltman: > Won't this result in potentially lots of validations running twice

Not necessarily. I imagine `validate` running at most twice per validator: once to detect the errors, once to verify we've fixed them. 

> In what circumstances would fix complete successfully without actually fixing the issue?

They're all sort of edge-casey, but maybe we can't write to the AndroidManifest.xml file for some reason. I wanted to guard against telling the user that we fixed issues without actually having done so.

packages/shorebird_cli/lib/src/validators/android_internet_permission_validator.dart
bryanoltman: The user would probably see an ugly error. I'm not particularly worried about this throwing because it's private to this class and the we only provide it the results of `listSync`.

packages/shorebird_cli/lib/src/validators/android_internet_permission_validator.dart
bryanoltman: Yes – `fix` is defined as `final Future<void> Function()? fix;`

bryanoltman: Saw your next comment, will remove

packages/shorebird_cli/lib/src/validators/validators.dart
bryanoltman: Done

packages/shorebird_cli/lib/src/commands/doctor_command.dart
felangel: Wouldn't fix throw in those cases? It just feels weird that we don't trust a completed fix to fix the issue and instead need to re-run the validator. Why is it not enough to run fix in a `try/catch` and if it throws mark it as failed otherwise assume it succeeded?

bryanoltman: Maybe! It may also be possible that fixing one issue allows the validator to find an error further along in the validation.

felangel: hmm I'm just worried that we'll end up with multiple "heavy" validators and the overall developer experience will be bad when trying to apply fixes 😅

felangel (approved):

packages/shorebird_cli/lib/src/commands/doctor_command.dart
bryanoltman: Talked offline, resolving

packages/shorebird_cli/lib/src/validators/android_internet_permission_validator.dart
bryanoltman: Talked offline. While I'm not worried about the one fix implemented in this PR, we probably should endeavor to have a user-friendly message displayed instead of a potentially cryptic error message. Updating now.

Issue #407, opened 2023-04-29
Add ability to list and delete releases


eseidel:
This may become less of a problem after https://github.com/shorebirdtech/shorebird/issues/368 is fixed, but right now you can get stuck if you ever put in a letter in a release name.

A TT just now had 3 releases for their app

1.0.0+1
1.0.0+2
Y

Y of course sorts later than 1, so was always the "latest" 🤦 

TT had no way to list or delete releases themselves w/o deleting the app.

bryanoltman:
Reopening until docs are added

PR #427, opened 2023-05-02
feat(shorebird_cli): add releases list command

bryanoltman:
## Description

Adds `shorebird releases list` command to list releases for the current app, as determined by the app id in shorebird.yaml.

Part of https://github.com/shorebirdtech/shorebird/issues/407

Screenshot 2023-05-02 at 4 28 36 PM

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/releases/list_releases_command.dart
```diff
@@ -0,0 +1,104 @@
...
+  String get description => 'List all releases for this app.';
+
+  @override
+  Future<int> run() async {
+    if (!auth.isAuthenticated) {
+      printNeedsAuthInstructions();
+      return ExitCode.noUser.code;
+    }
+
+    if (!hasShorebirdYaml) {
```
felangel, line 35: It’s probably fine as is but might be nicer to default to the shorebirdYaml appId and allow it to be overridden via an option

packages/shorebird_cli/lib/src/commands/releases/list_releases_command.dart
```diff
@@ -0,0 +1,104 @@
...
+            ],
+          )
+        ],
+      ),
+      body: TableSection(
+        rows: [
+          for (final release in this)
+            Row(
+              cells: [
+                Cell(release.displayName ?? '(no name)'),
```
felangel: I think we used -- in other places as an empty placeholder

packages/shorebird_cli/lib/src/commands/releases/list_releases_command.dart
bryanoltman: Yeah, I think that can be added as needed. I imagine most usage of this command will be from within the project dir.

packages/shorebird_cli/test/src/commands/releases/list_releases_command_test.dart
```diff
@@ -0,0 +1,160 @@
...
+      final exitCode = await IOOverrides.runZoned(
+        command.run,
+        getCurrentDirectory: () => tempDir,
+      );
+
+      expect(exitCode, ExitCode.success.code);
+      verify(
+        () => logger.info('''
+┌───────────┬─────────┐
+│ Name      │ Version │
```
felangel: Minor but might want to switch the order to show version first since that is guaranteed to be set

packages/shorebird_cli/lib/src/commands/releases/list_releases_command.dart
bryanoltman: Done

packages/shorebird_cli/test/src/commands/releases/list_releases_command_test.dart
bryanoltman: Done.

packages/shorebird_cli/lib/src/commands/releases/list_releases_command.dart
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (!hasShorebirdYaml) {
+      logger.err(
+        '''Shorebird is not initialized. Did you run ${lightCyan.wrap('shorebird init')}?''',
+      );
+      return ExitCode.config.code;
+    }
+
+    final appId = getShorebirdYaml()!.appId;
```
felangel:
```suggestion
    final appId = getShorebirdYaml()!.appId.value!;
```
so we don't need to force unwrap a bunch of times later on

bryanoltman: Done

felangel (approved):

Issue #389, opened 2023-04-27
fix: Support for multiple Flavors


MaximilianFlechtner:
**Description**

I’m using [Very Good CLI](https://cli.vgv.dev/), which has the ability to add multiple flavors to the app, something that is not currently supported by Shoebird.

**Steps To Reproduce**

1. Create a new app with Very Good CLI 
2. shorebird release -- --flavor production --target lib/main_production.dart

**Expected Behavior**

Should build the app with the right flavor

**Screenshots**

<img width="2274" alt="Screenshot 2023-04-27 at 08 40 20" src="https://user-images.githubusercontent.com/66118057/234780270-f70741b9-56c4-4c6e-a428-297af56a8d50.png">

eseidel:
Ah, this should be relatively straightforward to fix.  We just need to teach `shorebird release` where the resulting artifacts get built.  Thanks for the report!

felangel:
Proposal:

1. Adjust the `app_id` in `shorebird.yaml` to be either a `String` or a `Map<String, String>`

```yaml
# No Flavors (default)
app_id: 6ff122d8-6d01-461b-a2bd-20dd774a5143
```

```yaml
# With Flavors
app_id:
  development: 6ff122d8-6d01-461b-a2bd-20dd774a5143
  production: 8d3155a8-a048-4820-acca-824d26c29b71
```

2. `shorebird init` to detect flavors are present and generate an app per flavor with the corresponding `shorebird.yaml` and include the flavor as part of the app's `display_name`
3. `shorebird release --flavor <FLAVOR>` to build artifacts for the specified flavor and ensure the bundled `shorebird.yaml` always has the correct `app_id` for the corresponding flavor

```
shorebird release --flavor development
```

resulting `flutter_assets/shorebird.yaml`

```
app_id: 6ff122d8-6d01-461b-a2bd-20dd774a5143
```

This way there are no changes to the backend or engine required.

PR #429, opened 2023-05-02
feat(shorebird_cli): `shorebird init` supports flavors

felangel:
## Description

- feat(shorebird_cli): `shorebird init` supports flavors (part of #389)

https://user-images.githubusercontent.com/8855632/235798840-63ba92eb-4ccb-4f7e-9593-45e55f1674e7.mp4

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/test/src/commands/init_command_test.dart
```diff
@@ -215,8 +264,18 @@ Please make sure you are running "shorebird init" from the root of your Flutter
       );
       expect(
         File(p.join(tempDir.path, 'shorebird.yaml')).readAsStringSync(),
-        contains('app_id: $appId'),
+        contains('''
+app_id:
```
eseidel, line 360: We will need to update the rust library to handle this too I presume?

felangel: No because I’m planning to add a post build step to adjust the bundled shorebird.yaml to include the app_id of the respective flavor so no changes should be required on the rust or backend.

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
```diff
@@ -9,7 +9,10 @@ import 'package:yaml/yaml.dart';
 import 'package:yaml_edit/yaml_edit.dart';
 
 mixin ShorebirdConfigMixin on ShorebirdCommand {
-  bool get hasShorebirdYaml => getShorebirdYaml() != null;
+  bool get hasShorebirdYaml {
+    final file = File(p.join(Directory.current.path, 'shorebird.yaml'));
+    return file.existsSync();
+  }
```
bryanoltman: This is more efficient, but I'm curious if there was a different motivation for rewriting this. 

Also, take or leave, but this could be a one-liner:

```dart
bool get hasShorebirdYaml => File(p.join(Directory.current.path, 'shorebird.yaml')).existsSync();
```

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
```diff
@@ -50,20 +53,24 @@ mixin ShorebirdConfigMixin on ShorebirdCommand {
...
 # Learn more at https://shorebird.dev
 # This file should be checked into version control.
 
 # This is the unique identifier assigned to your app.
 # It is used by your app to request the correct patches from Shorebird servers.
-app_id: $appId
-''');
+app_id:
+''';
+    final yaml = File(p.join(Directory.current.path, 'shorebird.yaml'));
```
bryanoltman: Nit: this would be a tiny bit easier to follow if this were created after the `editor`, possibly as a one-liner:

```dart
final editor = YamlEditor(content)
  ..update(['app_id'], appId.value ?? appId.values);
File(p.join(Directory.current.path, 'shorebird.yaml')).writeAsStringSync(editor.toString());
```

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
```diff
@@ -0,0 +1,115 @@
+import 'dart:io';
+
+import 'package:collection/collection.dart';
+import 'package:path/path.dart' as p;
+import 'package:platform/platform.dart';
+import 'package:shorebird_cli/src/command.dart';
+
+mixin ShorebirdFlavorMixin on ShorebirdCommand {
```
bryanoltman, line 10: Add class and method docs here

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
```diff
@@ -0,0 +1,115 @@
...
+      if (match != null) {
+        final variant = match.group(1)!.toLowerCase();
+        if (!variant.endsWith('test')) {
+          variants.add(variant);
+        }
+      }
+    }
+
+    final productFlavors = <String>{};
+    for (final variant1 in variants) {
```
bryanoltman: A comment explaining what this is doing and why would be helpful

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
```diff
@@ -0,0 +1,115 @@
...
+        return Directory(location).existsSync();
+      });
+    }
+
+    return null;
+  }
+
+  String? _getJavaPath(String directory, Platform platform) {
+    if (platform.isMacOS) {
+      final locations = [
```
bryanoltman: `candidateLocations` might be a clearer name, but don't feel strongly

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
felangel: Sure I'll update the motivation was to avoid parsing the shorebird.yaml file since this method was meant to just check whether the file exists.

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
felangel: done

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
felangel: Yup good call just added ✅

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
felangel: Done ✅

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
felangel: Sounds good to me, done ✅

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
nilsreichardt: Just out of interest, why is that more efficient?

felangel: I assume Bryan just meant we don't need to create a temporary variable.

bryanoltman: Yeah, sorry, by "this" I meant "checking that the file exists but not parsing it". The one-liner is functionally the same and simply a matter of preference.

nilsreichardt: Ah, thanks 👍

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -30,81 +35,59 @@ class InitCommand extends ShorebirdCommand
...
     }
 
-    final ShorebirdYaml? shorebirdYaml;
-    try {
-      shorebirdYaml = getShorebirdYaml();
-    } catch (_) {
-      progress.fail('Error parsing "shorebird.yaml".');
+    if (hasShorebirdYaml) {
+      logger.err('''
+"shorebird.yaml" already exists.
```
bryanoltman: This assumes that the user is running init intentionally. We might want to add language here to the tune of "you're already set up to use shorebird" to tell the user that this step has been completed.

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -30,81 +35,59 @@ class InitCommand extends ShorebirdCommand
...
+      logger.err('$error');
+      return ExitCode.software.code;
+    }
 
-      final app = apps.firstWhereOrNull(
-        (a) => a.id == shorebirdYaml!.appId.value,
+    final AppId appId;
+    try {
+      final displayName = logger.prompt(
+        '${lightGreen.wrap('?')} How should we refer to this app?',
```
bryanoltman, line 81: Should we standardize the green "?" at the beginning of prompts/confirms? I've noticed this in a few places and haven't been sure whether I should do this as well.

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
```diff
@@ -0,0 +1,120 @@
...
+      p.join(path, 'android', executable),
+      ['app:tasks', '--all', '--console=auto'],
+      runInShell: true,
+      workingDirectory: p.join(path, 'android'),
+      environment: {
+        if (javaPath != null) 'JAVA_HOME': javaPath,
+      },
+    );
+
+    if (result.exitCode != 0) throw Exception('${result.stderr}');
```
bryanoltman: When this errored on Windows, the output was simply

```
$ shorebird init
Exception:
```

We should probably include the command in the error output (or find out where Windows is writing to if not stderr)

bryanoltman (approved):

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -1,19 +1,24 @@
-import 'package:collection/collection.dart';
+import 'dart:io';
```
eseidel, line 1: Apropos of nothing, iirc flutter_tools tries never to include dart:io directly, always some layer above which is mockable.  🤷

packages/shorebird_cli/lib/src/commands/init_command.dart
felangel: Yeah I can do that in a subsequent PR maybe an extension on `Logger` called `confirm`?

packages/shorebird_cli/lib/src/commands/init_command.dart
felangel: Updated the wording and also included a `--force` flag if you want to reinitialize despite an existing `shorebird.yaml`

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
felangel: I updated the error to include both `stdout` and `stderr` to hopefully provide more useful output.

packages/shorebird_cli/lib/src/commands/init_command.dart
bryanoltman: That sounds great, I was just asking so to make sure my prompts are consistent :)

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
```diff
@@ -0,0 +1,120 @@
...
+/// Mixin on [ShorebirdCommand] which exposes methods for extracting
+/// product flavors from the current app.
+mixin ShorebirdFlavorMixin on ShorebirdCommand {
+  /// Return the set of product flavors configured for the app at [path].
+  /// Returns an empty set for apps that do not use product flavors.
+  Future<Set<String>> extractProductFlavors(
+    String path, {
+    Platform platform = const LocalPlatform(),
+  }) async {
+    final executable = platform.isWindows ? 'gradlew.bat' : 'gradlew';
```
eseidel, line 17: I think we are going to want to pull this out into some sort of class at some point (not necessarily now).  We will have a lot of these kinds of binaries we switch on windows and also many we need to find/configure the paths for.

eseidel (approved):

packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart
felangel: yeah I thought about introducing a new class for this but decided we can do that when we need to use this in two places 🤷

packages/shorebird_cli/lib/src/commands/init_command.dart
felangel: We're able to mock most of io using zone overrides but we can definitely refactor if it becomes an issue 👍

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -30,81 +42,64 @@ class InitCommand extends ShorebirdCommand
...
-
-    if (shorebirdYaml != null) {
-      final codePushClient = buildCodePushClient(
-        httpClient: auth.client,
-        hostedUri: hostedUri,
-      );
+    if (hasShorebirdYaml) {
+      logger.err('''
+A "shorebird.yaml" already exists.
+If you want to reinitialize Shorebird, please run "shorebird init --force".''');
```
bryanoltman, line 66:
```suggestion
If you want to reinitialize Shorebird, please run ${lightCyan.wrap('shorebird init --force')}.''');
```

bryanoltman: Although maybe this won't work as cleanly because it's in an err log 🤔

Issue #288, opened 2023-04-13
feat: `shorebird upgrade` should print the new version number.


eseidel:
e.g. moving from 0.0.4 to 0.0.5 just printed:

```
eseidel@Erics-MacBook-Pro build_engine % shorebird upgrade
✓ Checked for updates (0.5s)
✓ Updated successfully. (19ms)
```

But it should also tell me that it moved to 0.0.5.

mateusfccp:
Considering the discussion in #447, is this not planned anymore?

eseidel:
We always auto-upgrade now, so I'm not sure we even need this anymore.

PR #447, opened 2023-05-06
feat: Print current and latest shorebird version while upgrading

rkishan516:
## Description

I have added `Upgrading Shorebird to vX to vY..." log while shorebird is upgrading, 
where X is latest git tag and Y is current git tag,
as asked by @eseidel at issue #288.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/upgrade_command.dart
```diff
@@ -56,6 +56,8 @@ class UpgradeCommand extends ShorebirdCommand with ShorebirdVersionMixin {
       return ExitCode.success.code;
     }
 
+    logger.info('Updating Shorebird to $latestVersion from $currentVersion...');
```
felangel: Thoughts on logging the new version only if the upgrade was successful? E.g “Upgraded to ${newVersion}”

rkishan516: @felangel Can update to that too, but I created like this because while `Flutter Upgrade`, we get same message like <i>Upgrading Flutter from vX to xY</I>.
So, just for sake of consistency for our users.

felangel: Good call seems reasonable to me to keep it consistent 👍

felangel: Can we add a test for this as well?

felangel: Another thing I noticed is `currentVersion` and `latestVersion` are commit hashes so this might not be super helpful for users. We probably need to fix our tags and derive the correct versions before the output is useful for people.

rkishan516:
@felangel have added tag based version. I'll also update tests.

packages/shorebird_cli/lib/src/commands/upgrade_command.dart
```diff
@@ -56,6 +57,30 @@ class UpgradeCommand extends ShorebirdCommand with ShorebirdVersionMixin {
...
+      process,
+      fetchTags: true,
+      workingDirectory: workingDirectory,
+      gitRef: currentVersion,
+    );
+
+    final currentTagVersion =
+        currentVersionTag.frameworkVersionFor(currentVersion);
+
+    logger.info(
```
felangel, line 82: Thanks so much for the contribution but after thinking about it and discussing with @eseidel it seems it'd be simpler to call `flutter --version` after the upgrade. Wdyt?

rkishan516:
@felangel any update ?

rkishan516:
@felangel can we take any action on this MR ?

packages/shorebird_cli/lib/src/commands/upgrade_command.dart
felangel: @rkishan516 what are your thoughts?

felangel:
> @felangel can we take any action on this MR ?

Did you have a chance to see my comment: https://github.com/shorebirdtech/shorebird/pull/447#discussion_r1188954706?

rkishan516:
> > @felangel can we take any action on this MR ?
> 
> Did you have a chance to see my comment: [#447 (comment)](https://github.com/shorebirdtech/shorebird/pull/447#discussion_r1188954706)?

Hey @felangel 

I think with current implementation we should have tag based version system. Because we are quite close to flutter cli itself. So, let suppose in future if we add something that required old or latest version, this will definitely help. Also, we can change the repo link and make it more generic, such that we can also check for flutter versions.

rkishan516:
Closing this since it doesn't make sense anymore

Issue #400, opened 2023-04-27
feat: shorebird release shouldn't allow me to continue w/o internet permission?


eseidel:
Glad that it shows a warning, but it seems like it should be an error?

```
PS C:\Users\micro\Documents\GitHub\win_test> shorebird release
[✗] C:\Users\micro\Documents\GitHub\win_test\android\app\src\main\AndroidManifest.xml is missing the INTERNET permission.
✓ Building release (107.6s)
✓ Fetching apps (0.2s)

What is the version of this release? (1.0.0+1) 1.0.0+1

🚀 Ready to create a new release!

📱 App: win_test (319c0cb4-35f7-4e5e-83bc-6116aff84d7d)
📦 Release Version: 1.0.0+1
🕹️  Platform:  android (arm64, arm32, x86_64)

Would you like to continue? (y/N)
```

eseidel:
related: https://github.com/shorebirdtech/shorebird/issues/259

PR #450, opened 2023-05-07
feat: Add validation issue check before shorebird release

rkishan516:
## Description

I have added validation issue check before release as asked by @eseidel  at issue #400.
Adding check for all validation issues, because we might not want to allow release with any validation issue.
If required we might go with filter some later when required.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release_command.dart
@@ -32,6 +32,7 @@ class ReleaseCommand extends ShorebirdCommand
     super.buildCodePushClient,
     super.validators,
     HashFunction? hashFn,
+    super.blockOnValidationIssues = true,
   }) : _hashFn = hashFn ?? ((m) => sha256.convert(m).toString()) {
     argParser
       ..addOption(
@@ -88,10 +89,10 @@ make smaller updates to your app.
       return ExitCode.noUser.code;
     }
 
-    final validationIssues = await logAndGetValidationIssues();
-    if (validationIssues.isNotEmpty) {
+    final criticalIssues = await logAndGetCriticalIssueCount();
+    if (criticalIssues > 0 && blockOnValidationIssues) {
       logger.err(
-        '''Shorebird release cannot continue until all issues are fixed.''',
+        '''Shorebird $name cannot continue until all issues are fixed.''',
       );
       return ExitCode.config.code;
     }
--- a/packages/shorebird_cli/lib/src/commands/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release_command.dart
@@ -88,7 +88,13 @@ make smaller updates to your app.
       return ExitCode.noUser.code;
     }
 
-    await logValidationIssues();
+    final validationIssues = await logAndGetValidationIssues();
+    if (validationIssues.isNotEmpty) {
+      logger.err(
+        '''Shorebird release cannot continue until all issues are fixed.''',
+      );
+      return ExitCode.config.code;
+    }
 
     final flavor = results['flavor'] as String?;
     final target = results['target'] as String?;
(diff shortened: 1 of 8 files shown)


packages/shorebird_cli/lib/src/shorebird_validation_mixin.dart
```diff
@@ -4,7 +4,7 @@ import 'package:shorebird_cli/src/validators/validators.dart';
 
 mixin ShorebirdValidationMixin on ShorebirdCommand {
   /// Runs [Validator.validate] on all [validators] and writes issues to stdout.
-  Future<void> logValidationIssues() async {
+  Future<List<ValidationIssue>> logValidationIssues() async {
```
felangel: I don't think we should keep this named `logValidationIssues` when it also returns the list.

packages/shorebird_cli/lib/src/commands/release_command.dart
```diff
@@ -82,7 +82,13 @@ make smaller updates to your app.
       return ExitCode.noUser.code;
     }
 
-    await logValidationIssues();
+    final validationIssues = await logValidationIssues();
+    if (validationIssues.isNotEmpty) {
```
felangel: I don't think this is exactly what we want because in this case the only check we care about is the network permission check but `logValidationIssues` performs other checks as well (and that list will likely only continue to grow in the future). Tagging @bryanoltman since he worked on the validators.

rkishan516: Just a thought Should we return Map<Validator, ValidationIssue> from this method ?
Then here we can easily check for Specific validator.

bryanoltman: If we want to map validators to the issues they produce, the return type would be `Map<Validator, List<ValidationIssue>>`, as `validate` produces a list of `ValidationIssue`s.

That said, I think we may want do perform this check differently for `release` than for other commands, for a couple reasons:

1) Not all issues should necessarily block a release (we can probably check `validationIssue.severity` and only block on errors)
2) `AndroidInternetPermissionValidator` checks *all* AndroidManifest.xml files, while the release command should only care about `android/app/src/main/AndroidManifest.xml`. It won't affect a release if, for example, `android/app/src/debug/AndroidManifest.xml` is missing the internet permission.

One approach:

1. Separate the validator execution from the logging of validation issues
1. Update `logValidationIssues` to accept a `List<ValidationIssue>` or `Map<Validator, List<ValidationIssues>>`
1. Add some logic, probably per-command, that determines whether to treat `ValidationIssues` as blockers. 
1. Create a new android manifest validator (or modify the existing one) to only check the manifest xml file used for releases.

rkishan516: Great idea. will add a property on all commands called `blockOnValidationIssues`.
For manifest validator, will try to modify first, if not then create validator of it.

rkishan516: @bryanoltman Few things I have done

- Separated logic for logging validation issue and getting validation issue
- logValidationsIssues takes Map<Validator, List<ValidationIssues>>
- Add `blockOnValidationIssues` flag on shorebird command and getCriticalIssuesCount method
```
    final criticalIssues = await logAndGetCriticalIssueCount();
    if (criticalIssues > 0 && blockOnValidationIssues) {
      logger.err(
        '''Shorebird $name cannot continue until all issues are fixed.''',
      );
      return ExitCode.config.code;
    }
```
- For release, I have added `ReleaseAndroidInternetPermissionValidator` and in release command we are checking like this
```
 final validations = await logAndGetValidationIssues();
    if (validations.isNotEmpty && blockOnValidationIssues) {
      for (final validation in validations.entries) {
        if (validation.key is ReleaseAndroidInternetPermissionValidator) {
          logger.err(
            '''Shorebird $name cannot continue until all issues are fixed.''',
          );
          return ExitCode.config.code;
        }
      }
    }
```

For some reason, last part doesn't look good to me. Can you suggest how to improve this ?

bryanoltman: After thinking about it some more, I decided to go a slightly different direction with https://github.com/shorebirdtech/shorebird/pull/469. Instead of trying to be too specific about which combinations of validators and issues are blocking, I've changed the logic to block if issues with a severity level of `error` are found. Thanks for your help! I've added you as a coauthor to the new PR.

rkishan516: Thanks. Nice approach @bryanoltman. I think we can close this PR.

bryanoltman:
Closing in favor of https://github.com/shorebirdtech/shorebird/pull/469

Issue #414, opened 2023-05-01
fix: `shorebird` build mixin should warn if shorebird.yaml is missing


eseidel:
Right now it just carries on and crashes (which is a separate issue: https://github.com/shorebirdtech/shorebird/issues/413).

```
PS C:\Users\micro\demo\hello_shorebird> shorebird run
Running app...
Flutter assets will be downloaded from https://download.shorebird.dev. Make sure you trust this source!

Using hardware rendering with device sdk gphone64 x86 64. If you notice graphics artifacts, consider enabling software rendering with "--enable-software-rendering".

Launching lib\main.dart on sdk gphone64 x86 64 in release mode...

Running Gradle task 'assembleRelease'...


Flutter assets will be downloaded from https://download.shorebird.dev. Make sure you trust this source!

Running Gradle task 'assembleRelease'...
   33.2s

√  Built build\app\outputs\flutter-apk\app-release.apk (9.0MB).

Installing build\app\outputs\flutter-apk\app-release.apk...     
 1,487ms



Flutter run key commands.

h List all available interactive commands.
c Clear the screen
q Quit (terminate the application on the device).

E/FlutterJNI( 5305): Failed to load shorebird.yaml

E/FlutterJNI( 5305): java.io.FileNotFoundException: flutter_assets/shorebird.yaml
E/FlutterJNI( 5305):    at android.content.res.AssetManager.nativeOpenAsset(Native Method)        
E/FlutterJNI( 5305):    at android.content.res.AssetManager.open(AssetManager.java:904)
E/FlutterJNI( 5305):    at android.content.res.AssetManager.open(AssetManager.java:881)
E/FlutterJNI( 5305):    at io.flutter.embedding.engine.FlutterJNI.init(Unknown Source:51)
E/FlutterJNI( 5305):    at m.d.e(Unknown Source:454)
E/FlutterJNI( 5305):    at io.flutter.embedding.engine.a.<init>(Unknown Source:184)
E/FlutterJNI( 5305):    at io.flutter.embedding.engine.a.<init>(Unknown Source:12)
E/FlutterJNI( 5305):    at io.flutter.embedding.android.e.G(Unknown Source:106)
E/FlutterJNI( 5305):    at io.flutter.embedding.android.e.p(Unknown Source:7)
E/FlutterJNI( 5305):    at io.flutter.embedding.android.d.onCreate(Unknown Source:13)
[...]

PR #451, opened 2023-05-08
fix: Add check for shorebird yaml file in run command

rkishan516:
## Description

I have added shorebird missing file check for run and build command as asked by @eseidel   at issue #414.
After this we will check for shorebird file just after authentication check in run and builds command's

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/build/build_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build/build_command.dart
@@ -1,14 +1,19 @@
 import 'package:collection/collection.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/commands/build/build.dart';
+import 'package:shorebird_cli/src/shorebird_config_mixin.dart';
+import 'package:shorebird_cli/src/validators/shorebird_yaml_validator.dart';
+import 'package:shorebird_cli/src/validators/validators.dart';
 
 /// {@template build_command}
 /// `shorebird build`
 /// Build a new release of your application.
 /// {@endtemplate}
-class BuildCommand extends ShorebirdCommand {
+class BuildCommand extends ShorebirdCommand with ShorebirdConfigMixin {
   /// {@macro build_command}
   BuildCommand() {
+    validators = _allValidators(baseValidators: validators);
+
     addSubcommand(BuildAarCommand());
     addSubcommand(BuildApkCommand());
     addSubcommand(BuildAppBundleCommand());
--- a/packages/shorebird_cli/lib/src/commands/build/build_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build/build_command.dart
@@ -1,48 +1,22 @@
-import 'package:collection/collection.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/commands/build/build.dart';
-import 'package:shorebird_cli/src/shorebird_config_mixin.dart';
-import 'package:shorebird_cli/src/validators/validators.dart';
 
 /// {@template build_command}
 /// `shorebird build`
 /// Build a new release of your application.
 /// {@endtemplate}
-class BuildCommand extends ShorebirdCommand with ShorebirdConfigMixin {
+class BuildCommand extends ShorebirdCommand {
   /// {@macro build_command}
   BuildCommand() {
-    validators = _allValidators(baseValidators: validators);
-
     addSubcommand(BuildAarCommand());
     addSubcommand(BuildApkCommand());
     addSubcommand(BuildAppBundleCommand());
     addSubcommand(BuildIpaCommand());
   }
 
-  final List<Validator> _buildValidators = [
-    ShorebirdYamlValidator(),
-    AndroidInternetPermissionValidator(),
-  ];
-
   @override
   String get description => 'Build a new release of your application.';
 
   @override
   String get name => 'build';
-
-  /// Creates a list that is the union of [baseValidators] and
-  /// [_buildValidators].
-  List<Validator> _allValidators({
-    required List<Validator> baseValidators,
-  }) {
-    final missingValidators = _buildValidators
-        .where(
-          (runValidator) => baseValidators.none(
-            (baseValidator) => baseValidator.id == runValidator.id,
-          ),
-        )
-        .toList();
-
-    return baseValidators + missingValidators;
-  }
 }
--- a/packages/shorebird_cli/lib/src/commands/build/build_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build/build_command.dart
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/commands/build/build.dart';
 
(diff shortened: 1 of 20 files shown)


packages/shorebird_cli/lib/src/commands/build/build_apk_command.dart
```diff
@@ -49,6 +49,13 @@ class BuildApkCommand extends ShorebirdCommand
       return ExitCode.noUser.code;
     }
 
+    if (!hasShorebirdYaml) {
+      logger.err(
```
felangel: I wonder whether this should be another validation check? cc @bryanoltman

bryanoltman: I think that's a great idea

rkishan516: Sure, Will update check with validation.

rkishan516:
@bryanoltman @felangel I have added validator for checking shorebird yaml.

rkishan516:
@bryanoltman @felangel any update, Should I added tests here ?

bryanoltman:
We'll definitely need tests. It also looks like a few of the existing tests are broken (see https://github.com/shorebirdtech/shorebird/actions/runs/4966615389/jobs/8888149349?pr=451)

rkishan516:
> We'll definitely need tests. It also looks like a few of the existing tests are broken (see https://github.com/shorebirdtech/shorebird/actions/runs/4966615389/jobs/8888149349?pr=451)

Yaa, I just wanted this to get reviewed.

rkishan516:
@bryanoltman I have added all required test + fixed failing tests.
Sorry for the delay.

rkishan516:
@bryanoltman @felangel can we take any action on this MR ?

packages/shorebird_cli/lib/src/validators/shorebird_yaml_validator.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:shorebird_cli/src/shorebird_process.dart';
+import 'package:shorebird_cli/src/validators/validators.dart';
+
+class ShorebirdYamlValidator extends Validator {
+  ShorebirdYamlValidator({required this.hasShorebirdYaml});
+  // Adding as Function
+  /// Returns true if the project has a shorebird.yaml file.
+  final bool Function() hasShorebirdYaml;
```
felangel: Why are we injecting a function to do the validation? Thoughts on performing the validation directly in the validator? We can still unit test it and it removes one level of indirection imo. Let me know what you think!

packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -1,10 +1,13 @@
 import 'dart:convert';
 
+import 'package:collection/collection.dart';
 import 'package:mason_logger/mason_logger.dart';
 import 'package:shorebird_cli/src/auth_logger_mixin.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/shorebird_config_mixin.dart';
 import 'package:shorebird_cli/src/shorebird_validation_mixin.dart';
+import 'package:shorebird_cli/src/validators/shorebird_yaml_validator.dart';
```
felangel:
```suggestion
```
Minor but we should update the barrel file to include the `shorebird_yaml_validator.dart` and this additional import won't be necessary

packages/shorebird_cli/lib/src/commands/build/build_command.dart
```diff
@@ -1,21 +1,47 @@
+import 'package:collection/collection.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/commands/build/build.dart';
+import 'package:shorebird_cli/src/shorebird_config_mixin.dart';
+import 'package:shorebird_cli/src/validators/shorebird_yaml_validator.dart';
```
felangel:
```suggestion
```

Same here

packages/shorebird_cli/lib/src/commands/build/build_command.dart
```diff
@@ -1,21 +1,47 @@
...
+class BuildCommand extends ShorebirdCommand with ShorebirdConfigMixin {
   /// {@macro build_command}
   BuildCommand({required super.logger}) {
+    validators = _allValidators(baseValidators: validators);
     addSubcommand(BuildApkCommand(logger: logger));
     addSubcommand(BuildAppBundleCommand(logger: logger));
     addSubcommand(BuildIpaCommand(logger: logger));
   }
 
+  late final List<Validator> _buildValidators = [
```
felangel: Why is this `late`?

packages/shorebird_cli/lib/src/commands/build/build_command.dart
```diff
@@ -1,21 +1,47 @@
...
+    ShorebirdYamlValidator(hasShorebirdYaml: () => hasShorebirdYaml),
```
felangel:
```suggestion
    ShorebirdYamlValidator(),
```

packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -83,4 +93,20 @@ class RunCommand extends ShorebirdCommand
 
     return flutter.exitCode;
   }
+
+  /// Creates a list that is the union of [baseValidators] and
+  /// [_runValidators].
+  List<Validator> _allValidators({
```
felangel: Can we avoid duplicating this code? I think it'd be nice to refactor `validators` to be a `Set` and each command can add it's own custom validators to the default set without having to compute the union like this. Thoughts?

packages/shorebird_cli/lib/src/validators/shorebird_yaml_validator.dart
rkishan516: Yup, make sense. I think we should do it directly inside validator. I'll update

packages/shorebird_cli/lib/src/commands/run_command.dart
rkishan516: Yaa, making it a `Set` will solve this. Also we have validators on command level only. So, I think we should do this change.

rkishan516: I'll update it accordingly

packages/shorebird_cli/lib/src/commands/build/build_command.dart
rkishan516: My bad, should not be late.

packages/shorebird_cli/lib/src/commands/build/build_command.dart
rkishan516: Will update it

packages/shorebird_cli/lib/src/commands/run_command.dart
rkishan516: Will update it

packages/shorebird_cli/lib/src/commands/run_command.dart
rkishan516: @felangel Have updated to Set. but `ensure_build` test is failing. Can you check that ?

felangel: Did you re-run `build_runner`?

rkishan516: Yes, I did.

felangel:
Closing this since run is deprecated, thanks!

PR #452, opened 2023-05-08
docs: make artifact proxy `/` return an explainer

eseidel:
We've seen a number of requests to /, so figured might as well return an explainer there.


felangel (approved):
LGTM ✅

PR #468, opened 2023-05-11
feat: add a `cutler print-versions` command

eseidel:
I also wrote up what I think our branching strategy could look like.

### DIFF
--- a/packages/cutler/FORKING_FLUTTER.md
+++ b/packages/cutler/FORKING_FLUTTER.md
@@ -3,6 +3,28 @@ These are our old docs for our manual update process.
 `cutler` replaces some of this, but not all of it. Keeping these until
 `cutler` is fully functional.
 
+## Repository structure
+
+We maintain forks of:
+* flutter/flutter
+* flutter/engine
+* flutter/buildroot
+
+We keep our forked changes on the `main` branch of each repo which we rebase
+periodically on top of `main` from the upstream repo.
+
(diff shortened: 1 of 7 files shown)


eseidel:
The changes here are rather small.  It's mostly restructuring `cutler` to use CommandRunner and shoving the previous implementation into a RebaseCommand.

I also updated the FORKING_FLUTTER with `cutler` specific instructions (I thought I had already posted that for review, but I guess not?)

And then I added the bit about a branching strategy to FORKING_FLUTTER.md.

packages/cutler/FORKING_FLUTTER.md
```diff
@@ -3,90 +3,172 @@ These are our old docs for our manual update process.
...
-ec975089acb540fc60752606a3d3ba809dd1528b
+Currently the tool assumes you're using our internal repository `_shorebird`
+to check out `shorebird`.  Example:
+```
+cd $HOME/Documents/GitHub
+git clone https://github.com/shorebirdtech/_shorebird
+```
+To check out the engine, you should follow:
+https://github.com/shorebirdtech/updater/blob/main/BUILDING_ENGINE.md
+it will result in an `engine` directory in the same directory as `_shorebird`.
```
bryanoltman:
```suggestion
which will result in an `engine` directory in the same directory as `_shorebird`.
```

bryanoltman (approved):

felangel (approved):

PR #474, opened 2023-05-11
fix(shorebird_cli): make ShorebirdJavaMixin work on Windows

bryanoltman:
## Description

Without this change, `shorebird release` fails with the error: 

```
✗ Exception: Failed to extract version name from app bundle: 'java' is not recognized as an internal or external command, operable program or batch file.
```

I'm not sure if this is the most elegant or correct fix, but it works. We'll likely want to reference what flutter_tool does to handle this.

I've verified this fix on my Windows 11 laptop and my M2 MBP. 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
@@ -13,7 +13,7 @@ mixin ShorebirdJavaMixin on ShorebirdCommand {
 
     final javaPath = getJavaPath(platform);
     if (javaPath == null) return null;
-    return p.join(javaPath, 'java.exe');
+    return p.join(javaPath, 'bin', 'java.exe');
   }
 
   String? getJavaPath([Platform platform = const LocalPlatform()]) {
@@ -36,8 +36,8 @@ mixin ShorebirdJavaMixin on ShorebirdCommand {
     }
 
     final candidateLocations = [
-      p.join(androidStudioPath, 'jbr', 'bin'),
-      p.join(androidStudioPath, 'jre', 'bin'),
+      p.join(androidStudioPath, 'jbr'),
+      p.join(androidStudioPath, 'jre'),
     ];
 
     return candidateLocations.firstWhereOrNull(
--- a/packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
@@ -11,12 +11,12 @@ mixin ShorebirdJavaMixin on ShorebirdCommand {
   String? getJavaExecutable([Platform platform = const LocalPlatform()]) {
     if (!platform.isWindows) return 'java';
 
-    final javaPath = getJavaPath(platform);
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
```diff
@@ -8,6 +8,14 @@ import 'package:shorebird_cli/src/command.dart';
 /// Mixin on [ShorebirdCommand] which exposes methods
 /// for determining the java path.
 mixin ShorebirdJavaMixin on ShorebirdCommand {
+  String? getJavaExecutable([Platform platform = const LocalPlatform()]) {
```
felangel, line 11: I’m confused. Is what’s on main working? I thought we verified main and these changes were just reorganization 🤔

packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
```diff
@@ -28,8 +36,8 @@ mixin ShorebirdJavaMixin on ShorebirdCommand {
     }
 
     final candidateLocations = [
-      p.join(androidStudioPath, 'jbr'),
-      p.join(androidStudioPath, 'jre'),
+      p.join(androidStudioPath, 'jbr', 'bin'),
```
felangel: Are these changes needed? I didn’t change this logic and I thought we verified that main worked.

packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
bryanoltman: What's on `main` does work, but on `main`, this code is only used to set `JAVA_HOME` as an environment variable when we invoke gradle. For that gradle invocation, we have to specify the full path to the executable (see https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/shorebird_flavor_mixin.dart#L17) like we're doing here.

packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
bryanoltman: This was a mistake - just reverted it. These are the correct values for `JAVA_HOME`.

packages/shorebird_cli/lib/src/shorebird_java_mixin.dart
bryanoltman: My brain is shutting down for the night, but the tl;dr is:

1. Gradle needs the `JAVA_HOME` environment variable to be set. JAVA_HOME is a path to the root of a JDK installation, _not_ to the JDK's bin folder where the actual java executable lives.
2. `process.run` needs either a path to the executable or for the executable to be on the path.

I've renamed `getJavaPath` to make this distinction a bit clearer. We can chat more tomorrow in the very likely event that I'm not making sense here.

felangel (approved):
LGTM thanks so much for validating windows and catching this!

PR #476, opened 2023-05-12
feat(shorebird_cli): `shorebird patch` validates flutter revision

felangel:
## Description

- feat(shorebird_cli): `shorebird patch` validates flutter revision

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
@@ -5,6 +5,7 @@ import 'package:path/path.dart' as p;
 import 'package:pubspec_parse/pubspec_parse.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/config/config.dart';
+import 'package:shorebird_cli/src/shorebird_environment.dart';
 import 'package:yaml/yaml.dart';
 import 'package:yaml_edit/yaml_edit.dart';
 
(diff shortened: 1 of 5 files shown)


packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
```diff
@@ -145,26 +145,25 @@ This can cause unexpected behavior if you are switching between the tools and th
...
+    return _getFlutterVersion(
+      process: process,
+      useVendedFlutter: false,
+    );
+  }
 
   Future<String> _getFlutterVersion({
     required ShorebirdProcess process,
-    required bool checkPathFlutter,
+    bool useVendedFlutter = true,
```
felangel, line 137: Adjusted this since I was looking at this code and felt it was a bit confusing but happy to revert if folks prefer the old naming.

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
```diff
@@ -105,4 +106,16 @@ app_id:
...
+    final result = await process.run(
+      'git',
+      ['rev-parse', 'HEAD'],
+      workingDirectory: ShorebirdEnvironment.flutterDirectory.path,
+    );
+    if (result.exitCode != 0) {
+      throw Exception('Unable to determine flutter revision: ${result.stderr}');
+    }
+    return (result.stdout as String).trim();
+  }
```
bryanoltman, line 120: We have similar logic in shorebird/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart. I think this implementation is a bit more robust, so we probably want to find a way to use this there (`_flutterDirectoryTracksCorrectRevision` is the function name)

bryanoltman (approved):

packages/shorebird_cli/lib/src/shorebird_config_mixin.dart
felangel: Are you referring to `_getFlutterVersion`? I saw it was similar but it seemed to be version-based instead of hash-based.

PR #492, opened 2023-05-15
refactor: upgrade to Dart 3

felangel:
## Description

- refactor: upgrade to Dart 3

## Type of Change

- [X] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_process.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_process.dart
@@ -30,7 +30,7 @@ class ShorebirdProcess {
   final ProcessWrapper processWrapper;
   final EngineConfig engineConfig;
 
-  Future<ProcessResult> run(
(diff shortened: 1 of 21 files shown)


packages/shorebird_cli/lib/src/shorebird_process.dart
```diff
@@ -113,23 +113,40 @@ class ShorebirdProcess {
   }
 }
 
+class ShorebirdProcessResult {
```
eseidel, line 116: Why?  Was it sealed and thus couldn't be mocked?

eseidel (approved):

packages/shorebird_cli/lib/src/shorebird_process.dart
felangel: It’s now final so it can’t be implemented.

PR #521, opened 2023-05-22
feat(shorebird_cli): upload aab as part of release

### LINKED ISSUE #519: teach command line to upload full aab

bryanoltman:
## Description

Upload the full `.aab` file when running `shorebird release`.

Fixes https://github.com/shorebirdtech/shorebird/issues/519

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/release_command.dart
```diff
@@ -250,6 +250,22 @@ ${summary.join('\n')}
 
     createArtifactProgress.complete();
 
+    final uploadAppBundleProgress = logger.progress('Uploading app bundle');
```
felangel: rather than splitting this out since it technically is just "another artifact", wdyt of moving this as part of the `createArtifactProgress` step instead of defining a separate progress step? It could still run last but I don't think there's much value is letting the user know we're both uploading artifacts and and app bundle (since they are all artifacts as far as the end user is concerned imo). Let me know what you think.

bryanoltman: Makes sense, done.

packages/shorebird_cli/lib/src/commands/release_command.dart
```diff
@@ -248,6 +248,19 @@ ${summary.join('\n')}
       }
     }
 
+    try {
+      await codePushClient.createReleaseArtifact(
+        releaseId: release.id,
+        artifactPath: bundlePath,
+        arch: 'universal',
```
eseidel: I might call it "multi" or "aab", we don't know what archs are in it.  We could tag it with the archs if we wanted. 🤷   e.g. 'arm-arm64-x86'

bryanoltman: Renamed to "aab"

felangel (approved):

Issue #518, opened 2023-05-18
cli / server support for iOS


eseidel:
The CLI and server side of iOS support.

- [X] `shorebird build ipa`
- [x] `shorebird release`
- [x] `shorebird patch`

PR #534, opened 2023-05-23
feat(shorebird_cli): add `shorebird build ipa`

felangel:
## Description

- feat(shorebird_cli): add `shorebird build ipa` (part of #518)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
@@ -1,0 +1,104 @@
+import 'dart:io';
+
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/auth_logger_mixin.dart';
+import 'package:shorebird_cli/src/command.dart';
+import 'package:shorebird_cli/src/shorebird_build_mixin.dart';
+import 'package:shorebird_cli/src/shorebird_config_mixin.dart';
+import 'package:shorebird_cli/src/shorebird_validation_mixin.dart';
+
+/// {@template build_ipa_command}
+/// `shorebird build ipa`
+/// Builds an .xcarchive and optionally .ipa for an iOS app to be generated for
+/// App Store submission.
+/// {@endtemplate}
+class BuildIpaCommand extends ShorebirdCommand
+    with
+        AuthLoggerMixin,
+        ShorebirdValidationMixin,
+        ShorebirdConfigMixin,
+        ShorebirdBuildMixin {
+  /// {@macro build_ipa_command}
+  BuildIpaCommand({required super.logger, super.auth, super.validators}) {
+    argParser
+      ..addOption(
+        'target',
+        abbr: 't',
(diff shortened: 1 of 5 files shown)


packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
```diff
@@ -0,0 +1,104 @@
...
+  String get name => 'ipa';
+
+  @override
+  Future<int> run() async {
+    if (!auth.isAuthenticated) {
+      printNeedsAuthInstructions();
+      return ExitCode.noUser.code;
+    }
+
+    final validationIssues = await runValidators();
```
bryanoltman, line 54: This is a fun wrinkle – the default validators list only contains `AndroidInternetPermissionValidator`, which doesn't apply here. I think we might need to have something similar in the app's entitlements file and/or its Info.plist.

packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
```diff
@@ -0,0 +1,104 @@
...
+
+  @override
+  String get description =>
+      '''Builds an .xcarchive and optionally .ipa for an iOS app to be generated for App Store submission.''';
+
+  @override
+  String get name => 'ipa';
+
+  @override
+  Future<int> run() async {
```
bryanoltman, line 48: Do we want to verify that the current OS can build ipas? I'm sure `flutter build ipa` will spit up appropriately so maybe we don't really need to.

packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
```diff
@@ -0,0 +1,104 @@
...
+        codesign: codesign,
+      );
+    } on ProcessException catch (error) {
+      buildProgress.fail('Failed to build: ${error.message}');
+      return ExitCode.software.code;
+    }
+
+    buildProgress.complete();
+
+    const xcarchivePath = './build/ios/archive/Runner.xcarchive';
```
bryanoltman, line 83: Would probably be good to do a `p.join` here, even though this should work anywhere you can build an ipa

bryanoltman (approved):
LGTM, although we'll need to figure out validators at some point

bryanoltman (approved):
LGTM, although we'll need to figure out validators at some point

packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
felangel: Yeah I was planning to leverage the error handling built into the flutter tool but happy to adjust this

packages/shorebird_cli/lib/src/commands/build/build_ipa_command.dart
felangel: Sounds good, will adjust in the next PR 👍

PR #535, opened 2023-05-24
fix(shorebird_cli): canonicalize build paths

felangel:
## Description

- refactor(shorebird_cli): canonicalize build paths

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/build/build_apk_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build/build_apk_command.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 
 import 'package:mason_logger/mason_logger.dart';
+import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/auth_logger_mixin.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/shorebird_build_mixin.dart';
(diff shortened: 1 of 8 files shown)


packages/shorebird_cli/lib/src/commands/build/build_apk_command.dart
```diff
@@ -68,8 +69,10 @@ class BuildApkCommand extends ShorebirdCommand
     buildProgress.complete();
 
     final apkPath = flavor != null
-        ? './build/app/outputs/apk/$flavor/release/app-$flavor-release.apk'
-        : './build/app/outputs/apk/release/app-release.apk';
+        ? p.normalize(
+            './build/app/outputs/apk/$flavor/release/app-$flavor-release.apk',
+          )
+        : p.normalize('./build/app/outputs/apk/release/app-release.apk');
```
bryanoltman: TIL about normalize.

Does `normalize` handle different path separators? Based on the docs, it looks like it just removes nested `..`s, etc.

felangel: Ah good catch...didn't want to use canonicalize since it returns absolute paths :/

felangel: Adjusted to use `join`

packages/shorebird_cli/test/src/commands/build/build_apk_command_test.dart
```diff
@@ -127,7 +127,7 @@ void main() {
         () => logger.info(
           '''
 📦 Generated an apk at:
-${lightCyan.wrap("./build/app/outputs/apk/release/app-release.apk")}''',
```
bryanoltman, line 130: I think these will fail on Windows.

felangel: yeah good call will update the tests (although I'm sure that's not all that will fail on windows 🤣)

felangel: done 👍

bryanoltman (approved):

PR #538, opened 2023-05-24
feat(shorebird_cli): alert user of non-patchable changes

### LINKED ISSUE #385: feat: shorebird should detect/warn about non-dart app changes
TTs have run into this.  Both by changing flutter_assets as well as changing java/kotlin code.

We should detect that there have been changes to the app that we do not support patching and warn users.

bryanoltman:
## Description

Updates `shorebird patch` to compare the contents of the aab generated during the patch build with the contents of the aab generated by the corresponding release.

When a native code change is detected (i.e., a change to a `.dex` file), the patch command fails.
When an asset change is detected, the patch command warns the user and asks if they are sure they want to continue.

Fixes https://github.com/shorebirdtech/shorebird/issues/385

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/aab/aab_differ.dart
```diff
@@ -0,0 +1,105 @@
...
+///      - This seems to change with every build, regardless of whether any code
+///        or assets were changed.
+///
+/// See https://developer.android.com/guide/app-bundle/app-bundle-format for
+/// reference.
+class AabDiffer {
+  /// Match files that change when Dart code changes.
+  final _dartChangePatters = [
+    RegExp(r'.+libapp\.so$'),
+    RegExp(r'.+libflutter\.so$'),
```
felangel: Wouldn't the basename always be `libapp.so` and `libflutter.so`?

bryanoltman: Yep. Even thought this would be pretty trivial to handle without a RegExp, I opted to use them in all of the matchers since we have some cases (and possible more that aren't handled here) that are not as straightforward to detect.

felangel: but is this regexp correct? Wouldn't this match `mylibapp.so` for example?

bryanoltman: Ahhhh I see what you mean. I can make this a bit more sophisticated.

packages/shorebird_cli/lib/src/aab/aab_differ.dart
```diff
@@ -0,0 +1,105 @@
...
+    final mfEntries2 = MfReader.parse(mfContents2).toSet();
+    return mfEntries1.difference(mfEntries2).map((entry) => entry.name).toSet();
+  }
+
+  /// Returns a set of difference types detected between the aabs at [aabPath1]
+  /// and [aabPath2].
+  Set<AabDifferences> aabContentDifferences(String aabPath1, String aabPath2) {
+    final fileDifferences = aabFileDifferences(aabPath1, aabPath2);
+
+    final hasAssetChanges = fileDifferences.any((file) {
```
felangel: minor but could we refactor to loop through the files once to determine what has changed?

packages/shorebird_cli/lib/src/commands/patch_command.dart
```diff
@@ -310,8 +324,42 @@ https://github.com/shorebirdtech/shorebird/issues/472
...
+    final contentDiffs = _aabDiffer.aabContentDifferences(
+      releaseAabPath,
+      bundlePath,
+    );
+
+    logger.detail('content diffs detected: $contentDiffs');
+
+    if (contentDiffs.contains(AabDifferences.native)) {
+      logger.err(
+        '''Your aab contains native changes, which cannot be applied in a patch. Please create a new release or revert these changes.''',
```
felangel: Can we list the native changes? I'm thinking developers would likely want to know which changes to revert.

packages/shorebird_cli/lib/src/aab/aab_differ.dart
bryanoltman: Fixed, lmk what you think

packages/shorebird_cli/lib/src/commands/patch_command.dart
```diff
@@ -310,8 +324,42 @@ https://github.com/shorebirdtech/shorebird/issues/472
...
+      );
+      return ExitCode.software.code;
+    }
+
+    if (contentDiffs.contains(AabDifferences.assets)) {
+      final shouldContinue = logger.confirm(
+        '''Your aab contains asset changes, which will not be included in the patch. Continue anyways?''',
```
felangel: Can we list the asset changes?

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: We would need to be able to parse `.dex` files for this, which seems like a heavy lift https://source.android.com/docs/core/runtime/dex-format

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: This seems doable, but there is some subtlety to it. Types of asset changes that will need to be handled differently:

1. Adding an asset (base/assets/flutter_assets/AssetManifest.json changes)
2. Changing an existing asset (the asset file has a different hash)
3. Changing a non-Flutter asset (a layout.xml or a .png that Flutter is not aware of)

I think this would be useful but I'd opt to do this in a followup PR.

packages/shorebird_cli/lib/src/commands/patch_command.dart
felangel: Ah yeah makes sense. Created https://github.com/shorebirdtech/shorebird/issues/539 and maybe we can revisit in the future?

packages/shorebird_cli/lib/src/commands/patch_command.dart
felangel: Yeah it would be super useful imo and I'm fine with it being a follow-up PR 👍

packages/shorebird_cli/test/src/aab/mf_reader_test.dart
```diff
@@ -0,0 +1,77 @@
...
+        name: 'base/assets/flutter_assets/shorebird.yaml',
+        sha256Digest: 'WkS2WF7aQZuQ91vLhCXq0vXjRx702kn5K6nAYwQiDec=',
+      ),
+      const MfEntry(
+        name: 'base/dex/classes.dex',
+        sha256Digest: 'wCxl8B3GnKZdBjEOIaW/AhLmIQYXIlYwVC9VRisYGsw=',
+      ),
+    };
+
+    test('parses content string into MFEntrys', () {
```
felangel:
```suggestion
    test('parses content string into MFEntries', () {
```

packages/shorebird_cli/test/src/aab/mf_reader_test.dart
```diff
@@ -0,0 +1,77 @@
...
+      final entries = MfReader.parse(fileContent);
+      expect(entries.toSet(), expectedEntries);
+    });
+
+    test('parses MFEntrys from file', () {
```
felangel:
```suggestion
    test('parses MFEntries from file', () {
```

packages/shorebird_cli/lib/src/commands/patch_command.dart
eseidel: I might say "The Android App Bundle" instead of "Your aab" 🤷

packages/shorebird_cli/lib/src/commands/patch_command.dart
```diff
@@ -293,6 +297,16 @@ https://github.com/shorebirdtech/shorebird/issues/472
         return ExitCode.software.code;
       }
     }
+
+    final ReleaseArtifact releaseAabArtifact;
+    try {
+      releaseAabArtifact =
+          await codePushClient.getAabArtifact(releaseId: release.id);
+    } catch (error) {
+      fetchReleaseArtifactProgress.fail('$error');
```
felangel: Do we want to error? Wouldn't this break existing users? I would think we would just ignore apps which haven't uploaded their app bundles yet for backward compatibility.

packages/shorebird_cli/lib/src/commands/patch_command.dart
eseidel: Similarly, I might say "The Android App Bundle appears to contain Kotlin or Java changes which cannot be applied via a patch.  Please create a new release or revert those changes to create a patch.  If you believe you're seeing this in error, please reach out to us for support.  shorebird.dev/support". (That url doesn't exist yet, but we should make something similar.)

packages/shorebird_cli/lib/src/aab/aab_differ.dart
bryanoltman: I've updated this code since this comment was written, but as it stands now, I worry that we'd losing quite a bit of readability for a negligible performance improvement. Happy to be shown otherwise though!

packages/shorebird_cli/test/src/aab/mf_reader_test.dart
bryanoltman: Updated, wanted to keep "MfEntry" in there as it's the class name.

packages/shorebird_cli/test/src/aab/mf_reader_test.dart
bryanoltman: Updated, wanted to keep "MfEntry" in there as it's the class name.

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: Ahhh good catch! Will fix.

bryanoltman: Fixed.

packages/shorebird_cli/lib/src/commands/patch_command.dart
```diff
@@ -293,6 +297,16 @@ https://github.com/shorebirdtech/shorebird/issues/472
         return ExitCode.software.code;
       }
     }
+
+    ReleaseArtifact? releaseAabArtifact;
+    try {
+      releaseAabArtifact =
+          await codePushClient.getAabArtifact(releaseId: release.id);
+    } catch (error) {
+      // Do nothing for now, not all releases will have an associated aab
```
felangel, line 309: Maybe we should include a todo to introduce an error in the future once most of our releases have app bundles?

packages/shorebird_cli/test/src/commands/patch_command_test.dart
```diff
@@ -586,6 +622,135 @@ Please create a release using "shorebird release" and try again.
...
+      setUpTempArtifacts(tempDir);
+      final exitCode = await IOOverrides.runZoned(
+        command.run,
+        getCurrentDirectory: () => tempDir,
+      );
+
+      expect(exitCode, ExitCode.software.code);
+      verify(
+        () => logger.err(
+          '''Your aab contains native changes, which cannot be applied in a patch. Please create a new release or revert these changes.''',
```
felangel:
```suggestion
          '''Your Android App Bundle contains native changes, which cannot be applied in a patch. Please create a new release or revert these changes.''',
```

packages/shorebird_cli/test/src/commands/patch_command_test.dart
```diff
@@ -586,6 +622,135 @@ Please create a release using "shorebird release" and try again.
...
+        command.run,
+        getCurrentDirectory: () => tempDir,
+      );
+
+      expect(exitCode, ExitCode.success.code);
+      verify(
+        () => logger.confirm(
+          any(
+            that: contains(
+              '''Your aab contains asset changes, which will not be included in the patch. Continue anyways?''',
```
felangel:
```suggestion
              '''Your Android App Bundle contains asset changes, which will not be included in the patch. Continue anyways?''',
```

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -338,6 +338,15 @@ class CodePushClient {
         .toList();
   }
 
+  /// Get a release artifact for a specific [releaseId] with architecture "aab"
+  /// and platform "android".
+  Future<ReleaseArtifact> getAabArtifact({required int releaseId}) =>
```
felangel: I feel we should either remove this or also add a `createAabArtifact` for consistency/symmetry. I personally prefer to remove this since I don't think it improves the readability a ton and it's a direct proxy to an existing public method.

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: Done.

bryanoltman: And filed https://github.com/shorebirdtech/shorebird/issues/542

packages/shorebird_cli/test/src/commands/patch_command_test.dart
bryanoltman: Fixed.

packages/shorebird_cli/test/src/commands/patch_command_test.dart
bryanoltman: Fixed.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
bryanoltman: Yeah, that makes sense. Removed.

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: Done.

packages/shorebird_cli/lib/src/commands/patch_command.dart
bryanoltman: Done, and filed https://github.com/shorebirdtech/website/issues/15 to add a support page.

felangel (approved):
LGTM ✅

PR #554, opened 2023-05-26
fix(shorebird_cli): don't fail if a release artifact has already been uploaded

### LINKED ISSUE #553: chore(shorebird_code_push_client): add detail to error messages when requests fail
**Description**

We often just throw an exception containing the error test or "an unknown error occurred" instead of providing context about things like which call failed, etc.

There may also be associated work to improve this messaging on the backend.

bryanoltman:
## Description

`shorebird release` currently fails if it detects that an artifact already exists for a given architecture. This can cause problems if a user's connection is unstable and only some artifact uploads succeed, leaving them in a state where they are unable to complete their release.

This change updates the release artifact upload logic to log when 409s (HttpStatus.conflict) are received instead of aborting the release.

Fixes https://github.com/shorebirdtech/shorebird/issues/553

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_client/lib/src/code_push_client.dart
+++ b/packages/shorebird_code_push_client/lib/src/code_push_client.dart
@@ -21,12 +21,12 @@ class CodePushException implements Exception {
   String toString() => '$message${details != null ? '\n$details' : ''}';
 }
 
-/// {@template resource_conflict_exception}
+/// {@template code_push_conflict_exception}
 /// Exception thrown when a 409 response is received.
 /// {@endtemplate}
-class ResourceConflictException extends CodePushException {
-  /// {@macro resource_conflict_exception}
-  ResourceConflictException({required super.message, super.details});
+class CodePushConflictException extends CodePushException {
+  /// {@macro code_push_conflict_exception}
+  const CodePushConflictException({required super.message, super.details});
 }
 
 /// {@template code_push_client}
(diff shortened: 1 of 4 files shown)


packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -139,7 +147,13 @@ class CodePushClient {
     final response = await _httpClient.send(request);
     final body = await response.stream.bytesToString();
 
-    if (response.statusCode != HttpStatus.ok) throw _parseErrorResponse(body);
+    if (response.statusCode == HttpStatus.conflict) {
+      throw ResourceConflictException(
```
felangel: Thoughts on not overwriting the message from the server? We're duplicating the same message in multiple places and the CLI isn't using the message for anything at the moment so it seems unnecessary.

felangel: Thoughts on adjusting to:

```dart
class CodePushConflict extends CodePushException {
  CodePushConflict({super.message, super.details});
}
```

And adjusting `_parseErrorResponse` to handle conflicts in a generic way?

felangel: See https://github.com/shorebirdtech/shorebird/pull/561

felangel (approved):

PR #565, opened 2023-05-30
chore(shorebird_cli): add logging around artifact creation

bryanoltman:
## Description

Add logs when creating individual artifacts in `shorebird release` and `shorebird patch`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch_command.dart
@@ -350,7 +350,7 @@ https://github.com/shorebirdtech/shorebird/issues/472
             bundlePath,
           );
 
-    logger.detail('content diffs detected: $contentDiffs');
+    logger.detail('aab content differences: $contentDiffs');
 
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/patch_command.dart
```diff
@@ -398,6 +398,7 @@ If you believe you're seeing this in error, please reach out to us for support a
         archMetadata.path,
         'libapp.so',
       );
+      logger.detail('Creating artifact for $patchArtifactPath');
```
bryanoltman, line 401: @felangel we've been using `[DOMAIN]` prefixes elsewhere, but I'm not sure what a sensible one would be for the top-level command. Maybe we don't need one? Thoughts?

felangel: Yeah I guess we don't need one here

felangel (approved):

PR #578, opened 2023-06-01
feat: Give cutler some sane default behavior

eseidel:
I found it frustrating that cutler would always ask me to be explicit both about the root directory as well as print-versions about what revision I wanted, so I fixed both to have sane defaults.

Also added mason_logger for logging instead of using print everywhere.  I did not fix existing uses of print for now.

### DIFF
--- a/packages/cutler/bin/cutler.dart
+++ b/packages/cutler/bin/cutler.dart
@@ -5,17 +5,20 @@ import 'package:args/command_runner.dart';
 import 'package:cutler/commands/commands.dart';
 import 'package:cutler/config.dart';
 import 'package:cutler/model.dart';
+import 'package:mason_logger/mason_logger.dart';
+import 'package:path/path.dart' as p;
 
 class Cutler extends CommandRunner<int> {
-  Cutler() : super('cutler', 'A tool for maintaining forks of Flutter.') {
-    addCommand(RebaseCommand());
-    addCommand(PrintVersionsCommand());
+  Cutler({Logger? logger})
+      : _logger = logger ?? Logger(),
+        super('cutler', 'A tool for maintaining forks of Flutter.') {
+    addCommand(RebaseCommand(logger: _logger));
+    addCommand(PrintVersionsCommand(logger: _logger));
 
     argParser
       ..addFlag('verbose', abbr: 'v')
       ..addOption(
         'root',
-        defaultsTo: '.',
         help: 'Directory in which to find checkouts.',
       )
       ..addOption(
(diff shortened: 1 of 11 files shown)


packages/cutler/bin/cutler.dart
```diff
@@ -27,24 +30,60 @@ class Cutler extends CommandRunner<int> {
       ..addFlag('update', defaultsTo: true, help: 'Update checkouts.');
   }
 
+  final Logger _logger;
+
+  Iterable<String> missingDirectories(String rootDir) {
```
bryanoltman, line 35: A docstring would be helpful here. "Returns paths to all missing repositories (see [Repo.values])" or something

packages/cutler/bin/cutler.dart
```diff
@@ -27,24 +30,60 @@ class Cutler extends CommandRunner<int> {
...
+    return Repo.values.map((repo) => '$rootDir/${repo.path}').where(
+          (path) => !Directory(path).existsSync(),
+        );
+  }
+
+  String fallbackRootDir() {
```
bryanoltman, line 41: Same here re: doc string. What is being fallen back to/from? Why would fallback happen?

packages/cutler/lib/commands/print_versions_command.dart
```diff
@@ -14,12 +14,30 @@ class PrintVersionsCommand extends Command<int> {
 
   @override
   int run() {
-    final shorebirdHash = argResults!.rest.first;
+    late final String shorebirdHash;
+    if (argResults!.rest.isEmpty) {
```
bryanoltman, line 20: A named arg with a default value might be cleaner here

packages/cutler/lib/commands/print_versions_command.dart
```diff
@@ -14,12 +14,30 @@ class PrintVersionsCommand extends Command<int> {
...
+
+    final flutterForkpoint = Repo.flutter.getForkPoint(shorebird.flutter.hash);
+    // This is slightly error-prone in that we're assuming that our engine and
+    // buildroot forks started from the correct commit.  But I'm not sure how
+    // to determine the forkpoint otherwise.  engine and buildroot don't have
+    // a stable branch, yet they do seem to "branch" for stable releases at the
+    // x.x.0 release.
+    final forkpoints = getFlutterVersions(flutterForkpoint.hash);
+    print('Forkpoints:');
+    printVersions(forkpoints, 2);
```
bryanoltman: Potentially out-of-scope for this PR, but named arguments would make `printVersions` a bit more readable. I'm not totally sure what `forkpoints` is (I presume a list of the form `['1.0.0', '1.1.0', ...]` but am not sure) and don't know what the significance of `2` is.

felangel (approved):

packages/cutler/lib/commands/print_versions_command.dart
```diff
@@ -14,12 +14,30 @@ class PrintVersionsCommand extends Command<int> {
...
+      print('No Shorebird hash provided, using `origin/stable`.');
+      shorebirdHash = 'origin/stable';
+    } else {
+      shorebirdHash = argResults!.rest.first;
+    }
+
     final shorebirdFlutter = Repo.shorebird
         .contentsAtPath(shorebirdHash, 'bin/internal/flutter.version');
     final shorebird = getFlutterVersions(shorebirdFlutter);
     print('Shorebird $shorebirdHash:');
```
bryanoltman:
```suggestion
    logger.info('Shorebird $shorebirdHash:');
```

packages/cutler/lib/commands/print_versions_command.dart
eseidel: I hear you.  I was going for more of a `git foo [rev]` feel.

packages/cutler/lib/commands/print_versions_command.dart
eseidel: Done.

packages/cutler/lib/commands/print_versions_command.dart
eseidel: forkpoints is a VersionSet.  I'll make indent be a named arg for now.

packages/cutler/bin/cutler.dart
eseidel: Added.

packages/cutler/bin/cutler.dart
eseidel: Added.

Issue #580, opened 2023-06-01
feat: basic `shorebird release ios` and `shorebird patch ios`


eseidel:
In order to have a working ios demo, I need the server/cli parts of `shorebird` to know what iOS is.

You can see my hacks at this:
https://github.com/shorebirdtech/shorebird/tree/ios_hacks

Specifically i would like to be able to run a `shorebird release ios` and have it create the necessary server entries.  It could upload the ipa parts, but that's not required.

I need `shorebird patch ios` to be able to create an artifact on the server that is *not* a patch (e.g. skip the patch creation logic on the client, the server can think it's a patch, that's fine, I can teach the rust to ignore that for iOS for now).

To create the file for the artifact we will use:
`/Users/eseidel/Documents/GitHub/ios_engine/src/out/ios_release_unopt_arm64/clang_x64/gen_snapshot_arm64 --deterministic --snapshot-kind=app-aot-elf --elf=out.aot .dart_tool/flutter_build/4d4af174f68149fadd13bb2aa91c5a1c/app.dill`
(which I can just hard code into my local `shorebird` for the demo).

The app.dill path is the most recently modified app.dill in the `.dart_tool` directory (I'm sure there is a better way to find it, but that will work for now).  The `gen_snapshot` path would get resolved from the --local-engine path, but again, it's not worth wiring up for real, just getting something basic checked in and then I can hack it locally to work for my demo.

Other issues I hit when trying this locally, were that the code doesn't really know that iOS exists and has global assumptions like "architectures" assuming android, and validators assuming android.  Some of that could certainly get cleaned up by someone hacking on that part. see https://github.com/shorebirdtech/shorebird/tree/ios_hacks for where I hacked things.

Another piece is probably reading out the version information from the Info.plist, which should be straighforward with the `defaults` command line tool build into mac which I believe is a generic plist reader/editor despite the name.

Once these basic commands are in place, I can finish the rust side of the work which should mean we have a working demo!

eseidel:
The reason why we're calling gen_snapshot ourselves is to produce the elf binary which we're using as a bundle format for the compiled dart code.  On iOS gen_snapshot is called with --snapshot-kind=aot-assembly to create an assembly file instead to hand off to the xcode build system for creation of an App.framework.  We don't need to create a framework (or even involve XCode) to compile the patch from dart code.  We might do that later (diff between App.frameworks), but since Dart already has a parser for ELF, we're just using that as a bundling format for this demo.

[...]

PR #594, opened 2023-06-05
feat(shorebird_cli): create a release with `shorebird release ios-preview`

bryanoltman:
## Description

This change updates `shorebird release ios` to create a new iOS release. For now, this doesn't involve building or uploading anything, and simply:

1. Derives the version of the release from the app's `Info.plist` and `*.xcconfig` files.
2. Creates a release entry on the backend.

Part of https://github.com/shorebirdtech/shorebird/issues/580

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
```diff
@@ -144,7 +144,13 @@ of the Android app that is using this module.''',
     );
 
     final appId = shorebirdYaml.getAppId(flavor: flavor);
-    final app = await getApp(appId: appId, flavor: flavor);
+    final App? app;
+    try {
+      app = await getApp(appId: appId, flavor: flavor);
+    } catch (_) {
```
eseidel, line 150: Are we supposed to silently exit here?  I guess getApp does the logging?

eseidel: If so, maybe these should also do the exit?  I presume we repeat these lines elsewehre. 🤷

packages/shorebird_cli/lib/src/shorebird_ios_release_version_mixin.dart
```diff
@@ -0,0 +1,97 @@
...
+    );
+    var releaseVersion = plist[releaseVersionKey] as String?;
+    var buildNumber = plist[buildNumberKey] as String?;
+
+    if (releaseVersion == null) {
+      throw Exception('Could not determine release version');
+    }
+
+    if (configPropertyRegex.hasMatch(releaseVersion)) {
+      releaseVersion = pListVariables[
```
eseidel: Kinda wild that we have to do our own plist variable substitution.  Feels like it belongs in a function somewhere. 🤷

eseidel (commented):
lgtm.  Felix knows this code better than I do though.  Thanks!

eseidel (approved):

packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
bryanoltman: > I guess getApp does the logging?

Correct. `getApp` is defined in the `ShorebirdCodePushClientMixin`, which I created to wrap a lot of repeated logging and code push client interaction.

> maybe these should also do the exit?

I'm not sure there's a way to do that while still keeping the code testable – `exit(code)` terminates the dart test process. I'm open to ideas though – I don't love this design, it just seemed better than copy/pasting a bunch of code from existing commands.

packages/shorebird_cli/lib/src/shorebird_ios_release_version_mixin.dart
bryanoltman: I think we'll eventually be able to rely on the xcode build to do this for us, but I think this should be robust enough for now.

packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
eseidel: I suspect flutter_tools (or other large dart CLI apps) must solve this?  https://github.com/flutter/flutter/blob/a89d475ca2c44b1cfdec4563c26cea6c261fc350/packages/flutter_tools/lib/src/base/io.dart#L115

bryanoltman: I like this a lot! Thanks for the link, I'll look into adopting this pattern tomorrow.

packages/shorebird_cli/lib/src/shorebird_ios_release_version_mixin.dart
bryanoltman: Alternatively, given that we will be using the ipa to get the version number when running `shorebird patch ios`, maybe it would make sense to just build the ipa for the version number? It seems wasteful, but it's more likely to be consistent and it means that we have a more consistent (and probably more correct) way to determine version number.

felangel: Yeah I think that makes sense. We should be consistent with how we determine the version so I agree building the ipa seems like the best option (unless we can also get it from the xcarchive)

packages/shorebird_cli/lib/src/shorebird_ios_release_version_mixin.dart
```diff
@@ -0,0 +1,97 @@
...
+  static const releaseVersionKey = 'CFBundleShortVersionString';
+
+  /// The version of the build that identifies an iteration of the bundle.
+  ///
+  /// See https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion
+  static const buildNumberKey = 'CFBundleVersion';
+
+  /// Checks the iOS Info.plist and referenced .xcconfig files to determine the
+  /// app's release version and build number.
+  Future<String> determineIosReleaseVersion() async {
```
felangel: As discussed can we build the ipa/xcarchive and derive the release version from the artifacts similar to how android works?

bryanoltman: Done.

packages/shorebird_cli/lib/src/shorebird_ios_release_version_mixin.dart
bryanoltman: Done.

packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
bryanoltman: Working on this in a separate branch, will merge as part of a larger refactor.

eseidel (approved):

PR #595, opened 2023-06-06
feat(shorebird_cli): add `patch ios-preview` command

bryanoltman:
## Description

Adds `shorebird patch ios` command, which generates an aot snapshot and uploads it as a patch artifact.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/archive_analysis/ipa.dart
```diff
@@ -0,0 +1,64 @@
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:archive/archive_io.dart';
+import 'package:propertylistserialization/propertylistserialization.dart';
+
+class IpaReader {
```
eseidel: I suspect  the linter will yell at you for lack of docs. 🤣

packages/shorebird_cli/lib/src/archive_analysis/ipa.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  /// The version number of the IPA, as derived from the app's Info.plist.
+  String get versionNumber {
+    final plist = _getPlist();
+    final releaseVersion = plist[releaseVersionKey] as String?;
+    final buildNumber = plist[buildNumberKey] as String?;
+    if (releaseVersion == null) {
+      throw Exception('Could not determine release version');
+    }
+    if (buildNumber != null) {
```
eseidel: I'd probably write this as a ternary because it fits on one line. 🤷

packages/shorebird_cli/lib/src/archive_analysis/ipa.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  Map<String, Object> _getPlist() {
+    final plistPathRegex = RegExp(r'Payload/[\w]+.app/Info.plist');
+    final content = ZipDecoder()
+        .decodeBuffer(InputFileStream(_ipaFile.path))
+        .files
+        .where((file) {
+          return file.isFile && plistPathRegex.hasMatch(file.name);
+        })
+        .first
```
eseidel: Will this ever fail?  Is an exception on failure ok?

packages/shorebird_cli/lib/src/archive_analysis/ipa.dart
```diff
@@ -0,0 +1,64 @@
...
+        .content as Uint8List;
+
+    return PropertyListSerialization.propertyListWithData(
+      ByteData.view(content.buffer),
+    ) as Map<String, Object>;
```
eseidel: I'm confused.  This returns Object, but then uses it as String?  Is Object nullable?

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+        ShorebirdCodePushClientMixin {
+  /// {@macro patch_ios_command}
+  PatchIosCommand({
+    required super.logger,
+    super.auth,
+    super.buildCodePushClient,
+    super.validators,
+    HashFunction? hashFn,
+    IpaReader? ipaReader,
+  })  : _hashFn = hashFn ?? ((m) => sha256.convert(m).toString()),
```
eseidel, line 37: Feels like this should be defaultHashFn or something somewhere to avoid us ever forgetting to change one.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+        _ipaReader = ipaReader ?? IpaReader() {
+    argParser
+      ..addOption(
+        'release-version',
+        help: 'The version of the release (e.g. "1.0.0").',
+      )
+      ..addOption(
+        'channel',
+        help: 'The channel the patch should be promoted to (e.g. "stable").',
```
eseidel: I assume channels are dead code still?

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+        allowed: ['stable'],
+        allowedHelp: {
+          'stable': 'The stable channel which is consumed by production apps.'
+        },
+        defaultsTo: 'stable',
+      )
+      ..addOption(
+        'target',
+        abbr: 't',
+        help: 'The main entrypoint file of the application.',
```
eseidel, line 43: --target feels like it will come back to bite us when we try to do these via push to deploy. 🤷

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+        checkValidators: true,
+      );
+    } on PreconditionFailedException catch (error) {
+      return error.exitCode.code;
+    }
+
+    const platform = 'ios';
+    final force = results['force'] == true;
+    final dryRun = results['dry-run'] == true;
+    final flavor = results['flavor'] as String?;
```
eseidel, line 89: One day dart will have a type-safe args parser.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+    }
+
+    if (app == null) {
+      logger.err(
+        '''
+Could not find app with id: "$appId".
+Did you forget to run "shorebird init"?''',
+      );
+      return ExitCode.software.code;
+    }
```
eseidel, line 113: This feels like code we can share in a later pass.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+        platform: 'ios',
+        hash: _hashFn(await aotFile.readAsBytes()),
+      );
+    } catch (error) {
+      createArtifactProgress.fail('$error');
+      return ExitCode.software.code;
+    }
+    createArtifactProgress.complete();
+
+    Channel? channel;
```
eseidel, line 281: Does the channel code do anything?

packages/shorebird_cli/lib/src/shorebird_artifact_mixin.dart
```diff
@@ -56,4 +56,23 @@ mixin ShorebirdArtifactMixin on ShorebirdCommand {
     await unzipFn(zipPath, extractedAarDir);
     return extractedAarDir;
   }
+
+  File newestAppDill() {
```
eseidel, line 63: I recommend adding a note explaining how much of a hack this is.  We don't actually know if this is even finding the correct dill for example. :)

packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
```diff
@@ -182,4 +184,54 @@ mixin ShorebirdBuildMixin on ShorebirdCommand {
...
+    );
+
+    if (result.exitCode != 0) {
+      throw Exception('Failed to create diff: ${result.stderr}');
+    }
+
+    return diffPath;
+  }
+
+  Future<File> buildAotSnapshot({required String appDillPath}) async {
```
eseidel: Specifically this is building an elf aot snapthot (there are other kinds).

packages/shorebird_cli/lib/src/shorebird_environment.dart
```diff
@@ -63,4 +63,16 @@ abstract class ShorebirdEnvironment {
           'flutter',
         ),
       );
+
+  static File get genSnapshotFile => File(
```
eseidel, line 67: This will end up with the wrong one when we specify --local-engine-src-path (which is the only way which can work today).  flutter_tools has a way by which they get all their paths like this from some sort of shared config object which knows how to respect --local-engine.  We don't necessarily need something fancy, but this will produce the wrong result for now.   I guess that's also fine as I can hack it locally for the demo, but I would at least leave a fixme to say this doesn't support --local-engine.

eseidel:
Very exciting, thank you!

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
felangel: Yeah we should probably avoid exposing this since it's not something users can change and they don't need to know about this for now.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
felangel: yeah it's needed to promote the patch to the stable channel since the act of creating a patch doesn't immediately make it available atm. We can change that and remove the concept of a channel entirely but it would involve backend changes as well.

packages/shorebird_cli/lib/src/shorebird_artifact_mixin.dart
felangel: Would another option be to wipe out `.dart_tool/flutter_build` before generating the `app.dill` to ensure there is only a single one?

packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
felangel:
```suggestion
  Future<File> buildElfAotSnapshot({required String appDillPath}) async {
```

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -0,0 +1,332 @@
...
+        final ipa = _ipaReader.read(
+          p.join(
+            Directory.current.path,
+            'build',
+            'ios',
+            'ipa',
+            '${pubspec.name}.ipa',
+          ),
+        );
+        releaseVersion = releaseVersionArg ?? ipa.versionNumber;
```
felangel: Shouldn't we disallow manually overriding the releaseVersion?

packages/shorebird_cli/lib/src/shorebird_artifact_mixin.dart
eseidel: We could also parse it out of the logs that flutter build makes. 🤷  I think the current "newest" hack is fine for now (we might end up just using the output from the ipa eventually, that just requires writing a mach-o parser), just wanted it marked as a hack.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
bryanoltman: WIP on https://github.com/shorebirdtech/shorebird/tree/bo/code-push-client-wrapper

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
bryanoltman: Ah yeah, good call, will update

packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
bryanoltman: Done

bryanoltman: Clarified

packages/shorebird_cli/lib/src/shorebird_artifact_mixin.dart
bryanoltman: Added a comment :)

packages/shorebird_cli/lib/src/shorebird_environment.dart
bryanoltman: Added a TODO, will fix in an upcoming PR

packages/shorebird_cli/lib/src/archive_analysis/ipa.dart
bryanoltman: I think this is a safe assumption for iOS app plists – I can't imagine a case where an Info.plist wouldn't be a dictionary.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
bryanoltman: Removed.

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
bryanoltman: Removed the argument. We now always use "stable".

packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
bryanoltman: We don't actually override it anywhere. @felangel – do you remember why this was parameterized?

felangel: I thought this was overridden for testing?

bryanoltman: I presume that was the original idea, but I'm not seeing us actually doing that anywhere.

bryanoltman: ¯\_(ツ)_/¯ we can clean it up as a chore in the future

eseidel (approved):
My sense is this is good enough, we should check it in and iterate. :)

PR #597, opened 2023-06-06
feat: teach `cutler version` how to print flutter versions

eseidel:
I wanted a way to print flutter version dependencies without needing
a shorebird version (e.g. for seeing what they will be when we
update shorebird).

While I was editing Cutler, I also made it not crash when run
via `dart run` with no path.

I also renamed `print-versions` to just `versions`.

I also made `versions` know how to update repositories.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/cutler/lib/commands/versions_command.dart
```diff
@@ -0,0 +1,72 @@
...
+      defaultsTo: 'shorebird',
+    );
+  }
+  @override
+  final name = 'versions';
+  @override
+  final description =
+      'Print the versions a Shorebird or Flutter release hash depends on.';
+
+  @override
```
bryanoltman, line 26:
```suggestion
    );
  }
  
  @override
  final name = 'versions';
  
  @override
  final description =
      'Print the versions a Shorebird or Flutter release hash depends on.';

  @override
```

bryanoltman (approved):

PR #615, opened 2023-06-09
chore: split versions_test from cutler_test, mostly to trigger coverage

eseidel:
split versions_test from cutler_test, mostly to trigger coverage


eseidel:
@felangel maybe `secrets` is not valid in actions, only workflows?

felangel:
> @felangel maybe `secrets` is not valid in actions, only workflows?

Ah yeah good call we’ll need to add it as an input to the action. I’ll make that change shortly 👍

felangel:
Should be fixed by #619

bryanoltman:
`14.035087719298245 is greater than or equal to min_coverage 10.` 🙌

bryanoltman (approved):

PR #618, opened 2023-06-09
fix(shorebird_cli): handle case where IPA generation failed with exit code 0

### LINKED ISSUE #606: fix(shorebird_cli): ipa builds can fail silently
**Description**

When running `shorebird release ios-preview` or other commands that build an ipa, if the build fails (due to incorrect developer team/cert setup in xcode, for example), the command fails silently. The shorebird command ultimately fails because it cannot find an ipa file, but it's not obvious why that is.

Running the underlying flutter command directly will show more helpful output:

```
Building App Store IPA...                                           5.9s
Encountered error while creating the IPA:
error: exportArchive: Communication with Apple failed
error: exportArchive: No signing certificate "iOS Distribution" found
```

We should communicate this failure so it's obvious what went wrong, possibly just by displaying this output when running the shorebird command.

## Reproduction steps

1. Create a new project with `flutter create`
2. Initialize shorebird with `shorebird init`
3. Open the .xcworkspace in Xcode, assign a development team
4. Run `shorebird release ios-preview`

bryanoltman:
## Description

Handles the case where the command to build an IPA "succeeds" but does not actually generate an IPA.

Fixes https://github.com/shorebirdtech/shorebird/issues/606

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/.github/actions/dart_package/action.yaml
+++ b/.github/actions/dart_package/action.yaml
@@ -34,6 +34,9 @@ inputs:
     required: false
     default: "vm"
     description: Platform to use when running tests
+  secrets:
+    required: true
+    description: GitHub secretes
 
 runs:
   using: "composite"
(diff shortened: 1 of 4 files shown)


eseidel (approved):

.github/actions/dart_package/action.yaml
```diff
@@ -34,6 +34,9 @@ inputs:
     required: false
     default: "vm"
     description: Platform to use when running tests
+  secrets:
+    required: true
+    description: GitHub secretes
```
eseidel: secretes! 🎩

bryanoltman: lol

PR #631, opened 2023-06-12
fix(shorebird_cli): put ios patch artifact in build directory 

### LINKED ISSUE #622: fix: commands that build ipas leave an `aot.out` file in the project root directory
**Description**

After running `shorebird release ios-preview` or `shorebird patch ios-preview`, there is an `aot.out` file in the project's root directory. We should either clean this up or place it in a hidden directory.

bryanoltman:
## Description

Deletes the patch artifact generated by `shorebird patch ios` after it is uploaded.

Fixes https://github.com/shorebirdtech/shorebird/issues/622

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_ios_command.dart
```diff
@@ -246,6 +246,14 @@ ${summary.join('\n')}
       },
     );
 
+    final cleanupProgress = logger.progress('Cleaning up');
+    try {
+      await aotFile.delete();
```
felangel: What are your thoughts on writing it to a temp directory instead? I feel it would be better than to be modifying the user's workspace especially because partial successes would still lead to having the aotFile around.

bryanoltman: This is probably a better idea. Will update.

bryanoltman: Done.

felangel (approved):

PR #635, opened 2023-06-12
feat(shorebird_cli): automatically upgrade when newer version is available

felangel:
## Description

- feat(shorebird_cli): automatically upgrade when newer version is available

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/command_runner.dart
```diff
@@ -126,6 +127,31 @@ class ShorebirdCliCommandRunner extends CompletionCommandRunner<int> {
...
+    if (!isUpToDate) {
+      final progress = logger.progress(
+        'A new version of Shorebird is available! Upgrading',
+      );
+      try {
+        logger.detail('[upgrader] Upgrading...');
+        await upgrader.upgrade();
+        logger.detail('[upgrader] Upgrading complete.');
+        progress.complete();
+        return ExitCode.success.code;
```
bryanoltman: If feels a bit odd to exit after a successful upgrade. Is there no way to run the command post-update?

packages/shorebird_cli/lib/src/command_runner.dart
```diff
@@ -126,6 +127,31 @@ class ShorebirdCliCommandRunner extends CompletionCommandRunner<int> {
...
+      } catch (error) {
+        logger.detail('[upgrader] Failed to upgrade: $error');
+        progress.fail(error.toString());
```
bryanoltman: Probably want to switch these lines around so the `detail` doesn't interfere with the `progress`

packages/shorebird_cli/lib/src/upgrader.dart
```diff
@@ -0,0 +1,115 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:scoped/scoped.dart';
+import 'package:shorebird_cli/src/shorebird_process.dart';
+
+// A reference to a [Logger] instance.
```
bryanoltman:
```suggestion
// A reference to a [Upgrader] instance.
```

packages/shorebird_cli/lib/src/upgrader.dart
```diff
@@ -0,0 +1,115 @@
...
+final upgraderRef = create(Upgrader.new);
+
+// The [Upgrader] instance available in the current zone.
+Upgrader get upgrader => read(upgraderRef);
+
+class Upgrader {
```
bryanoltman, line 16: Needs docs here and throughout

packages/shorebird_cli/lib/src/upgrader.dart
```diff
@@ -0,0 +1,115 @@
...
+      workingDirectory: workingDirectory,
+    );
+    // Get the latest commit revision of the upstream
+    return _gitRevParse('@{upstream}', workingDirectory: workingDirectory);
+  }
+
+  /// Returns the local HEAD shorebird hash.
+  ///
+  /// Exits if HEAD isn't pointing to a branch, or there is no upstream.
+  Future<String> fetchCurrentGitHash({
```
bryanoltman: I think this can be private

packages/shorebird_cli/lib/src/command_runner.dart
felangel: I don't think it will include the updates if we continue execution.

packages/shorebird_cli/lib/src/command_runner.dart
felangel: ah yeah good catch

packages/shorebird_cli/lib/src/upgrader.dart
felangel: yup good call

packages/shorebird_cli/lib/src/command_runner.dart
bryanoltman: Maybe we could kick off a new process with same arguments?

felangel: yeah sg will make that change in a few mins

bryanoltman: If there are some drawbacks to that approach (i.e., it's not trivial to forward stdout etc.), we should at least tell them that they command they ran didn't actually execute and they should try again.

felangel: done ✅

eseidel:
How do server rollouts work in this world?

If we ship a new server while someone is in the middle of executing the shorebird command, what will happen?

Is there a race between when you check if the server you talked to is the right version and the next command you send to it?

eseidel:
This seems very similar to the problem that "please refresh the page, the app has changed" problem web apps have.

PR #638, opened 2023-06-12
Add a channel command

sladyn98:
## Description

This PR would add support for a channel command that uses git branches as channels.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/channel_command.dart
```diff
@@ -0,0 +1,26 @@
...
+  final String name = 'channel';
+
+  @override
+  final String description = 'Switch to a different Shorebird channel.';
+
+  @override
+  void run() async {
+    final branch = argResults.rest.isNotEmpty ? argResults.rest[0] : null;
+    if (branch == null) {
+      print('Please provide a branch name to switch to.');
```
felangel, line 15: We should use the `logger` instead of `print`

packages/shorebird_cli/lib/src/commands/channel_command.dart
```diff
@@ -0,0 +1,26 @@
...
+      return;
+    }
+
+    final result = await Process.run('git', ['checkout', branch]);
```
felangel, line 19: We should use `ShorebirdProcess` instead of `Process`

packages/shorebird_cli/lib/src/commands/channel_command.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:io';
+import 'package:args/command_runner.dart';
+
+class ChannelCommand extends Command {
```
felangel, line 4: We'll need to add tests for the command in order for this to be mergable.

packages/shorebird_cli/lib/src/commands/channel_command.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:io';
+import 'package:args/command_runner.dart';
```
felangel, line 2:
```suggestion
import 'package:args/args.dart';
```

packages/shorebird_cli/lib/src/commands/commands.dart
```diff
@@ -2,6 +2,7 @@ export 'account/account.dart';
 export 'apps/apps.dart';
 export 'build/build_command.dart';
 export 'cache/cache.dart';
+export 'channel_command.dart';
```
felangel, line 5: We'll also need to add the new command to the `CommandRunner` in order for it to be usable.

felangel (changes requested):
Thanks for the PR! I left a few comments but I also am not sure this is needed at the moment since we release quite frequently. We'll discuss a bit more an let you know but either way we really appreciate you taking the time to contribute! 💙

bryanoltman:
This seems abandoned and doesn't have any tests. Going to close, please reopen if you're still interested in landing.

PR #641, opened 2023-06-13
fix(shorebird_cli): tell xcodebuild not to manage ipa build number

### LINKED ISSUE #625: fix: patching ios after uploading to app store builds wrong version of IPA
**Description**

When patching an existing iOS release, `shorebird patch ios` will build an IPA with the incorrect build number (one higher than what was uploaded to App Store Connect).

**Steps To Reproduce**

1. Build release with `shorebird release ios`
2. Upload .xcarchive to app store
3. Run `shorebird patch ios`

**Expected Behavior**

The patch applies to the release version specified in `pubspec.yaml`

**Actual Behavior**

`shorebird patch ios` builds an IPA with version number+1 (if `shorebird release ios` generated a ipa for 1.0.0+1, `shorebird patch ios` will generate an ipa for 1.0.0+2)

bryanoltman:
## Description

By default, xcodebuild will auto-increment an ipa's build number if there is an existing build on app store connect (see https://developer.apple.com/forums/thread/690647?answerId=689925022#689925022). To fix this, we provide an ExportOptions.plist telling it specifically not to do so.

Fixes https://github.com/shorebirdtech/shorebird/issues/625

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/test/src/commands/patch/patch_ios_command_test.dart
+++ b/packages/shorebird_cli/test/src/commands/patch/patch_ios_command_test.dart
@@ -5,6 +5,7 @@ import 'package:http/http.dart' as http;
 import 'package:mason_logger/mason_logger.dart';
 import 'package:mocktail/mocktail.dart';
 import 'package:path/path.dart' as p;
+import 'package:propertylistserialization/propertylistserialization.dart';
 import 'package:scoped/scoped.dart';
 import 'package:shorebird_cli/src/archive_analysis/archive_analysis.dart';
 import 'package:shorebird_cli/src/auth/auth.dart';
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/test/src/commands/patch/patch_ios_command_test.dart
```diff
@@ -474,6 +475,42 @@ base_url: $baseUrl''',
       );
     });
 
+    test('provides appropriate ExportOptions.plist to build ipa command',
```
felangel, line 478: ideally we shouldn't need to duplicate the test :(

eseidel:
So this makes the behavior of `shorebird build` different from `flutter build ipa` I guess?  Do we think that will matter to users?

Can we re-state the problem we're solving (mostly for posterity)?  I presume it's that we're trying to get a reliable build number out of the ios project somehow so we can tell shorebird "here is the build number of this release"?

packages/shorebird_cli/test/src/commands/patch/patch_ios_command_test.dart
bryanoltman: Yep, we'll get that once we write a ShorebirdBuildWrapper or similar and move this code out of a mixin :)

bryanoltman:
> So this makes the behavior of shorebird build different from flutter build ipa I guess?

Less than you might think. Flutter does something similar (https://github.com/flutter/flutter/blob/4e706975b7bfaa3887c39ed277d3b83cc23427b2/packages/flutter_tools/lib/src/commands/build_ios.dart#L535-L548), and our ExportInfo.plist is a superset of theirs.

> Can we re-state the problem we're solving (mostly for posterity)? I presume it's that we're trying to get a reliable build number out of the ios project somehow so we can tell shorebird "here is the build number of this release"

Added more detail in the shorebird_build_mixin.dart doc comment

felangel (approved):

PR #667, opened 2023-06-15
feat(code_push_client): Allow release commands to be run if no release artifacts exist for the given platform

### LINKED ISSUE #610: fix: can't release to both iOS and Android
When we write up a tutorial explaining how to release to both this will become obvious.  But when I tried just now to release to android (after having released to iOS) I got:

```
eseidel@erics-mbp ios_test % shorebird --local-engine-src-path=$HOME/Documents/GitHub/engine/src release android
✓ Building release (33.6s)
✓ Fetching apps (0.6s)
✓ Detecting release version (0.2s)

🚀 Ready to create a new release!

📱 App: ios_test (55517612-9946-4c25-97b2-e0df1a2ba951)
📦 Release Version: 1.0.0+1
🕹️  Platform: android (arm64, arm32, x86_64)

Would you like to continue? (y/N) Yes
✓ Fetching releases (88ms)
It looks like you have an existing release for version 1.0.0+1.
Please bump your version number and try again.
```

### LINKED ISSUE #658: feat: check for existing platform artifacts if release version already exists

felangel:
## Description

If a `shorebird release` command detects an existing release, verify that there are existing release artifacts before telling the user that they cannot release. If there are none, assume that the release was created for a different platform and let them continue.

Fixes https://github.com/shorebirdtech/shorebird/issues/658
Fixes https://github.com/shorebirdtech/shorebird/issues/610

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -124,6 +124,35 @@ Did you forget to run "shorebird init"?''',
     }
   }
 
+  /// Exits if [platform] release artifacts already exist for an
+  /// [existingRelease].
+  Future<void> verifyCanRelease({
+    required Release existingRelease,
+    required String platform,
+  }) async {
+    logger.detail('Verifying ability to release');
+
+    final artifacts = await codePushClient.getReleaseArtifacts(
+      releaseId: existingRelease.id,
+      platform: platform,
+    );
+
+    logger.detail(
+      '''
+Artifacts for release:${existingRelease.version} platform:$platform
+  $artifacts''',
+    );
+
+    if (artifacts.isNotEmpty) {
+      logger.err(
+        '''
+It looks like you have an existing $platform release for version ${lightCyan.wrap(existingRelease.version)}.
+Please bump your version number and try again.''',
+      );
+      exit(ExitCode.software.code);
+    }
+  }
+
   Future<Release> getRelease({
     required String appId,
     required String releaseVersion,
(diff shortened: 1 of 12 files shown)


packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -220,13 +257,19 @@ Please create a release using "shorebird release" and try again.
       'Fetching $arch artifact',
     );
     try {
-      final artifact = await codePushClient.getReleaseArtifact(
+      final artifacts = await codePushClient.getReleaseArtifacts(
```
felangel, line 260: can `getReleaseArtifact` call `maybeGetReleaseArtifact` to avoid code duplication?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -189,18 +218,26 @@ Please create a release using "shorebird release" and try again.
...
+    // TODO(bryanoltman): update this function to only make one call to
+    // getReleaseArtifacts.
     final releaseArtifacts = <Arch, ReleaseArtifact>{};
     final fetchReleaseArtifactProgress = logger.progress(
       'Fetching release artifacts',
     );
     for (final entry in architectures.entries) {
       try {
-        final releaseArtifact = await codePushClient.getReleaseArtifact(
+        final artifacts = await codePushClient.getReleaseArtifacts(
```
felangel, line 229: can we call `getReleaseArtifact` here to avoid code duplication until the TODO is resolved?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: That would simplify this a lot, but doing so would show a new `progress` for every artifact, which I don't think is what we want.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: Same here re: the progress per artifact

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -124,6 +124,35 @@ Did you forget to run "shorebird init"?''',
     }
   }
 
+  /// Exits if [platform] release artifacts already exist for an
+  /// [existingRelease].
+  Future<void> verifyCanRelease({
```
```suggestion
  Future<void> ensureReleaseHasNoArtifacts({
```

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
felangel: That shouldn't be a problem if we also refactor this to be `getReleaseArtifacts` instead and make a single call instead of 3 right?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: Done.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: Correct, this will go away when we refactor the commands to just use `getReleaseArtifacts`. That's part of a larger refactoring to include aabs and aars in the list of artifacts. As it stands, we call `getReleaseArtifacts` for the libapp.so files and `getReleaseArtifact` for the aab/aar files, so we can't hide `getReleaseArtifact`.

bryanoltman: I see what you're saying – I'm just punting that refactor to a future PR to keep this one reasonably sized

bryanoltman (approved):

PR #671, opened 2023-06-16
fix(shorebird_cli): only run project-specific validators when in a shorebird project directory

### LINKED ISSUE #669: fix: `shorebird doctor` should warn if it is not being run at the root of a project directory

bryanoltman:
## Description

Adds `ValidationScope` enum to validators to allow them to specify whether they perform project-specific validation or system-wide validation and updates `shorebird doctor` to only run project-specific validators when in a project directory.

Fixes https://github.com/shorebirdtech/shorebird/issues/669

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/shorebird_version_validator.dart
+++ b/packages/shorebird_cli/lib/src/validators/shorebird_version_validator.dart
@@ -11,10 +11,11 @@ class ShorebirdVersionValidator extends Validator {
   final Future<bool> Function({required String workingDirectory})
       isShorebirdVersionCurrent;
 
-  // coverage:ignore-start
   @override
   String get description => 'Shorebird is up-to-date';
-  // coverage:ignore-end
+
+  @override
+  ValidatorScope get scope => ValidatorScope.installation;
 
   @override
   Future<List<ValidationIssue>> validate(ShorebirdProcess process) async {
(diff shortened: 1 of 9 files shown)


packages/shorebird_cli/lib/src/validators/shorebird_version_validator.dart
```diff
@@ -11,10 +11,11 @@ class ShorebirdVersionValidator extends Validator {
   final Future<bool> Function({required String workingDirectory})
       isShorebirdVersionCurrent;
 
-  // coverage:ignore-start
   @override
   String get description => 'Shorebird is up-to-date';
-  // coverage:ignore-end
+
+  @override
+  ValidatorScope get scope => ValidatorScope.installation;
```
felangel, line 18: can a validator have multiple scopes?

felangel (approved):
LGTM just wondering whether a validator could be associated with more than one scope. That can be a future thing if we have a use-case though.

packages/shorebird_cli/lib/src/validators/shorebird_version_validator.dart
bryanoltman: I don't think so? If a validator is checking both project setup and global installation setup, it should probably be split into two validators.

PR #680, opened 2023-06-16
feat(shorebird_cli): add `shorebird account usage` command

felangel:
## Description

- feat(shorebird_cli): add `shorebird account usage` command

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -527,4 +527,18 @@ aar artifact already exists, continuing...''',
 
     await promotePatch(patchId: patch.id, channel: channel);
   }
+
+  Future<List<AppUsage>> getUsage() async {
+    final progress = logger.progress('Fetching usage');
+    final List<AppUsage>? appsUsage;
+    try {
+      appsUsage = await codePushClient.getUsage();
+      progress.complete();
+    } catch (error) {
+      progress.fail(error.toString());
+      exit(ExitCode.software.code);
+    }
+
+    return appsUsage;
+  }
 }
--- a/packages/shorebird_cli/lib/src/commands/account/account_usage_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/account/account_usage_command.dart
@@ -1,0 +1,85 @@
+import 'dart:async';
+
+import 'package:barbecue/barbecue.dart';
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
+import 'package:shorebird_cli/src/command.dart';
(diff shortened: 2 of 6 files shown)


packages/shorebird_cli/lib/src/commands/account/account_usage_command.dart
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    final progress = logger.progress('Fetching usage');
+    final List<AppUsage>? appsUsage;
+    try {
+      appsUsage = await codePushClientWrapper.codePushClient.getUsage();
+      progress.complete();
+    } catch (error) {
+      progress.fail(error.toString());
+      return ExitCode.software.code;
```
bryanoltman: Accessing the codePushClient through the wrapper seems a bit wrong. I'd add a `getUsage` method to CodePushClientWrapper and move this logging and error handling there.

felangel: yeah good call, done.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -527,4 +527,18 @@ aar artifact already exists, continuing...''',
...
+    final List<AppUsage>? appsUsage;
+    try {
+      appsUsage = await codePushClient.getUsage();
+      progress.complete();
+    } catch (error) {
+      progress.fail(error.toString());
+      exit(ExitCode.software.code);
+    }
+
+    return appsUsage;
```
bryanoltman, line 542: nit: this could just return from within the try and you wouldn't need to declare it separately on line 533

packages/shorebird_cli/lib/src/commands/account/account_usage_command.dart
```diff
@@ -0,0 +1,77 @@
...
+  @override
+  String get description => 'Get usage information for your Shorebird account.';
+
+  @override
+  String get name => 'usage';
+
+  @override
+  Future<int> run() async {
+    try {
+      await validatePreconditions(checkUserIsAuthenticated: true);
```
bryanoltman, line 27: We probably want `checkShorebirdInitialized` here too

bryanoltman (approved):

packages/shorebird_cli/lib/src/commands/account/account_usage_command.dart
felangel: why? this is an account level command so I don't think it needs to be run within a shorebird app

bryanoltman: Ah you're right, I was thinking this would be getting app id from shorebird.yaml, but nope!

PR #692, opened 2023-06-19
fix: forward `stdin` from shorebird to flutter process

### LINKED ISSUE #155: `shorebird run` shows output from `flutter run` but doesn't forward input
This may be the correct behavior, but it was confusing to at least one trusted tester.

We show the output from `flutter run` that talks about 'h', 'c', 'q', but if you type those they don't work.

image

I think we would need to listen to stdin and forward to the process if we wanted them to:
https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/commands/run_command.dart#L53

DarthBenro008:
## Description

- Fixes #155 
- Adds a `dart:io` dependency import to enable usage of `stdin` of the main proccess, so that the input can be captured and forwarded

I was unsure how to add tests, any help would be appreciated, but i have attached a snippet below as part of my testing :D

ezgif com-video-to-gif

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -73,6 +74,13 @@ class RunCommand extends ShorebirdCommand
       logger.err(utf8.decode(event));
     });
 
+    stdin
```
felangel: Is there any issue with just piping directly?

```dart
flutter.stdin.addStream(stdin);
```

DarthBenro008: the `addStream()` returns a future and is a blocking process (requires await). I wasn't sure if blocking threads would be the right approach, hence took a non-blocking approach.

felangel: I think we can wrap it in an unawaited. It seems unnecessary to transform and decode everything imo.

DarthBenro008: Makes sense! I have updated the commit!

felangel (approved):

PR #693, opened 2023-06-20
feat: expose --dart-define directly in `shorebird run` cli

### LINKED ISSUE #684: feat: expose --dart-define directly
**Description**

Expose the dart define option in Shorebird CLI directly since it’s commonly used.

DarthBenro008:
## Description

- Fixes #684 
- Added test cases

Usage

```
shorebird run --dart-define=KEY1=VAL1,KEY2=VAL2
```

Respective Flutter command

```
flutter run --dart-define=KEY1=VAL1 --dart-define=KEY2=VAL2
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -24,6 +24,12 @@ class RunCommand extends ShorebirdCommand
         abbr: 't',
         help: 'The main entrypoint file of the application.',
       )
+      // ignore: lines_longer_than_80_chars
+      ..addOption(
+        'dart-define',
+        help:
+            'comma seperated key-value pairs (eg: FOO=BAR,FOO1=BAR1) that will be available as constants',
+      )
```
bryanoltman:
```suggestion
      ..addOption(
        'dart-define',
        help:
            '''comma seperated key-value pairs (eg: FOO=BAR,FOO1=BAR1) that will be available as constants''',
      )
```

packages/shorebird_cli/lib/src/commands/run_command.dart
```diff
@@ -24,6 +24,12 @@ class RunCommand extends ShorebirdCommand
         abbr: 't',
         help: 'The main entrypoint file of the application.',
       )
+      // ignore: lines_longer_than_80_chars
+      ..addOption(
```
felangel: This should be a multi-option and then you don't need to split on commas etc, `package:args` handles it for you.

```suggestion
      ..addMultiOption(
```

felangel (changes requested):
see comments

packages/shorebird_cli/lib/src/commands/run_command.dart
DarthBenro008: Thank you for this! I def missed to read the docs throughly 😓 , have amended the changes!

ScottMacDougall: One comment on this, we build our apps with package:args as well, and some of our dart defines have commas in them. 

We set `splitCommas: false` on `addMultiOption` to support this, otherwise it splits on that dart-define entry's comma. I don't think dart-define supports setting values using comma separation anyway (?), so I don't think setting this to false would break anything.

felangel (approved):

PR #694, opened 2023-06-20
feat(shorebird_cli): `login:ci` command

felangel:
## Description

- feat(shorebird_cli): `login:ci` command

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -68,38 +68,63 @@ class LoggingClient extends http.BaseClient {
 }
 
 class AuthenticatedClient extends LoggingClient {
-  AuthenticatedClient({
-    required super.httpClient,
+  AuthenticatedClient.credentials({
+    required http.Client httpClient,
     required OnRefreshCredentials onRefreshCredentials,
-    required oauth2.AccessCredentials? credentials,
-    required String? ciToken,
+    required oauth2.AccessCredentials credentials,
+    RefreshCredentials refreshCredentials = oauth2.refreshCredentials,
+  }) : this._(
+          httpClient: httpClient,
+          onRefreshCredentials: onRefreshCredentials,
+          credentials: credentials,
+          refreshCredentials: refreshCredentials,
+        );
+
+  AuthenticatedClient.token({
+    required http.Client httpClient,
+    required String token,
+    OnRefreshCredentials? onRefreshCredentials,
+    RefreshCredentials refreshCredentials = oauth2.refreshCredentials,
+  }) : this._(
+          httpClient: httpClient,
+          token: token,
+          onRefreshCredentials: onRefreshCredentials,
+          refreshCredentials: refreshCredentials,
+        );
+
+  AuthenticatedClient._({
+    required super.httpClient,
+    OnRefreshCredentials? onRefreshCredentials,
+    oauth2.AccessCredentials? credentials,
+    String? token,
     RefreshCredentials refreshCredentials = oauth2.refreshCredentials,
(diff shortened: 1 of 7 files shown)


packages/shorebird_cli/lib/src/command_runner.dart
```diff
@@ -62,6 +62,7 @@ class ShorebirdCliCommandRunner extends CompletionCommandRunner<int> {
     addCommand(DoctorCommand());
     addCommand(InitCommand());
     addCommand(LoginCommand());
+    addCommand(LoginCICommand());
```
bryanoltman: Ultranit: I think the proper capitalization is `LoginCiCommand`

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -66,30 +68,75 @@ class LoggingClient extends http.BaseClient {
...
-      _credentials = await _refreshCredentials(
+    var credentials = _credentials;
+
+    if (credentials == null) {
+      final token = _token!;
+      credentials = _credentials = await _refreshCredentials(
+        _clientId,
+        oauth2.AccessCredentials(
+          // This isn't relevant for a refresh operation.
+          AccessToken('Bearer', '', DateTime.now().toUtc()),
```
eseidel: Datetime.timestamp() is .now().toUtc()

bryanoltman (approved):

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -114,21 +161,47 @@ class Auth {
...
   }
 
+  Future<AccessCredentials> loginCI(void Function(String) prompt) async {
+    final client = http.Client();
+    try {
+      final credentials = await _obtainAccessCredentials(
+        _clientId,
+        _scopes,
+        client,
+        prompt,
```
eseidel, line 191: Doesn't matter, but I think this trailing comma caused this to expand to 5 lines instead of 2 🤷

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -201,11 +274,16 @@ class Auth {
 
   String? get email => _email;
 
-  bool get isAuthenticated => _email != null;
+  bool get isAuthenticated => _email != null || _token != null;
```
eseidel, line 282: I think we're going to end up with a list of authenticators which we'll just check in order.

packages/shorebird_cli/test/src/auth/auth_test.dart
```diff
@@ -1,16 +1,18 @@
 import 'dart:convert';
-import 'dart:io';
+import 'dart:io' hide Platform;
```
eseidel, line 2: I wonder if we need dart:io at all or if there is some other mocking layer we're missing?

eseidel (approved):
lgtm.  Bryan may have helpful comments for you too.

packages/shorebird_cli/lib/src/auth/auth.dart
felangel: Without the trailing comma the indentation looks a bit off since the line wraps because it's longer than 80 characters 🤷

packages/shorebird_cli/test/src/auth/auth_test.dart
felangel: We are using it because we are creating temporary directories for the credentials instead of mocking the file system.

eseidel:
sgtm, fwiw, I've had good success with package:file in the past.

PR #705, opened 2023-06-23
feat(shorebird_cli): upload ipa as part of shorebird release ios

### LINKED ISSUE #677: feat: upload release artifacts for iOS
**Description**

We don't currently upload the release .ipa when running `shorebird release ios`. This is fine, because we don't currently do any diffing for iOS patches. However, to determine whether `shorebird release` has been run for a given platform + version, we check for release artifacts. This means `shorebird release ios` will always succeed no matter how many times it is run, and whether or not there are already patches available for a given release.

We could check for patch artifacts, but for consistency with other platforms, it would probably make more sense to upload the ipa built by `shorebird release ios`

bryanoltman:
## Description

Upload the release ipa file when running `shorebird release ios`

Fixes https://github.com/shorebirdtech/shorebird/issues/677

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -442,6 +442,29 @@ aar artifact already exists, continuing...''',
...
+    required int releaseId,
+    required String ipaPath,
+  }) async {
+    final createArtifactProgress = logger.progress('Creating artifacts');
+    final ipaFile = File(ipaPath);
+    try {
+      await codePushClient.createReleaseArtifact(
+        releaseId: releaseId,
+        artifactPath: ipaPath,
+        arch: 'aarch64',
```
```suggestion
        arch: 'ipa',
```

Is this guaranteed to be true for all ipas?

bryanoltman: Ah good catch. Changed to ipa to be in line with aabs and aars

felangel (approved):

PR #722, opened 2023-06-28
chore(shorebird_cli): improve error messaging for add collaborator 403s

bryanoltman:
## Description

Updates the error message shown to a user who cannot add collaborators from `Insufficient permissions.` to `You do not have permission to add collaborators to this app.`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_client/lib/src/code_push_client.dart
+++ b/packages/shorebird_code_push_client/lib/src/code_push_client.dart
@@ -22,6 +22,14 @@ class CodePushException implements Exception {
   String toString() => '$message${details != null ? '\n$details' : ''}';
 }
 
+/// {@template code_push_permission_exception}
+/// Exception thrown when a 403 response is received.
+/// {@endtemplate}
+class CodePushPermissionException extends CodePushException {
+  /// {@macro code_push_permission_exception}
+  CodePushPermissionException({required super.message});
+}
+
 /// {@template code_push_conflict_exception}
 /// Exception thrown when a 409 response is received.
 /// {@endtemplate}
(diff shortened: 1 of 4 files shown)


packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -104,6 +112,12 @@ class CodePushClient {
       body: json.encode(CreateAppCollaboratorRequest(email: email).toJson()),
     );
 
+    if (response.statusCode == HttpStatus.forbidden) {
```
felangel: instead of handling the error here can we surface the CodePushForbiddenException with the server error and then return a custom error in the CLI since the error is CLI-specific?

bryanoltman: I'm not sure I follow – what do you mean "return a custom error in the CLI since the error is CLI-specific"? I think this message makes sense to any client that would be attempting to add a collaborator, no?

felangel: I think the api client should ideally return what the server responds with and the CLI should be the one to interpret and override messages if we want to do that.

felangel: If it makes sense for any client then we should update the server error imo.

bryanoltman: Ah gotcha, updated. PTAL

felangel (approved):

PR #723, opened 2023-06-28
feat(code_push_protocol): adjust `UsageResponse` to support unlimited installs

felangel:
## Description

- feat(code_push_protocol): adjust `UsageResponse` to support unlimited installs

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
@@ -12,8 +12,8 @@ class GetUsageResponse {
   const GetUsageResponse({
     required this.apps,
     required this.patchInstallLimit,
-    this.currentPeriodStart,
-    this.currentPeriodEnd,
+    required this.currentPeriodStart,
+    required this.currentPeriodEnd,
   });
 
   /// Converts a Map<String, dynamic> to a [GetUsageResponse].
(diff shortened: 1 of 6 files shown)


packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
```diff
@@ -26,14 +26,15 @@ class GetUsageResponse {
...
 
-  /// The end of the current billin period.
-  final DateTime? currentPeriodEnd;
+  /// The end of the current billing period.
+  final DateTime currentPeriodEnd;
 
   /// The upper limit of patch installs for the current billing period.
-  final int patchInstallLimit;
+  /// If `null`, there is no limit.
+  final int? patchInstallLimit;
```
bryanoltman, line 37: Didn't catch this before, but what do you think about `patchDownloadLimit` instead, as that's what we're actually measuring?

bryanoltman (approved):

packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
felangel: Since installs is what we want to guard against I thought we'd leave it named as is and improve the mechanism as we go?

PR #769, opened 2023-07-05
fix(shorebird_cli): remove runInShell from bundle version commands

### LINKED ISSUE #756: fix: 'C:\Program' is not recognized as an internal or external command when shorebird install path contains space on Windows
**Description**

When I try to execute `shorebird release android` in my Windows 11, it throws
 ```
Exception: Failed to extract version name from app bundle: 'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

bryanoltman:
## Description

This addresses the issue described in https://github.com/shorebirdtech/shorebird/issues/756 (`'C:\Program' is not recognized as an internal or external command`).

Fixes https://github.com/shorebirdtech/shorebird/issues/756

I've verified that `shorebird release android` and `shorebird patch android` work in zsh on macOS and cmd.exe and Powershell on Windows.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_release_version_mixin.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_release_version_mixin.dart
@@ -39,7 +39,7 @@ mixin ShorebirdReleaseVersionMixin on ShorebirdJavaMixin {
       process.run(
         javaExecutable,
         versionNameArguments,
-        runInShell: true,
+        // runInShell: true,
         environment: {
           if (javaHome != null) 'JAVA_HOME': javaHome,
         },
@@ -47,7 +47,7 @@ mixin ShorebirdReleaseVersionMixin on ShorebirdJavaMixin {
       process.run(
         javaExecutable,
         versionCodeArguments,
-        runInShell: true,
+        // runInShell: true,
         environment: {
           if (javaHome != null) 'JAVA_HOME': javaHome,
         },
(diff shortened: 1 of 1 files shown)


eseidel (commented):
Do we know why we added this in the first place?  Can we test so we don't break this?

bryanoltman:
> Do we know why we added this in the first place?

I forget the original reasoning. Maybe @felangel knows?

> Can we test so we don't break this?

I'd have to think a bit - I'm not sure what a test would look like that would be purely tautological (i.e., "are the arguments we're passing to this function the arguments we're passing to this function"). I'm also not sure what a useful comment would be, as this change updates the call to use the default value of `runInShell`.

eseidel:
I guess I had imagined a test which tried to use the path.  I'd imagine creating a file system on disk with a space in it and trying to execute the thing.  But I guess that's hard because we'd have to have something executable checked in?

bryanoltman:
We can check an executable in, but to exercise this change, we would need to be running on Windows, which I don't think we're currently set up for. This would be pretty easy to do in an end-to-end test, but I don't believe we have any of those for the CLI at the moment.

bryanoltman:
tl;dr: I think we need end-to-end tests to meaningfully test code that interfaces with the OS using `Process.run` or `Process.start`, and we don't have those at the moment.

felangel:
> We can check an executable in, but to exercise this change, we would need to be running on Windows, which I don't think we're currently set up for. This would be pretty easy to do in an end-to-end test, but I don't believe we have any of those for the CLI at the moment.

We should use the e2e tests for this. We do have the basic structure in place for running e2e on all platforms.

bryanoltman:
Filed https://github.com/shorebirdtech/shorebird/issues/770 to add e2e tests to the CLI. I think there's a decent chunk of work involved in actually writing these tests, as they will be communicating with our backend, need to manage state (i.e., `shorebird release android` won't work twice in a row without reading external state and updating the pubspec).

felangel (approved):

PR #776, opened 2023-07-06
feat(shorebird_code_push_protocol): add ShorebirdPlan model

bryanoltman:
## Description

Adds a `ShorebirdPlan` model to `shorebird_code_push_protocol` and adds a `plan` property to the `Subscription` model. This will enable us to communicate current and available plans to the CLI and console.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
@@ -24,7 +24,7 @@ class GetUsageResponse {
   /// Converts a [GetUsageResponse] to a Map<String, dynamic>.
   Json toJson() => _$GetUsageResponseToJson(this);
 
-  /// The name of the user's plan tier.
+  /// The user's current plan.
   final ShorebirdPlan plan;
 
   /// The usage per app.
--- a/packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
@@ -10,7 +10,7 @@ part 'get_usage_response.g.dart';
 class GetUsageResponse {
   /// {@macro get_usage_response}
   const GetUsageResponse({
-    required this.tierName,
+    required this.plan,
     required this.apps,
     required this.currentPeriodStart,
     required this.currentPeriodEnd,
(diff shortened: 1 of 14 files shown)


packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
```diff
@@ -23,6 +24,9 @@ class GetUsageResponse {
   /// Converts a [GetUsageResponse] to a Map<String, dynamic>.
   Json toJson() => _$GetUsageResponseToJson(this);
 
+  /// The name of the user's plan tier.
```
felangel:
```suggestion
  /// The user's current plan.
```

packages/shorebird_code_push_protocol/lib/src/models/shorebird_plan.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:json_annotation/json_annotation.dart';
+import 'package:shorebird_code_push_protocol/shorebird_code_push_protocol.dart';
+
+part 'shorebird_plan.g.dart';
+
+/// {@template shorebird_plan}
+/// A Shorebird plan.
+/// {@endtemplate}
+@JsonSerializable()
+class ShorebirdPlan {
```
felangel, line 10: Any reason why we diverged from the existing plan model on the backend? https://github.com/shorebirdtech/_shorebird/blob/main/packages/code_push_server/lib/src/plans/shorebird_plan.dart

The main reason I ask is introducing a maxTeamSize for example will affect the logic on the backend since we'll need to query for the number of existing collaborators which is not something the backend currently does or has to do.

bryanoltman: I can be convinced otherwise, but the idea behind this was to allow us to separate things the backend needs (stripe price id, for example) from things the client needs. The client doesn't (and shouldn't) have a notion of the StripeEnvironment, as another example.

bryanoltman: (this is the same thinking behind duplicating something like AppMetadata)

felangel: Yeah I understand that I'm specifically wondering about the maxTeamSize vs collaboratorsEnabled

felangel (approved):

packages/shorebird_code_push_protocol/lib/src/models/shorebird_plan.dart
bryanoltman: `maxTeamSize` is more flexible than `collaboratorsEnabled`. Given our current messaging on shorebird.dev ("1 developer"), I can imagine a world where we introduce a new tier that supports teams of up to, say, 3.

PR #820, opened 2023-07-11
chore(shorebird_code_push_protocol): use Money object instead of int for dollar values

bryanoltman:
## Description

To avoid issues like the one fixed by https://github.com/shorebirdtech/shorebird/pull/795 going forward, use a dedicated type to represent monetary values.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/extensions/money2_extensions.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/extensions/money2_extensions.dart
@@ -1,7 +1,0 @@
-import 'package:money2/money2.dart';
-
-/// Defines currencies used by Shorebird.
-extension ShorebirdCurrency on Currency {
-  /// The US dollar.
-  static Currency get usd => Currency.create('USD', 2);
-}
--- a/packages/shorebird_code_push_protocol/lib/src/extensions/money2_extensions.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/extensions/money2_extensions.dart
@@ -1,0 +1,7 @@
+import 'package:money2/money2.dart';
+
+/// Defines currencies used by Shorebird.
+extension ShorebirdCurrency on Currency {
+  /// The US dollar.
+  static Currency get usd => Currency.create('USD', 2);
+}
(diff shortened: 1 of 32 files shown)


packages/shorebird_code_push_protocol/lib/src/extensions/money2_extensions.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:money2/money2.dart';
+
+/// Defines currencies used by Shorebird.
+extension ShorebirdCurrency on Currency {
```
eseidel: I'm confused why this is an extension only to have a static?  A normal class would have worked if your goal was to namespace the usd name?

packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
```diff
@@ -46,6 +47,16 @@ class GetUsageResponse {
   final int? patchInstallLimit;
 }
 
+/// Converts [currentPeriodCost] to a [Money] object.
+extension GetUsageResponseMoney on GetUsageResponse {
+  /// [currentPeriodCost] as a [Money] object. Prefer this over using
+  /// [currentPeriodCost] directly.
+  Money get currentPeriodCostMoney => Money.fromIntWithCurrency(
+        currentPeriodCost,
```
eseidel: Can we hide/remove currentPeriodCost?

eseidel (approved):
Felix is probably a better reviewer, I now realize, but lgtm

packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
bryanoltman: We could rename it, but I don't think it's possible to hide. I opted to leave it as-is to avoid making a breaking change. @felangel open to your thoughts on this.

felangel: I think it'd be best to use a custom `JsonConverter` so that json_serializable handles this for us. It's fine to make it a breaking API change imo.

packages/shorebird_code_push_protocol/lib/src/extensions/money2_extensions.dart
bryanoltman: Just made `usd` a top-level variable

packages/shorebird_code_push_protocol/lib/src/messages/get_usage/get_usage_response.dart
bryanoltman: Done.

felangel (approved):

PR #826, opened 2023-07-11
feat(shorebird_code_push_protocol): add RecordPatchInstallRequest type

bryanoltman:
## Description

Introduces a `RecordPatchInstallRequest` type to record successful patch installs on client devices.

Part of https://github.com/shorebirdtech/_shorebird/issues/281

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


felangel (approved):

bryanoltman:
Moving this to a draft state until we decide how we want to charge for overages.

bryanoltman:
@felangel let me know if there's any information we'd want in the logs that isn't in this request. My thinking was that we could just log this request as json and get all the info we need from that.

felangel (approved):

felangel:
> @felangel let me know if there's any information we'd want in the logs that isn't in this request. My thinking was that we could just log this request as json and get all the info we need from that.

I think we'd need more info (client_id, app_id, release_id, platform, arch, patch_id, timestamp) but we can get the remaining info on the backend.

bryanoltman:
- client_id – in request
- app_id – in request
- release_id – added to request
- platform – added to request
- arch – added to request
- patch_id – I think this can be easily queried for if we need it, not sure the updater has this info currently
- timestamp – It would probably be better to get this from the request log, as there's not a guarantee that user devices will have their clocks set correctly.

PR #845, opened 2023-07-13
feat(shorebird_cli): add `shorebird preview` command

### LINKED ISSUE #780: New developers often get to a HashMismatched error
I'm not entirely sure what we should do here.  Whether just make the rust logging much more clear what's going on?  Somehow fix the system in some other way (like `shorebird run` failing when it sees you running version X and having a libapp.so hash which doesn't match?  May also need to write docs which the Rust code points to?  Regardless, it's a too-common failure mode for new users.

### LINKED ISSUE #802: Replace `shorebird run` with `shorebird preview` to allow running published releases

felangel:
## Description

- feat(shorebird_cli): add `shorebird preview` command (closes #802, closes #780)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/preview_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/preview_command.dart
@@ -1,0 +1,173 @@
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:mason_logger/mason_logger.dart';
+import 'package:path/path.dart' as p;
+import 'package:shorebird_cli/src/adb.dart';
+import 'package:shorebird_cli/src/bundletool.dart';
+import 'package:shorebird_cli/src/cache.dart';
+import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
+import 'package:shorebird_cli/src/command.dart';
+import 'package:shorebird_cli/src/logger.dart';
+import 'package:shorebird_cli/src/shorebird_config_mixin.dart';
+import 'package:shorebird_cli/src/shorebird_validation_mixin.dart';
+import 'package:shorebird_cli/src/third_party/flutter_tools/lib/flutter_tools.dart';
+
+/// {@template preview_command}
+/// `shorebird preview` command.
+/// {@endtemplate}
+class PreviewCommand extends ShorebirdCommand
+    with ShorebirdConfigMixin, ShorebirdValidationMixin {
+  /// {@macro preview_command}
+  PreviewCommand({super.validators}) {
+    argParser
+      ..addOption(
+        'device-id',
+        help: 'The ID of the device or simulator to preview the release on.',
+      )
+      ..addOption(
+        'app-id',
+        help: 'The ID of the app to preview the release for.',
+      )
+      ..addOption(
+        'release-version',
+        help: 'The version of the release (e.g. "1.0.0").',
+      );
+  }
+
+  @override
+  String get name => 'preview';
+
+  @override
+  String get description => 'Preview a specific release on a device.';
+
+  @override
+  Future<int> run() async {
+    try {
+      await validatePreconditions(checkUserIsAuthenticated: true);
+    } on PreconditionFailedException catch (error) {
+      return error.exitCode.code;
+    }
+
+    Future<String> promptForApp() async {
+      final apps = await codePushClientWrapper.getApps();
+      final app = logger.chooseOne(
+        'Which app would you like to preview?',
+        choices: apps,
+        display: (app) => app.displayName,
+      );
+      return app.appId;
+    }
+
+    Future<String> promptForReleaseVersion(String appId) async {
+      final releases = await codePushClientWrapper.getReleases(appId: appId);
+      final release = logger.chooseOne(
+        'Which release would you like to preview?',
+        choices: releases,
+        display: (release) => release.version,
+      );
+      return release.version;
+    }
+
+    const platform = 'android';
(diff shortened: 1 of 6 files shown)


packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+      if (apps.isEmpty) return null;
+      final app = logger.chooseOne(
+        'Which app would you like to preview?',
+        choices: apps,
+        display: (app) => app.displayName,
+      );
+      return app.appId;
+    }
+
+    Future<String?> promptForReleaseVersion(String appId) async {
```
bryanoltman: Same here

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+
+  @override
+  Future<int> run() async {
+    try {
+      await validatePreconditions(checkUserIsAuthenticated: true);
+    } on PreconditionFailedException catch (error) {
+      return error.exitCode.code;
+    }
+
+    Future<String?> promptForApp() async {
```
bryanoltman: Can this be defined outside of the run() function? Functions in functions disrupt the control flow and make the code a bit harder to parse.

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+      if (releases.isEmpty) return null;
+      final release = logger.chooseOne(
+        'Which release would you like to preview?',
+        choices: releases,
+        display: (release) => release.version,
+      );
+      return release.version;
+    }
+
+    const platform = 'android';
```
bryanoltman, line 52: Not something to tackle in this PR, but we should probably make `platform` an enum instead of a string throughout

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+    final appId = results['app-id'] as String? ?? await promptForApp();
+
+    if (appId == null) {
+      logger.info('No apps found');
+      return ExitCode.success.code;
```
bryanoltman, line 57: Is success correct here? `usage` or `data` might be better

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    final releaseVersion = results['release-version'] as String? ??
+        await promptForReleaseVersion(appId);
+
+    if (releaseVersion == null) {
+      logger.info('No releases found');
+      return ExitCode.success.code;
```
bryanoltman, line 65: Same re: exit code

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    final previewDirectory = cache.getPreviewDirectory(appId);
+    final aabPath = p.join(
+      previewDirectory.path,
+      '${platform}_$releaseVersion.aab',
+    );
+    final apksPath = p.join(
```
bryanoltman: TIL that `apks` was a thing

bryanoltman: It would probably be better to define this further down, closer to where it's first used (I think line 133)

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -0,0 +1,186 @@
...
+      } catch (error) {
+        downloadArtifactProgress.fail('$error');
+        return ExitCode.software.code;
+      }
+    }
+
+    final extractMetadataProgress = logger.progress('Extracting metadata');
+    late String package;
+    try {
+      package = await bundletool.getPackageName(aabPath);
```
bryanoltman, line 100: Are we downloading the aab for any reason other than to get the package name?

packages/shorebird_cli/lib/src/commands/preview_command.dart
felangel: I think it's technically a success you just didn't have any apps. The user didn't use the command incorrectly imo 🤷

packages/shorebird_cli/lib/src/commands/preview_command.dart
felangel: we need it to generate the apks which we install/run on the device.

packages/shorebird_cli/lib/src/commands/preview_command.dart
felangel: yup agreed that needs to be done throughout

packages/shorebird_cli/lib/src/commands/preview_command.dart
felangel: done ✅

eseidel (approved):

packages/shorebird_cli/lib/src/commands/preview_command.dart
bryanoltman: Maybe? If I run `git checkout asdf` and have no `asdf` branch, it exits with a status of `1`, but I don't feel strongly

PR #891, opened 2023-07-23
fix(shorebird_cli): print more friendly error if cli upgrade is required

### LINKED ISSUE #884: fix: message that code push client is out-of-date is not clear
**Description**

When we push a breaking change to the code push protocol, the error message tells the user about `code_push_client`. This is confusing, it should instead tell the user to run `shorebird upgrade`.

bryanoltman:
## Description

When the CodePushClient throws a CodePushUpgradeRequiredException, print an error telling the user to run `shorebird upgrade` instead of the server-provided message about an out of date code push client version.

Screenshot 2023-07-22 at 10 52 24 PM

Fixes https://github.com/shorebirdtech/shorebird/issues/884

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -67,8 +67,7 @@ class CodePushClientWrapper {
       fetchAppsProgress.complete();
       return apps;
     } catch (error) {
-      fetchAppsProgress.fail('$error');
-      exit(ExitCode.software.code);
+      _handleErrorAndExit(error, progress: fetchAppsProgress);
     }
   }
 
@@ -105,8 +104,7 @@ This app may not exist or you may not have permission to view it.''',
       fetchChannelsProgress.complete();
       return channel;
     } catch (error) {
-      fetchChannelsProgress.fail('$error');
-      exit(ExitCode.software.code);
+      _handleErrorAndExit(error, progress: fetchChannelsProgress);
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -615,4 +618,26 @@ aar artifact already exists, continuing...''',
 
     await promotePatch(appId: appId, patchId: patch.id, channel: channel);
   }
+
+  /// Prints an appropriate error message for the given error and exits with
+  /// code 70. If [progress] is provided, it will be failed with the given
+  /// [message] or [error.toString()] if [message] is null.
+  Never _handleErrorAndExit(
```
felangel, line 625: Couldn’t we have handled the upgrade required case in the code_push_client just like the other exceptions and overridden the message there? I think that would have been a less invasive change and would’ve been more consistent with existing custom error handling?

bryanoltman: We could have, but my understanding of the code push client design was that it was intentionally decoupled from the CLI, so it wouldn't make sense to reference CLI commands there.

bryanoltman: (and I think it's important that we tell the user explicitly what command they need to run, as we do elsewhere in the CLI)

felangel: ah yeah misread this earlier, totally makes sense to have this in the CLI 👍

felangel (approved):

PR #904, opened 2023-07-25
fix(shorebird_cli): use platform status instead of artifacts when checking existing release

bryanoltman:
## Description

Instead of checking for existing artifacts (which might yield a false positive if a release command was terminated or otherwise failed before it completed), check the release's release status for the given platform.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -129,8 +129,8 @@ This app may not exist or you may not have permission to view it.''',
     }
   }
 
-  /// Checks whether [release] is in an active state for [platform]. If so,
-  /// exits with code 70.
+  /// Prints an error message and exits with code 70 if [release] is in an
+  /// active state for [platform].
   void ensureReleaseIsIsNotActive({
     required Release release,
     required ReleasePlatform platform,
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -131,7 +131,7 @@ This app may not exist or you may not have permission to view it.''',
 
   /// Prints an error message and exits with code 70 if [release] is in an
   /// active state for [platform].
-  void ensureReleaseIsIsNotActive({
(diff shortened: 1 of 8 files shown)


packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -129,31 +129,16 @@ This app may not exist or you may not have permission to view it.''',
...
   }
 
-  /// Exits if [platform] release artifacts already exist for an
-  /// [existingRelease].
-  Future<void> ensureReleaseHasNoArtifacts({
-    required String appId,
-    required Release existingRelease,
+  /// Checks whether [release] is in an active state for [platform]. If so,
+  /// exits with code 70.
+  void ensureReleaseIsIsNotActive({
```
```suggestion
  void ensureReleaseIsNotActive({
```

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -129,31 +129,16 @@ This app may not exist or you may not have permission to view it.''',
     }
   }
 
-  /// Exits if [platform] release artifacts already exist for an
-  /// [existingRelease].
-  Future<void> ensureReleaseHasNoArtifacts({
-    required String appId,
-    required Release existingRelease,
+  /// Checks whether [release] is in an active state for [platform]. If so,
```
felangel, line 132: This is a bit confusing, does it exit with code 70 if it's active or draft?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: This isn't quite true – a release may not have a status for the given platform

felangel: Ah I see in that case can we remove the extra `Is` -> `ensureReleaseIsNotActive`

bryanoltman: Hahaha yes we can 😳

felangel (approved):
💯

PR #916, opened 2023-07-26
fix: iOS status link is broken

eseidel:
Also moved this to be a shared function in a new shared file.
I suspect we'll end up with ios/ and android/ directories
for sharing files like this.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/ios.dart
+++ b/packages/shorebird_cli/lib/src/ios.dart
@@ -2,7 +2,8 @@ import 'package:mason_logger/mason_logger.dart';
 import 'package:shorebird_cli/src/logger.dart';
 
 void showiOSStatusWarning() {
-  final url =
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/ios.dart
```diff
@@ -0,0 +1,8 @@
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/logger.dart';
+
+void showiOSStatusWarning() {
+  final url =
+      link(uri: Uri.parse('https://docs.shorebird.dev/status#ios-alpha'));
```
bryanoltman:
```suggestion
  final url = link(
    uri: Uri.parse('https://docs.shorebird.dev/status#ios-alpha'),
  );
```

bryanoltman (approved):

PR #934, opened 2023-07-27
fix(shorebird_cli): don't rely on display-friendly git status output

### LINKED ISSUE #925: fix: commands that rely on git status output fail when run in non-US English locales.
**Description**

A user whose system uses Korean as its language is seeing a shorebird doctor failure because we rely on an English output message from git status. 

https://unix.stackexchange.com/questions/155046/determine-if-git-working-directory-is-clean-from-a-script looks like a promising solution.

bryanoltman:
## Description

Uses `git status --untracked-files=no --porcelain` instead of `git status` to avoid reliance on a specific English-language string in console output.

Fixes https://github.com/shorebirdtech/shorebird/issues/925.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
+++ b/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
@@ -114,12 +114,10 @@ This can cause unexpected behavior if you are switching between the tools and th
   Future<bool> _flutterDirectoryIsClean(ShorebirdProcess process) async {
     final result = await process.run(
       'git',
-      ['status'],
+      ['status', '--untracked-files=no', '--porcelain'],
       workingDirectory: ShorebirdEnvironment.flutterDirectory.path,
     );
-    return result.stdout
-        .toString()
-        .contains('nothing to commit, working tree clean');
+    return result.stdout.toString().isEmpty;
   }
 
   Future<String> _shorebirdFlutterVersion(ShorebirdProcess process) {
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
```diff
@@ -114,12 +114,10 @@ This can cause unexpected behavior if you are switching between the tools and th
...
     final result = await process.run(
       'git',
-      ['status'],
+      ['status', '--untracked-files=no', '--porcelain'],
       workingDirectory: ShorebirdEnvironment.flutterDirectory.path,
     );
-    return result.stdout
-        .toString()
-        .contains('nothing to commit, working tree clean');
+    return result.stdout.toString().isEmpty;
```
eseidel: There aren't even newlines in it? :)

eseidel (approved):

packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
bryanoltman: Nope. Added a call to `trim` to be safe, but I don't believe it's necessary.

eseidel:
Now that trim will be carried forward by generations, wondering on what system they needed to trim. 🤣

PR #945, opened 2023-07-28
chore(shorebird_cli): add http.ClientException to list of retry-able errors

bryanoltman:


packages/shorebird_cli/test/src/http_client/retrying_client_test.dart
```diff
@@ -44,6 +44,13 @@ void main() {
         );
       });
 
+      test('returns true on WebSocketException', () {
```
felangel:
```suggestion
      test('returns true on ClientException', () {
```

felangel (approved):

packages/shorebird_cli/test/src/http_client/retrying_client_test.dart
bryanoltman: Fixed

Issue #910, opened 2023-07-25
feat: Support flavors on iOS


eseidel:
To match parity with Android.

M97Chahboun:
Hi guys, I want contribute by implement this issue but the flavor already passed and used on [buildIpa method](https://github.com/shorebirdtech/shorebird/blob/4ac96fd5147daeda983bd815a2281e7f572f060a/packages/shorebird_cli/lib/src/commands/release/release_ios_command.dart#L98)...

felangel:
> Hi guys, I want contribute by implement this issue but the flavor already passed and used on [buildIpa method](https://github.com/shorebirdtech/shorebird/blob/4ac96fd5147daeda983bd815a2281e7f572f060a/packages/shorebird_cli/lib/src/commands/release/release_ios_command.dart#L98)...

I’m working on this and should have it ready sometime in the next few days 🙂

M97Chahboun:
@felangel OK, Good luck 👍🏼

PR #966, opened 2023-08-01
fix(shorebird_cli): `shorebird init` iOS flavor detection on non-macOS

felangel:
## Description

- fix(shorebird_cli): `shorebird init` iOS flavor detection on non-macOS (part of #910)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -203,6 +203,7 @@ For more information about Shorebird, visit ${link(uri: Uri.parse('https://shore
           .where((e) => p.basename(e.path).endsWith('.xcscheme'))
           .where((e) => p.basenameWithoutExtension(e.path) != 'Runner')
           .map((file) => p.basename(file.path).split('.xcscheme').first)
+          .sorted()
           .toSet();
     }
   }
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -2,10 +2,12 @@ import 'dart:io';
 
 import 'package:collection/collection.dart';
 import 'package:mason_logger/mason_logger.dart';
+import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/doctor.dart';
 import 'package:shorebird_cli/src/gradlew.dart';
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -172,11 +174,35 @@ For more information about Shorebird, visit ${link(uri: Uri.parse('https://shore
...
+    if (platform.isMacOS) {
+      try {
+        final info = await xcodeBuild.list(projectPath);
+        return info.schemes.whereNot((element) => element == 'Runner').toSet();
+      } on MissingIOSProjectException {
+        return null;
+      }
+    } else {
+      // When running on a non-macOS platform, we can't use xcodebuild to
+      // detect flavors so we fallback to looking for schemes in xcschemes.
```
bryanoltman: Worth noting here that this seems to be exactly the behavior of xcodebuild

packages/shorebird_cli/test/src/commands/init_command_test.dart
```diff
@@ -247,6 +255,106 @@ If you want to reinitialize Shorebird, please run "shorebird init --force".''',
       expect(exitCode, ExitCode.software.code);
     });
 
+    test('throws software error when unable to detect schemes (non-macos)',
+        () async {
+      when(() => platform.isMacOS).thenReturn(false);
```
bryanoltman: This is stylistic and completely optional, but I've found that this kind of test becomes slightly more readable when put in groups:

```
group('on macOS', () {
  setUp(() {
    when(() => platform.isMacOS).thenReturn(true);
  });

  test('throws software error when unable to detect schemes', ...)
});

group('on a non-mac OS', () {
  setUp(() {
    when(() => platform.isMacOS).thenReturn(false);
  });

  test('throws software error when unable to detect schemes', ...)
});
```

bryanoltman (approved):

packages/shorebird_cli/test/src/commands/init_command_test.dart
felangel: yup good call will reorganize.

packages/shorebird_cli/lib/src/commands/init_command.dart
felangel: done ✅

Issue #1037, opened 2023-08-04
feat: shorebird should support switching flutter versions


felangel:
`shorebird` does not need to be tightly coupled to a single `flutter` release.  We could offer `fvm`-like functionality, of downloading different versions of Flutter per project.

If we did this, then it would be possible for us to release a new version of Shorebird, which depend on a new version of Flutter without worrying that users then have to take explicit action to downgrade their version of Shorebird in order to manage older apps in the wild (patching already works this way but releasing currently requires being on the latest stable version).

Presumably if we did this we would only support a limited set of older flutter revisions, and only back to some certain point, as we do currently maintain (very small) patches on top of Flutter which we might have to also maintain on older branches of Flutter in this model.

felangel:
related to #472

PR #1063, opened 2023-08-08
feat(shorebird_cli): add `getVersions` to `ShorebirdFlutter`

felangel:
## Description

- feat(shorebird_cli): add `getVersions` to `ShorebirdFlutter` (part of #1037)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -60,4 +61,15 @@ class ShorebirdFlutter {
     );
     return status.isEmpty;
   }
+
+  Future<List<String>> getVersions({String? revision}) async {
+    final result = await git.forEachRef(
+      format: '%(refname:short)',
```
bryanoltman, line 67: Q: is this format string something we want the caller to be responsible for or should it be owned by `git`? lgtm either way, just curious.

bryanoltman (approved):

packages/shorebird_cli/lib/src/shorebird_flutter.dart
felangel: I left it as is for the sake of flexibility but happy to adjust in the future if we find it's more convenient to keep it internal 👍

PR #1068, opened 2023-08-08
chore(shorebird_cli): add windows and mac to CI test matrix

bryanoltman:


eseidel:
Is this still an active patch?

bryanoltman:
@felangel do we still want to do this? I had forgotten about this PR and I think it might help, but I'm sort of on the fence re: how much value it will provide if we can't combine coverage data between different OS runs

eseidel:
Is this still needed?

bryanoltman:
This has gotten stale and I'm not sure that it would add a ton of value atm. Closing.

PR #1070, opened 2023-08-08
feat(shorebird_cli): add `shorebird flutter version use`

### LINKED ISSUE #1037: feat: shorebird should support switching flutter versions

felangel:
## Description

- feat(shorebird_cli): add `shorebird flutter version use` (closes #1037)

https://github.com/shorebirdtech/shorebird/assets/8855632/6ad8eb68-4f00-45fa-9a97-c119de5aa82b

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
@@ -1,0 +1,74 @@
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/command.dart';
+import 'package:shorebird_cli/src/logger.dart';
+import 'package:shorebird_cli/src/shorebird_flutter.dart';
+
+/// {@template flutter_versions_use_command}
+/// `shorebird flutter versions use`
+/// Use a different Flutter version.
+/// {@endtemplate}
+class FlutterVersionsUseCommand extends ShorebirdCommand {
+  /// {@macro flutter_versions_use_command}
+  FlutterVersionsUseCommand();
+
+  @override
+  String get description => 'Use a different Flutter version.';
+
+  @override
+  String get name => 'use';
+
+  @override
+  Future<int> run() async {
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
```diff
@@ -0,0 +1,74 @@
...
+    } catch (error) {
+      fetchFlutterVersionsProgress.fail('Failed to fetch Flutter versions.');
+      logger.err('$error');
+      return ExitCode.software.code;
+    }
+
+    if (!versions.contains(version)) {
+      logger.err('''
+Version $version not found.
+Use `shorebird flutter versions list` to list available versions.''');
```
bryanoltman, line 56: What do you think about including a link to our [feature request template](https://github.com/shorebirdtech/shorebird/issues/new?assignees=&labels=feature&projects=&template=feature_request.md&title=feat%3A+) to tell people they can request specific versions?

bryanoltman (approved):
Looks great!

packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
felangel: Ah yeah great call! Will do in the next PR if that's cool 👍

PR #1099, opened 2023-08-15
feat(shorebird_cli): don't prompt if environment variable detects CI

### LINKED ISSUE #1097: feat: `shorebird` commands should not hang when run in CI
Right now your options are to run `shorebird` with no flags and it will hang in CI when it hits a prompt.

Or you can run it with `--force` and it will say "yes" to every prompt.

1. We should never hang (maybe that's a mason_logger fix to detect non-interactive and always return the defaults?)
2. We should support failing instead of --force, so that CI can fail on a prompt (e.g. "this contains unpatchable changes are you sure you want to continue?") rather than hanging or continuing.

bryanoltman:
## Description

Skips confirmation prompts if the `CONTINUOUS_INTEGRATION` environment variable is `"true"`

Fixes https://github.com/shorebirdtech/shorebird/issues/1097

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/shorebird_env.dart
```diff
@@ -169,4 +169,8 @@ class ShorebirdEnv {
       return null;
     }
   }
+
+  /// Whether the CONTINUOUS_INTEGRATION environment variable is set to true.
+  bool get isRunningOnCI =>
```
eseidel, line 177: `flutter` tool has some fancier logic here.
https://github.com/search?q=repo%3Aflutter%2Fflutter%20CONTINUOUS_INTEGRATION&type=code

eseidel: What a pile of 💩 
https://stackoverflow.com/questions/9738535/powershell-test-for-noninteractive-mode
https://stackoverflow.com/questions/6108330/checking-for-interactive-shell-in-a-python-script

bryanoltman: I saw this (and this is where I got the approach and `CONTINUOUS_INTEGRATION` flag name from), but wasn't sure how much of it we wanted to follow. Should I just plop the flutter file in third_party, or is checking for all the same keys in our source acceptable?

eseidel: Checking the same keys is acceptable.

bryanoltman: Updated!

packages/shorebird_cli/lib/src/patch_diff_checker.dart
```diff
@@ -67,7 +69,8 @@ class PatchDiffChecker {
             archiveDiffer.nativeFileSetDiff(contentDiffs).prettyString,
           ),
         );
-      final shouldContinue = force || logger.confirm('Continue anyways?');
+      final shouldContinue =
```
felangel: is this logic correct? I thought we wanted this to exit in CI instead of continuing.

bryanoltman: That's what it does. `shouldConfirm` is basically "we're not on CI". If the user doesn't pass the `--force` flag and `shouldConfirm` is false, `shouldContinue` will be false. Reworked a bit for clarity.

bryanoltman: This line is also a bit dense – I wouldn't be opposed to further refactoring it out into a few separate checks.

felangel (approved):

PR #1112, opened 2023-08-17
feat(shorebird_cli): `shorebird flutter versions use <revision>`

### LINKED ISSUE #1098: shorebird patch should not require release version
Assumptions:
- We need to know what release to base a patch on.
- The way we identify releases is via version number (since they're enforced unique by the stores)
- The only way to know a version number is to build a full app (since gradel or XCode? can change the version number during a build, it's not just the pubspec.yaml number).

Previously (before we supported multiple flutter versions) we would just try to build your current patch, grab the version number from the built product, and then look up the corresponding release on Shorebird's servers.

As part of adding multiple-version support, we moved to instead prompting the user for which release they were trying to patch.  This not only saved us a build, but also stopped problems of the user having changed the release number (I'm not sure if we later error if it's changed still?).

However this prompting broke CI builds.  Since there was no longer an easy way from the source to determine what release it was patching against.

We could add back auto-detection from the built product, but now we have the additional complexity of Flutter versions.  Which Flutter version should we build the patch with?  Whichever one you currently have selected?  Probably.

I think the root of this ugliness is that it's not possible to know the release number w/o building.

But I think probably our best bet is to:

1. Assume the user has already configured the correct flutter version (with `shorebird flutter version`).
2. Attempt to build the current sources with that flutter version.  (Maybe if the build fails, explain that they may not have selected the proper flutter version for this app, and show instructions on how to do so?)
3. Attempt to look up a corresponding release to the version number in the built product.

felangel:
## Description

- feat(shorebird_cli): `shorebird flutter versions use <revision>` (closes #1098)
  - this allows users to specify a specific git hash when switching flutter versions in addition to a semantic version.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
@@ -11,6 +11,8 @@ class FlutterVersionsUseCommand extends ShorebirdCommand {
   /// {@macro flutter_versions_use_command}
   FlutterVersionsUseCommand();
 
+  static final RegExp _shaRegExp = RegExp(r'\b([a-f0-9]{40})\b');
+
   @override
   String get description => 'Use a different Flutter version.';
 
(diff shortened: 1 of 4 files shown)


bryanoltman (approved):
I'd update the title/description of the PR to say what this does. Maybe "allows user to specify Flutter version using SHA hash instead of version number"?

PR #1133, opened 2023-08-18
fix(shorebird_cli): `shorebird doctor` flutter version validator

felangel:
## Description

- fix(shorebird_cli): `shorebird doctor` flutter version validator
  - previously we were using points at which was not necessarily accurate since additional hotfixes can be applied to a flutter_release branch so switching this to use `--contains` instead which should be what we want since we're cherry-picking hot-fixes to release branches.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/git.dart
+++ b/packages/shorebird_cli/lib/src/git.dart
@@ -93,11 +93,11 @@ class Git {
     required String directory,
     required String format,
     required String pattern,
-    String? pointsAt,
+    String? contains,
   }) async {
     final arguments = [
       'for-each-ref',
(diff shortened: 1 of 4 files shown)


eseidel (approved):
I don't really know what this does, but rslgtm.

felangel:
> I don't really know what this does, but rslgtm.

It adjusts the git command we use to determine the release branch we’re using to use --contains instead of --points-at so that if you’re on a revision and we apply a hot fix, shorebird can still determine which release branch your revision is part of.

PR #1146, opened 2023-08-23
fix(shorebird_cli): exit with non-zero code if unpatchable changes are detected on CI

### LINKED ISSUE #1145: fix: unpatchable warnings exit with code 0 on CI
This should use a non-zero exit code (and probably print "Exiting..." or something similar).

bryanoltman:
## Description

If an unpatchable change is detected in a non-interactive environment (i.e., CI), throw an exception and exit with code 70.

Fixes https://github.com/shorebirdtech/shorebird/issues/1145

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
```diff
@@ -206,18 +206,24 @@ Please re-run the release command for this version or create a new release.''');
...
-      ),
-      releaseArtifactUrl: Uri.parse(releaseAarArtifact.url),
-      archiveDiffer: _archiveDiffer,
-      force: force,
-    );
+        releaseArtifactUrl: Uri.parse(releaseAarArtifact.url),
+        archiveDiffer: _archiveDiffer,
+        force: force,
+      );
+    } on UnpatchableChangeException {
```
eseidel, line 223: I'm trying to think how I would do this in Rust.

I suspect what you may end up wanting is for both the normal exit and the failed exit to both be handled via exception and whatever caller to this be able to handle such an exception.

Then callers just look like this:

```
await patchDiffChecker.confirmUnpatchableDiffsIfNecessary();
```

And some outer layer knows about:

try {
command()
} on ExitException catch (e) {
  return e.exitCode;
}

Or similar?

bryanoltman: I like that. Just pushed a commit to standardize the control flow.

eseidel (approved):

PR #1154, opened 2023-08-24
fix(e2e): scoped logger in `LoggingClient`

felangel:
## Description

- fix(e2e): scoped logger in `LoggingClient`
  - e2e tests were failing because the `Auth` instance has a static default `LoggingClient` which has a dependency on `logger` and the lookup was failing.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart
+++ b/packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart
@@ -1,11 +1,13 @@
 import 'dart:io';
 
 import 'package:checked_yaml/checked_yaml.dart';
+import 'package:http/http.dart' as http;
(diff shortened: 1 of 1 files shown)


bryanoltman:
I'm a little confused – the title of this PR doesn't seem to match what it's doing. Can you include a description of why the change is being made?

felangel:
> I'm a little confused – the title of this PR doesn't seem to match what it's doing. Can you include a description of why the change is being made?

Yup, just updated. The static default client created by `Auth` isn't run in the correct zone so the logger lookup fails in e2e.

bryanoltman (approved):

PR #1159, opened 2023-08-24
feat: Use codecov's coverage "carryforward" flags

eseidel:
This allows us to not run every test on every commit and still
get coverage.
https://docs.codecov.com/docs/flags
https://docs.codecov.com/docs/carryforward-flags
https://docs.codecov.com/docs/codecov-yaml


felangel (approved):
🎉

PR #1161, opened 2023-08-24
test: rewrite cutler to be testable

eseidel:
I removed the 'rebase' command for now since it was dead.
'rebase' should be re-written later to just print the commands
needed since having it actually drive the rebase was a dream.
I separated "Repo" from "Checkout" and removed the global config object.
I moved to using a global logger and package scoped to match
other packages.
I added a bunch of tests.  This still doesn't have 100% coverage
but it should
be 100% testable now.


packages/cutler/lib/git_extensions.dart
```diff
@@ -50,20 +50,35 @@ extension RepoCommands on Repo {
...
       'git',
-      ['fetch', '--all'],
+      ['fetch', '--all', '--tags'],
       workingDirectory: workingDirectory,
     );
   }
 
   /// Returns a [Version] for the given [hash].
-  Version versionFrom(String hash, {bool lookupTags = true}) {
+  Version versionFrom(String commitish, {bool lookupTags = true}) {
```
bryanoltman: > commitish

Typo?

bryanoltman (approved):

packages/cutler/lib/git_extensions.dart
eseidel: https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish

bryanoltman: Huh! TIL

bryanoltman: Probably should update the doc comment to reflect the argument rename

eseidel: Fixed.

eseidel:
I ended up doing a massive rewrite of cutler.

felangel (approved):

PR #1168, opened 2023-08-25
fix: codecov does not allow '/' in the name of coverage flags

eseidel:
So fixed by using a stack overflow answer of how to split a
path in GitHub actions.  Now we only pass the package name
not the entire path.

Could consider moving this out of the action, and have the package
name be an input?


.github/actions/dart_package/action.yaml
```diff
@@ -67,10 +67,18 @@ runs:
         dart pub global activate coverage
         dart test -j ${{inputs.concurrency}} --coverage=coverage --platform=${{inputs.platform}} && dart pub global run coverage:format_coverage --lcov --in=coverage --out=coverage/lcov.info --packages=.dart_tool/package_config.json --report-on=${{inputs.report_on}} --check-ignore
 
+    # Per https://stackoverflow.com/questions/73402042/github-action-expressions-split-string
+    - name: Split package name
+      env:
+        PACKAGE_PATH: ${{ github.ref_name }}
+      id: split
+      shell: ${{ inputs.shell }}
+      run: echo "::set-output name=package_name::${PACKAGE_PATH##*/}"
```
felangel: I think set-output is deprecated
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

eseidel: Fixed.

felangel (approved):

eseidel:
Landing as this is strictly better than what we have now on main.  I will separately look into why cutler isn't being held to only 10%.

PR #1175, opened 2023-08-28
chore(shorebird_cli): add the concept of an extension command to Shorebird CLI

zaitrarrio:
An extension command follows the Git model of CLI extension where any command unknown to the git CLI that matches the pattern git-<command> and is executable can be used to extend the Git command set.

For example:

git lfs

If the 'lfs' command is not implemented within Git's internal commands then the Git command line parser would check to see if 'git-lfs' exists and is executable.  If so, then the Git CLI passes execution to that script.

refactor(shorebird_cli): introduce ShorebirdExtension concept for encapsulation and testability

refactor(shorebird_cli): simplified closer to original implementation

chore(shorebird_cli): reformat

chore(shorebird_cli): fix warnings

## Description

The scope of the change is fairly narrow.  It adds a preprocessing step to the CommandRunner to detect unknown commands that are executable as 'shorebird-abcdefg', where abcdefg is the command unknown to the CommandRunner.

This change also includes the introduction of a new 'ExtensionCommand' which adds the unknown command as a dynamically named command into the CommandRunner during the pre-processing step.  The benefit of this approach is that extension commands are 'first class citizens', and as such mechanisms for first-class usage/help and even command completion are possible options to consider in the future.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


eseidel:
I'm familiar with the idea, but I'm not sure what problem this solves here.  (I'm happy to be convinced!)  Part of why this works so well with Git is that git is broken down into lots of little commands that can call each other.  Shorebird isn't.  So I'm not sure what value being able to type `shorbird blah` and have it run `shorebird-blah` will have?

zaitrarrio:
I think shorebird has the potential to be a platform that can pull lots of various external components together in various ways (as it already does, though most of these composable pieces are bridged internally).

This submission simply takes the model you currently have of wrapping components like Java and Git and Adb and abstracts it into a generic black box.  Where it becomes interesting is when shorebird context is intentionally  exposed to extensions(much like ShorebirdEnv exposes environment context to commands). As an example the location of assets or build artifacts can be exposed as an environment variable accessible to an extension command.

The specific use case I am solving for is doing a Shorebird Fastlane integration.

I know I could do this as an internal command to Shorebird, I just don't think it should live in the main codebase.  I could be convinced otherwise. 🙂

Another example use case is inserting BugSnag into the Shorebird workflow such that debug symbols can be produced as build artifacts and uploaded by:

shorebird bugsnag upload

I can wait and see if my Fastlane integration becomes unwieldy but thought I would share the idea/code for input and feedback.

Keep up the awesome work on Shorebird! Always fun seeing what great minds and really good people
can do when they come together to work towards a shared vision!

👍🏾🤩👍🏾

PR #1108, opened 2023-08-17
feat(shorebird_cli): `shorebird patch android` detects release version

felangel:
## Description

- feat(shorebird_cli): `shorebird patch android` detects release version (part of #1098)

## Type of Change

- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

Issue #860, opened 2023-07-17
feat: provide a way to send patches to multiple versions


eseidel:
Customers would like the ability to send a patch to multiple versions of their application.

It's something we could do, but would require customers allowing us source-access to their application, and building some minimal CI/CD.

The idea is similar to how we maintain patches for Flutter across multiple versions.  If the user has 3.0.4, 3.0.5, 3.06 and wants to patch 3.0.6 with a change, we could also auto-generate the diff for that patch and see if it applied to 3.0.5, etc.

We're aways away from making this happen, but it should be technically possible with our system.

eseidel:
We've not gotten enough requests for this that this seems worth keeping open for now.

PR #1190, opened 2023-08-31
feat: patch --release-version option

borjandev:
## Description

Added an (optional) release-version option to the `shorebird patch` commands for both `ios` and `android`

Android example:

`shorebird patch android --release-version 1.0.0+7`

iOS example: 

`shorebird patch ios --release-version 1.0.0+7`

This allows more fine-grained control over the version which is being patched, instead of always reading the latest value from the `pubspec.yaml` file.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


bryanoltman (commented):
Nice! I'll want to chat with the rest of the team to make sure this is the way we want to proceed, but this looks generally fine. Add some tests and we should be good to go!

borjandev:
@bryanoltman Sounds great!

Since a `final` variable may only be assigned once, and since in this PR it's being assigned at the same place where it would normally be extracted, in theory it should work seamlessly.

Added tests, also restored the original line break formatting for an easier and more concise review.

felangel:
I’m not sure we want to do this because it opens the door to accidentally patching a release version that differs from the local version. I’d recommend tagging your releases and checking out the tag before patching. Generally, things like assets and native code have changed across releases so it seems risky to allow patching across release versions like this.

borjandev:
> I’m not sure we want to do this because it opens the door to accidentally patching a release version that differs from the local version. I’d recommend tagging your releases and checking out the tag before patching. Generally, things like assets and native code have changed across releases so it seems risky to allow patching across release versions like this.

This PR in particular doesn't "open a door" which is not already there. 

I have been manually changing the `pubspec.yaml` with the specific version number that I want to patch, executing the patch, and then changing it back again, which is painstaking. 

There are also scenarios where the iOS app is going to be on a different release version than the Android app, for various reasons, ex: certain build getting accepted by Google Play while rejected by the App Store.

Another example would be fixing dart based issues which only affect one platform, ex: fixing the cursor height on iOS TextFields.. In this case would it be expected for someone to release a new Android version as well which doesn't actually include any code which affects the Android experience and go through the whole app store release / review process just so that the version number is the same across the platforms? How about when MacOS / Linux / Windows start getting supported? 

Since shorebird performs checks for assets which are not present in the new release, I believe there is a warning shown, as well as a manual "yes" confirmation needed from the developer before the patch is published, which should be sufficient to identify potential conflicts?

If you still feel it's too dangerous as-is, maybe as a common ground, I can make the `--release-version` option available only if used in conjunction with the `--force` option?

eseidel:
Seems also related to https://github.com/shorebirdtech/shorebird/issues/860.

SebastianEngel:
So basically this PR reverts the changes of #1108 and brings back the `release-version` parameter that existed until 0.12.3 and which was [removed in 0.13.0](https://github.com/shorebirdtech/shorebird/releases/tag/v0.13.0). Is that correct?

bryanoltman:
As mentioned in [Discord](https://discord.com/channels/1030243211995791380/1030243213002412164/1147489861004185662), this is currently overridable with the `--build-name` and `--build-number` flutter args (e.g., `shorebird patch android -- --build-name=v1.0.0 --build-number=1002`). This will probably influence the direction we go in.

borjandev:
@bryanoltman That works, however.. 

It doesn't show up as an option when running `shorebird patch android --help`

The `--` in the middle of the command is required and it feels off.

Since @felangel was mentioning that this PR opens a door to accidental patching, but this `-- --build-name=v1.0.0 --build-number=1002` already existed in the official release prior to this PR, does that mean that this command is so hidden that even @felangel was not aware of it's existence in the official release? 

Is it built like that on purpose, in order for it to be a secret command or similar?

eseidel:
(I think the real reason is that various bits of the team have been out this week and we haven't had a chance to sit down and discuss how we're going to address this user need.) I suspect we can get back to you late this week when everyone is back "in the office".  Thanks!

bryanoltman:
This has been implemented in https://github.com/shorebirdtech/shorebird/pull/1400

PR #1197, opened 2023-08-31
fix(shorebird_cli): support device-id parameter for shorebird preview on android

### LINKED ISSUE #1186: fix: shorebird CLI do not work "--device-id" option of preview command.
**Description**

it will not work passed `--device-id` option with preview command.

**Steps To Reproduce**

1. Execute `shorebird preview --device-id <emulator device id>`
2. exception with message. "Exception: Failed to install apks: [BT:1.14.1] Error: More than one device connected, please provide --device-id."

**Expected Behavior**

I'm running two emulator (or real device) and i tried preview command with `--device-id` option target one emulator.
I expected run app but i caught exception message and failed run app.

**Additional Context**

```bash
$ adb devices
List of devices attached
emulator-5554   device
emulator-5556   device

$ shorebird preview --device-id emulator-5554
✓ Fetching apps (0.5s)
Which app would you like to preview? flutter_mask (dev)
✓ Fetching releases (0.3s)
Which release would you like to preview? 1.0.1+2
Which platform would you like to preview? android
✓ Extracting metadata (0.5s)
⠴ Installing apks... (0.4s)
✗ Exception: Failed to install apks: [BT:1.14.1] Error: More than one device connected, please provide --device-id.
com.android.tools.build.bundletool.model.exceptions.CommandExecutionException: More than one device connected, please provide --device-id.
        at com.android.tools.build.bundletool.model.exceptions.InternalExceptionBuilder.build(InternalExceptionBuilder.java:57)
        at com.android.tools.build.bundletool.device.DeviceAnalyzer.getTargetDevice(DeviceAnalyzer.java:170)
        at com.android.tools.build.bundletool.device.DeviceAnalyzer.getAndValidateDevice(DeviceAnalyzer.java:133)
        at com.android.tools.build.bundletool.device.DeviceAnalyzer.getDeviceSpec(DeviceAnalyzer.java:55)
        at com.android.tools.build.bundletool.commands.InstallApksCommand.execute(InstallApksCommand.java:193)
        at com.android.tools.build.bundletool.BundleToolMain.main(BundleToolMain.java:117)
        at com.android.tools.build.bundletool.BundleToolMain.main(BundleToolMain.java:55)
 (1.5s)
```

bryanoltman:
## Description

Forwards the `--device-id` parameter to adb and bundletool for android preview commands. Reproduced the issue using two open emulators and verified the fix with the same. 

Fixes https://github.com/shorebirdtech/shorebird/issues/1186

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/adb.dart
```diff
@@ -28,15 +28,32 @@ class Adb {
...
 
   /// Starts the app with the given [package] name.
-  Future<void> startApp(String package) async {
-    final result = await _exec('shell monkey -p $package 1');
+  Future<void> startApp({
+    required String package,
+    String? deviceId,
+  }) async {
+    final args = [
+      if (deviceId != null) '-s $deviceId',
```
eseidel: This (combining two args into one string) will only work if shell=true I think.  Otherwise the args will get passed as separate strings and fail to parse I think?

eseidel: But otherwise this is a fantastic use of collection control flow statements. :).  I suspect the runInShell:false safe version is if (deviceId != null) ...['-s', deviceId]?  But then again, maybe I'm misremembering.

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -211,7 +226,11 @@ class PreviewCommand extends ShorebirdCommand {
     return process.exitCode;
   }
 
-  Future<int> installAndLaunchIos(String appId, Release release) async {
+  Future<int> installAndLaunchIos({
+    required String appId,
+    required Release release,
+    String? deviceId,
```
felangel, line 232: are we intentionally leaving this as unused?

felangel (approved):
LGTM ✅

eseidel:
We must always be using runInShell=true.  Since that does exec(shell, "-c", "$command ${args.join(' ')").  When runInShell=false, I believe it calls exec directly with: exec(command, args):
https://man7.org/linux/man-pages/man3/exec.3.html

packages/shorebird_cli/lib/src/adb.dart
felangel: I don't think this will matter in this particular case because we join all the args at the end before calling `exec`.

bryanoltman: Done

PR #1233, opened 2023-09-07
fix(shorebird_cli): switch to release flutter revision in patch

### LINKED ISSUE #1216: feat: Automatically select correct flutter revision when creating a patch
**Description**

When creating a Shorebird release, and supplying a flutter version (ex: 3.13.1), if the latest Flutter revision changes on Shorebird's side for the same flutter version that I provided then the patch will fail due to a revision mismatch.

As a user, I don't want to concern myself with tracking what specific revision was used to build my app, (especially in a CI environment) as I just want to provide the flutter version that I build and test my app against, and trust that Shorebird will use the correct revision.

Given that the developer triggers a Shorebird Release using Flutter version 3.13.1, and the revision used is `123`, if Shorebird updates the revision for Flutter 3.13.1 to `456`, then when the developer runs Shorebird Patch using Flutter version 3.13.1, then Shorebird should correctly select revision `123`, instead of failing on a mismatch against `456`.

If instead the developer changes their Flutter version for Shorebird Patch to 3.13.2 for example, then Shorebird should fail on a revision mismatch, as the developer made an intentional change to their flutter version.

bryanoltman:
## Description

When running either `shorebird patch android` or `shorebird patch ios-alpha`, switch to the release's Flutter revision if it is not the one Shorebird is currently using, and switch back after we're done.

Fixes https://github.com/shorebirdtech/shorebird/issues/1216

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -144,26 +146,33 @@ Please re-run the release command for this version or create a new release.''');
...
+      final buildProgress = logger.progress('Building patch');
+      try {
+        await buildAppBundle(flavor: flavor, target: target);
+        buildProgress.complete();
+      } on ProcessException catch (error) {
+        buildProgress.fail('Failed to build: ${error.message}');
+        return ExitCode.software.code;
+      }
+
+      flutterVersionProgress = logger.progress(
```
felangel: If the build fails we'll fail to revert the flutter revision which seems wrong.

bryanoltman: Good catch, will fix

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -144,26 +146,33 @@ Please re-run the release command for this version or create a new release.''');
...
+        'Switching back to original Flutter revision $originalFlutterRevision',
```
felangel:
```suggestion
        'Reverting to Flutter revision $originalFlutterRevision',
```
nit feel free to ignore

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -144,26 +146,33 @@ Please re-run the release command for this version or create a new release.''');
...
-  ${lightCyan.wrap('shorebird flutter versions use ${release.flutterRevision}')}''',
+Current Flutter Revision: $originalFlutterRevision''');
+
+      var flutterVersionProgress = logger.progress(
+        'Switching to Flutter revision ${release.flutterRevision}',
+      );
+      await shorebirdFlutter.useRevision(revision: release.flutterRevision);
+      flutterVersionProgress.complete();
+
+      final buildProgress = logger.progress('Building patch');
```
felangel, line 162: won't the two `logger.progress` calls interfere with each other?

felangel (approved):

PR #1266, opened 2023-09-13
feat(redis_client): basic `RedisClient`

felangel:
## Description

- feat(redis_client): basic `RedisClient` API

```dart
import 'dart:async';

import 'package:redis_client/redis_client.dart';

Future<void> main() async {
  // Create an instance of a RedisClient.
  final client = RedisClient();

  // Connect to the Redis server.
  await client.connect();

  // Set the value of a key.
  await client.set(key: 'HELLO', value: 'WORLD');

  // Get the value of a key.
  final value = await client.get(key: 'HELLO'); // WORLD

  // Delete the key.
  await client.delete(key: 'HELLO');

  // Close the connection to the Redis server.
  await client.close();
}
```

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/redis_client/README.md
+++ b/packages/redis_client/README.md
@@ -16,36 +16,19 @@ import 'package:redis_client/redis_client.dart';
 
 Future<void> main() async {
   // Create an instance of a RedisClient.
-  final client = RedisClient(
-    // Specify the socket options for the Redis server.
-    socket: const RedisSocketOptions(host: '<HOST>'),
-  );
+  final client = RedisClient();
 
   // Connect to the Redis server.
   await client.connect();
 
-  // Authenticate with the Redis server.
-  final authenticated = await client.auth(
-    username: '<USERNAME>',
-    password: '<PASSWORD>',
-  );
-
-  if (!authenticated) {
-    print('Failed to authenticate. Exiting...');
-    await client.close();
-    return;
-  }
-
-  const key = 'HELLO';
-
   // Set the value of a key.
-  await client.set(key: key, value: 'WORLD');
+  await client.set(key: 'HELLO', value: 'WORLD');
 
   // Get the value of a key.
-  final value = await client.get(key: key); // WORLD
+  final value = await client.get(key: 'HELLO'); // WORLD
 
   // Delete the key.
-  await client.delete(key: key);
+  await client.delete(key: 'HELLO');
 
   // Close the connection to the Redis server.
   await client.close();
--- a/packages/redis_client/README.md
+++ b/packages/redis_client/README.md
@@ -7,6 +7,51 @@ A Dart library for interacting with [Redis](https://redis.io).
 [license_badge]: https://img.shields.io/badge/license-MIT-blue.svg
 [license_link]: https://opensource.org/licenses/MIT
 
+## Quick Start
+
+```dart
+import 'dart:async';
+
+import 'package:redis_client/redis_client.dart';
+
+Future<void> main() async {
+  // Create an instance of a RedisClient.
+  final client = RedisClient(
+    // Specify the socket options for the Redis server.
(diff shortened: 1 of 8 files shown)


packages/redis_client/README.md
```diff
@@ -7,6 +7,34 @@ A Dart library for interacting with [Redis](https://redis.io).
...
+import 'dart:async';
+
+import 'package:redis_client/redis_client.dart';
+
+Future<void> main() async {
+  // Create an instance of a RedisClient.
+  final client = RedisClient();
+
+  // Connect to the Redis server.
+  await client.connect();
```
bryanoltman, line 22: At what point is the server URL/other connection info specified?

packages/redis_client/lib/src/redis_client.dart
```diff
@@ -1,7 +1,241 @@
...
+
+/// {@template redis_socket_options}
+/// Options for connecting to a Redis server.
+/// {@endtemplate}
+class RedisSocketOptions {
+  /// {@macro redis_socket_options}
+  const RedisSocketOptions({
+    this.host = 'localhost',
+    this.port = 6379,
+    this.timeout = const Duration(seconds: 30),
```
bryanoltman, line 16: Ah! Here it is. Is 6379 the default Redis port?

bryanoltman: Generally, I'd update the doc comments with notes about the defaults.

packages/redis_client/test/src/redis_client_test.dart
```diff
@@ -1,10 +1,157 @@
...
+        client.connect(),
+        throwsA(
+          isA<StateError>().having(
+            (e) => e.message,
+            'message',
+            'RedisClient has been closed.',
+          ),
+        ),
+      );
+    });
```
bryanoltman: I'd consider putting these in a "connection" group

packages/redis_client/test/src/redis_client_test.dart
```diff
@@ -1,10 +1,157 @@
...
+            ),
+          ),
+        );
+      });
+
+      test('fails when username is incorrect', () async {
+        await expectLater(
+          client.auth(username: 'shorebird', password: 'password'),
+          completion(equals(false)),
+        );
```
bryanoltman: Stylistic question: is there an advantage to this form of `expect` over `expect(await myFunction(), isTrue)`?

bryanoltman: Either way, `isTrue`/`isFalse` is probably more idiomatic use of the matcher framework

bryanoltman (approved):
A few small nits, otherwise LGTM!

packages/redis_client/README.md
felangel: It's passed as `RedisSocketOptions` via the constructor but there are defaults preconfigured.

packages/redis_client/lib/src/redis_client.dart
felangel: > Ah! Here it is. Is 6379 the default Redis port?

Yup these are all defaults. I'll update docs.

packages/redis_client/test/src/redis_client_test.dart
felangel: yup good call 👍

packages/redis_client/test/src/redis_client_test.dart
felangel: > Stylistic question: is there an advantage to this form of `expect` over `expect(await myFunction(), isTrue)`?

I believe the matcher description will be nicer if the test fails with the former as opposed to the later but it's mostly habitual 😆

Issue #981, opened 2023-08-01
fix: `shorebird preview` lists aar and ios-framework releases


bryanoltman:
**Description**

Releases created with `shorebird release aar` and `shorebird release ios-framework` appear in the `shorebird preview` as `android` and `ios` targets, respectively. This is wrong and confusing, as there isn't anything to meaningfully preview without an apk. 

Preview should either download the `.aar`/`.xcframework` and give the user instructions about how to include it in their app, or simply not list these as available platforms.

eseidel:
Just got a report from someone who ran into this:

```flutter_module_test % shorebird preview
✓ Fetching apps (0.4s)
Which app would you like to preview?
Which app would you like to preview? flutter_module_test
✓ Fetching releases (0.3s)
Which release would you like to preview?
Which release would you like to preview? 8.18.1+249
Which platform would you like to preview?
Which platform would you like to preview? android
✗ No artifact found for architecture aab in release 6605 (0.3s)
this is case of addtoapp android```

PR #1274, opened 2023-09-15
feat(shorebird_code_push_protocol): add canSideload field to CreateReleaseArtifactRequest and add support to CLI

bryanoltman:
## Description

Adds the `canSideload` to CreateReleaseArtifactRequest to indicate whether the artifact can be downloaded and run using `shorebird preview`.

Part of https://github.com/shorebirdtech/shorebird/issues/981

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_code_push_protocol/lib/src/messages/create_release_artifact/create_release_artifact_request.dart
```diff
@@ -32,11 +36,19 @@ class CreateReleaseArtifactRequest {
   /// The hash of the artifact.
   final String hash;
 
+  /// Whether the artifact can installed and run on a device/emulator as-is.
+  @JsonKey(fromJson: _parseStringToBool, toJson: _parseBoolToString)
```
felangel, line 40: why is this needed? json_serializable can handle bools afaik why do we need to send it down as a string?

bryanoltman: I left a note in the doc comment for the class – this type is used to populate the `fields` of a http.MultipartRequest, which is `Map<String, String>`, so all fields serialize to a string. lmk if that could be clearer.

bryanoltman: The other part of this change was to `createReleaseArtifact` in `CodePushClient`

felangel (approved):

PR #1276, opened 2023-09-18
feat(shorebird_cli): only list previewable releases for `shorebird preview`

### LINKED ISSUE #981: fix: `shorebird preview` lists aar and ios-framework releases

bryanoltman:
## Description

Updates the `shorebird preview` release version and platforms list to only include those which can be previewed (releases with Android app bundle artifacts or codesigned iOS Mach-O binaries).

Fixes https://github.com/shorebirdtech/shorebird/issues/981

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/test/src/commands/preview_command_test.dart
+++ b/packages/shorebird_cli/test/src/commands/preview_command_test.dart
@@ -125,7 +125,10 @@ void main() {
         () => codePushClientWrapper.getApps(),
       ).thenAnswer((_) async => [app]);
       when(
-        () => codePushClientWrapper.getReleases(appId: appId),
+        () => codePushClientWrapper.getReleases(
+          appId: appId,
+          sideloadableOnly: true,
+        ),
       ).thenAnswer((_) async => [release]);
       when(
         () => codePushClientWrapper.getReleaseArtifact(
@@ -172,14 +175,20 @@ void main() {
     test('exits with code 70 when querying for releases fails', () async {
       final exception = Exception('oops');
       when(
-        () => codePushClientWrapper.getReleases(appId: any(named: 'appId')),
+        () => codePushClientWrapper.getReleases(
+          appId: any(named: 'appId'),
+          sideloadableOnly: any(named: 'sideloadableOnly'),
+        ),
       ).thenThrow(exception);
       await expectLater(
         () => runWithOverrides(command.run),
         throwsA(exception),
       );
       verify(
-        () => codePushClientWrapper.getReleases(appId: appId),
+        () => codePushClientWrapper.getReleases(
+          appId: appId,
+          sideloadableOnly: true,
+        ),
       ).called(1);
     });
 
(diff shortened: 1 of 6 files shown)


packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -445,9 +445,14 @@ class CodePushClient {
...
   /// List all release for the provided [appId].
-  Future<List<Release>> getReleases({required String appId}) async {
+  Future<List<Release>> getReleases({
+    required String appId,
+    bool sideloadableOnly = false,
+  }) async {
     final response = await _httpClient.get(
-      Uri.parse('$_v1/apps/$appId/releases'),
+      Uri.parse(
+        '$_v1/apps/$appId/releases${sideloadableOnly ? '?sideloadable=true' : ''}',
```
felangel: rather than manually interpolating the query param it would be nice to use `.replace(query: ...)` imo.

bryanoltman: Ah nice, I didn't know about `.replace` – I thought we had to provide it via either the `.http` or the `.https` constructor. Thanks, and fixed

packages/shorebird_cli/test/src/commands/preview_command_test.dart
```diff
@@ -125,7 +125,10 @@ void main() {
         () => codePushClientWrapper.getApps(),
       ).thenAnswer((_) async => [app]);
       when(
-        () => codePushClientWrapper.getReleases(appId: appId),
+        () => codePushClientWrapper.getReleases(
+          appId: appId,
+          sideloadableOnly: true,
```
felangel:
```suggestion
          sideloadableOnly: any(named: 'sideloadableOnly'),
```

nit but shouldn't this be an any matcher for consistency?

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -445,10 +445,22 @@ class CodePushClient {
...
+    required String appId,
+    bool sideloadableOnly = false,
+  }) async {
+    var uri = Uri.parse(
+      '$_v1/apps/$appId/releases',
     );
+    if (sideloadableOnly) {
+      uri = uri.replace(
+        queryParameters: {
+          if (sideloadableOnly) 'sideloadable': 'true',
```
felangel:
```suggestion
          'sideloadable': 'true',
```
Do we need another check?

bryanoltman: Good call, fixed.

packages/shorebird_cli/test/src/commands/preview_command_test.dart
bryanoltman: Done

felangel (approved):

PR #1282, opened 2023-09-18
feat(redis_client): support `JSONPath`

felangel:
## Description

- feat(redis_client): support [`JSONPath`](https://redis.io/docs/data-types/json/path/)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/redis_client/lib/src/redis_client.dart
+++ b/packages/redis_client/lib/src/redis_client.dart
@@ -380,22 +380,26 @@ class RedisJson {
   /// https://redis.io/commands/json.set
   Future<void> set({
     required String key,
-    required Map<String, dynamic> value,
+    required dynamic value,
+    String path = r'$',
   }) {
-    return _client.execute(['JSON.SET', key, r'$', json.encode(value)]);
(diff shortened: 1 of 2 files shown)


packages/redis_client/lib/src/redis_client.dart
```diff
@@ -380,22 +380,26 @@ class RedisJson {
   /// https://redis.io/commands/json.set
   Future<void> set({
     required String key,
-    required Map<String, dynamic> value,
+    required dynamic value,
+    String path = r'$',
```
bryanoltman, line 384: Does `path` always need to start with a `$`? Should we do some kind of validation here and below to verify that if so?

bryanoltman (approved):

packages/redis_client/lib/src/redis_client.dart
felangel: Not necessarily (there's legacy support for alternative syntaxes so I figured it's best to be flexible with this API).

bryanoltman: Makes sense!

PR #1313, opened 2023-09-22
feat(shorebird_cli): create diffs for iOS patches

bryanoltman:
## Description

When creating a patch for an iOS release, diff against the release artifact instead of using the entire new artifact as a patch, _except for_ releases created with an old Flutter revision (i.e., any revision that we have used thus far).

This results in a ~10x size reduction in iOS patch sizes based on the small sample of apps I've tested thus far.

This will require an engine built with https://github.com/shorebirdtech/updater/pull/88 to work.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_ios_framework_command.dart
```diff
@@ -22,12 +26,56 @@ import 'package:shorebird_cli/src/shorebird_flutter.dart';
 import 'package:shorebird_cli/src/shorebird_validator.dart';
 import 'package:shorebird_code_push_client/shorebird_code_push_client.dart';
 
+/// Releases created before we introduced diffing for iOS will need patches
+/// that are not diffs. This set contains the Flutter revisions of those
+/// releases.
+///
+/// Generated by running `select distinct flutter_revision from app_release_;`
+/// on September 22, 2023.
+const _nonIosDiffFlutterRevisions = {
```
eseidel, line 35: Is this the same as the above?  Did you want to share it?

bryanoltman:
Closing, as this approach is a non-starter due to codesigning

PR #1334, opened 2023-09-29
fix(shorebird_cli): don't check for ipa when generating an unsigned release

### LINKED ISSUE #1335: `shorebird release ios-alpha --no-codesign` fails to find ipa
This command does not generate an ipa and will consequently fail to find one. I believe this had only succeeded for me locally because I had previously built without the `--no-codesign` flag and had an ipa hanging around in my build folder.

bryanoltman:
## Description

We don't generate an ipa if codesigning is disabled, so only attempt to find an ipa path in that case.

Fixes https://github.com/shorebirdtech/shorebird/issues/1335

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/test/src/commands/release/release_ios_command_test.dart
+++ b/packages/shorebird_cli/test/src/commands/release/release_ios_command_test.dart
@@ -635,9 +635,7 @@ error: exportArchive: No signing certificate "iOS Distribution" found
       );
 
       expect(exitCode, equals(ExitCode.software.code));
-      verify(
-        () => logger.err(any(that: contains('No directory found'))),
-      ).called(1);
+      verify(() => logger.err('Unable to find .xcarchive directory')).called(1);
     });
 
     test('exits with code 70 if ipa build directory does not exist', () async {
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/test/src/commands/release/release_ios_command_test.dart
```diff
@@ -635,9 +635,7 @@ error: exportArchive: No signing certificate "iOS Distribution" found
       );
 
       expect(exitCode, equals(ExitCode.software.code));
-      verify(
-        () => logger.err(any(that: contains('No directory found'))),
```
felangel, line 639: Where was this No directory found previously coming from?

felangel: Ah nvm it was from the remaining error when an ipa file was not found.

felangel (approved):

packages/shorebird_cli/test/src/commands/release/release_ios_command_test.dart
bryanoltman: https://github.com/shorebirdtech/shorebird/blob/bo/unsigned-ipa/packages/shorebird_cli/lib/src/shorebird_artifact_mixin.dart#L127

PR #1360, opened 2023-10-04
feat(shorebird_cli): add xcodeVersion method to xcodebuild wrapper

bryanoltman:
## Description

Adds the ability to get the currently installed Xcode version using `xcodebuild -version`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/executables/xcodebuild.dart
+++ b/packages/shorebird_cli/lib/src/executables/xcodebuild.dart
@@ -3,6 +3,7 @@ import 'dart:io';
 
 import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 import 'package:scoped/scoped.dart';
 import 'package:shorebird_cli/src/process.dart';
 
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/executables/xcodebuild.dart
```diff
@@ -106,4 +107,41 @@ class XcodeBuild {
...
+
+    final lines = LineSplitter.split('${result.stdout}').map((e) => e.trim());
+    var versionString = lines.firstOrNull?.split(' ').lastOrNull;
+    if (versionString == null) {
+      throw FormatException(
+        'Could not parse Xcode version from output: "${result.stdout}".',
+      );
+    }
+
+    // [Version.parse] requires a patch number. If Xcode does not report a patch
```
felangel, line 138: I wonder if this will work for all XCode versions and/or whether it would be easier and less error-prone to try to parse the version as a double and see if it’s greater than 15 since clearly XCode doesn’t follow semver I’m not sure this would be the only issue we encounter (for example if there’s a beta version of XCode do we handle that correctly with this logic). 🤷‍♂️

felangel: I guess we should probably be consistent with the flutter_tool in this case

felangel (approved):

packages/shorebird_cli/lib/src/executables/xcodebuild.dart
bryanoltman: Good Q about the beta, let me download one and give that a shot

bryanoltman: Interestingly, Xcode betas don't report that they're beta in `xcodebuild -version`:

```
⑆ xcrun xcodebuild -version
Xcode 15.1
Build version 15C5028h
```

bryanoltman: To resolve this, I don't think parsing it as a double would work, as there is occasionally a patch version.

PR #1374, opened 2023-10-06
feat(shorebird_cli): add non-strict version parsing function

bryanoltman:
## Description

Adds a `Version` extension to allow non-strict parsing of version numbers (e.g., `2.0`, which is not a valid semver version).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/extensions/version.dart
+++ b/packages/shorebird_cli/lib/src/extensions/version.dart
@@ -1,0 +1,33 @@
+import 'package:pub_semver/pub_semver.dart';
+
+extension VersionParsing on Version {
+  /// Attempts to parse a [versionString] into a [Version] object. If [strict]
+  /// is `false`, and the [versionString] does not contain a patch number, a
+  /// patch number of 0 will be added
+  static Version? tryParse(String versionString, {bool strict = true}) {
+    try {
+      return Version.parse(versionString);
+    } on FormatException {
+      if (strict) {
+        return null;
+      }
+
+      var updatedVersionString = versionString;
+      // [Version.parse] requires a patch number. If we are not in strict mode,
+      // and the version string is of the form "12.0", add a patch number of 0
+      // and try again.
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/extensions/version.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:pub_semver/pub_semver.dart';
+
+extension VersionParsing on Version {
```
felangel: Minor but making this an extension on Version feels weird because we're introducing a static method so you don't get the benefit of the API looking like `Version.tryParse` and the implementation doesn't benefit from being an extension on version. It might be simpler and easier to work with as a top-level function `Version? tryParse(String value, {bool strict = true}) {...}`

Then you the usage could look like:

```dart
import 'package:shorebird/cli/src/version.dart' as version;

final maybeVersion = version.tryParse(someValue);
```

Not a huge deal just leaving my thoughts.

bryanoltman: I like that better than what I have, updating now

packages/shorebird_cli/test/src/extensions/version_test.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:pub_semver/pub_semver.dart';
+import 'package:shorebird_cli/src/extensions/version.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('NonStrictParsing', () {
```
felangel:
```suggestion
  group(VersionParsing, () {
```

since this group contains both strict and lax parsing shouldn't it be renamed?

felangel (approved):

packages/shorebird_cli/test/src/extensions/version_test.dart
bryanoltman: Yeah, this is a holdover from a previous extension name. Fixed

PR #1400, opened 2023-10-13
feat(shorebird_cli): add release-version flag to iOS and Android patch commands

### LINKED ISSUE #1396: feat: ability to specify build number and version name for patches
**Description**

Due to the fact that `shorebird release` allows you to provide `--build-number` and `--build-name` arguments when creating releases (they're passed to the underlying `flutter build` invocation), I expected to be able to do the same for `shorebird patch`. Because I can't, there's no way (AFAICT) for me to create a patch for a release that used `--build-number` and `--build-name` without manually changing my `pubspec.yaml` file.

**Rationale**

I never change my `pubspec.yaml` file and instead use a Github Action that determines (based on tags) the correct build name to use and a simple file within my repo to track build number (mainly to fulfill Android's `versionCode` always-increments requirement. Otherwise I'd use the aforementioned Github Action's `increment` output to satisfy build number).

**Requirements**

- [ ] Add flags for `--build-name` and `--build-number` to the `shorebird patch` command
- [ ] Use these flags to patch the corresponding version (if it exists/has been released)

**Additional Context**

Another potentially reasonable intermediate fix is to always patch the latest version of the release Shorebird last built/uploaded. This is actually what I expected to happen when running `shorebird patch` without any version arguments.

bryanoltman:
## Description

Adds the `release-version` flag to `shorebird patch ios-alpha` and `shorebird patch android`. If provided, the CLI will not attempt to read the release version from the patch artifact.

Fixes https://github.com/shorebirdtech/shorebird/issues/1396

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
+++ b/packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
@@ -270,7 +270,6 @@ flutter:
       when(() => argResults['staging']).thenReturn(false);
       when(() => argResults['dry-run']).thenReturn(false);
       when(() => argResults['force']).thenReturn(false);
-      when(() => argResults['release-version']).thenReturn(null);
       when(() => auth.isAuthenticated).thenReturn(true);
       when(() => auth.client).thenReturn(httpClient);
       when(() => logger.progress(any())).thenReturn(progress);
--- a/packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
+++ b/packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
@@ -549,10 +549,6 @@ Please re-run the release command for this version or create a new release.'''),
     });
 
     group('when release-version option is not provided', () {
-      setUp(() {
-        when(() => argResults['release-version']).thenReturn(null);
-      });
-
       test('extracts release version from app bundle', () async {
         final tempDir = setUpTempDir();
         setUpTempArtifacts(tempDir);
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
```diff
@@ -270,7 +270,7 @@ flutter:
       when(() => argResults['staging']).thenReturn(false);
       when(() => argResults['dry-run']).thenReturn(false);
       when(() => argResults['force']).thenReturn(false);
-      when(() => argResults['release-version']).thenReturn(release.version);
+      when(() => argResults['release-version']).thenReturn(null);
```
felangel: I think we can just delete this line since returning `null` is the default behavior.

packages/shorebird_cli/test/src/commands/patch/patch_ios_command_test.dart
```diff
@@ -254,7 +254,7 @@ flutter:
       when(() => argResults['arch']).thenReturn(arch);
       when(() => argResults['dry-run']).thenReturn(false);
       when(() => argResults['force']).thenReturn(false);
-      when(() => argResults['release-version']).thenReturn(release.version);
+      when(() => argResults['release-version']).thenReturn(null);
```
felangel: same here

felangel (approved):

packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
bryanoltman: Good call, done

packages/shorebird_cli/test/src/commands/patch/patch_ios_command_test.dart
bryanoltman: Good call, done.

Issue #1213, opened 2023-09-05
fix: ADB: not detected


ABausG:
**Description**

Shorebird is not able to find `adb`. It surfaced using `shorebird preview` but also `shorebird doctor -v` shows that it cannot find adb

This is the output from `shorebird doctor -v`
```
~ shorebird doctor -v
Shorebird 0.14.6 • git@github.com:shorebirdtech/shorebird.git
Flutter 3.13.2 • revision 9486b6431e6c808c4e131f134b5d88017b3c32ab
Engine • revision e7bdf38443dff65d64446d6ca4074a35a587c770

Android Toolchain
  • Android Studio: /Applications/Android Studio.app/Contents
  • Android SDK: /Users/antonborries/Library/Android/sdk
  • ADB: not detected
  • JAVA_HOME: /Users/antonborries/.jenv/versions/11

✓ Shorebird is up-to-date (0.8s)
✓ Flutter install is correct (0.7s)

No issues detected!
```

However adb is working in the terminal

```
~ adb --version
Android Debug Bridge version 1.0.41
Version 34.0.1-9680074
Installed as /usr/local/bin/adb
```

**Expected Behavior**

Shorebird should detect adb

**Additional Context**

Noticed during Workshop at Flutter 'n' Friends

ABausG:
Further investigation showed that actually the Android sdk location that was is output by `shorebird doctor -v` was actually not the correct one. Exporting the correct sdk location via `ANDROID_HOME` fixes the issues where Shorebird has the correct Android Location as well as finding adb

eseidel:
Oh?  I'm very curious to know more about the location `shorebird doctor -v` was showing you?

eseidel:
As far as I can tell, `shorebird doctor` is at least validating that the path it returns exists:
https://github.com/shorebirdtech/shorebird/blob/2da950246d0cb11fc3315c249e3519e97d91046f/packages/shorebird_cli/lib/src/android_sdk.dart#L63C57-L63C57

eseidel:
Sorry, look like you noted above.  That `shorebird doctor` showed:
```
Android Toolchain
  • Android Studio: /Applications/Android Studio.app/Contents
  • Android SDK: /Users/antonborries/Library/Android/sdk
  • ADB: not detected
  • JAVA_HOME: /Users/antonborries/.jenv/versions/11
```

Interesting it says "ADB: not detected" (Should probably make that red, etc if it does.)

[...]

Issue #1163, opened 2023-08-24
fix: `flutter` prefers Android Studio over JAVA_HOME for java


MohammedAbdulaziz:
When i use ```shorebird preview``` I get this error after ```shorebird release android```
 
this is the error 
```shorebird preview
Which app would you like to preview?
Which app would you like to preview? next
✓ Fetching releases (0.2s)
Which release would you like to preview?
❯ ◉  1.0.1+1

Which release would you like to preview? 1.0.1+1
Which platform would you like to preview? android
✓ Fetching aab artifact (0.2s)
✓ Downloading release (14.2s)
✗ Exception: Failed to extract package name from app bundle: Error while parsing the flags: Syntax error: flags should start with -- (Abdelaziz\.shorebird\bin\cache\previews\7475aa43-6d88-4cdb-8f0e-08f8d117e2eb\android_1.0.1+1.aab)
 (0.2s)
```

this is my ```shorebird doctor```

```shorebird doctor
Shorebird 0.14.3 • git@github.com:shorebirdtech/shorebird.git
Flutter 3.13.0 • revision d84d466eacbeb47d6e81e960c22c6fdfe5a3917d
Engine • revision a4613505045177f96d7cbfbb42335e2db7cf7b51

✓ Shorebird is up-to-date (1.4s)
✓ Flutter install is correct (1.4s)
✓ AndroidManifest.xml files contain INTERNET permission (99ms)

No issues detected!
```

bryanoltman:
I'm not able to reproduce this. It would be helpful to know:

1. What OS you're running on
2. What the full output of `shorebird preview -v` is

MohammedAbdulaziz:
> I'm not able to reproduce this. It would be helpful to know:
> 
> 1. What OS you're running on

This is my OS specifications
Edition	Windows 10 Home
Version	21H2
Installed on	‎8/‎14/‎2021
OS build	19044.1826
Experience	Windows Feature Experience Pack 120.2212.4180.0

> 2. What the full output of `shorebird preview -v` is

This is the full output of `shorebird preview -v`
```shorebird preview -v
[HTTP] POST https://oauth2.googleapis.com/token
⠋ Fetching apps... (1.6s)[HTTP] GET https://api.shorebird.dev/api/v1/apps
✓ Fetching apps (2.3s)
Which app would you like to preview? next
[...]

PR #1403, opened 2023-10-16
feat(shorebird_cli): Add OS wrapper for which/where.exe

bryanoltman:
## Description

Adds the `OperatingSystemInterface` class, which wraps OS-specific functionality (currently just `which`/`where.exe` for locating executables on a user's PATH).

Part of a solution for https://github.com/shorebirdtech/shorebird/issues/1163 and https://github.com/shorebirdtech/shorebird/issues/1213, which involves using the same logic as `flutter` to find these tools.

See https://github.com/flutter/flutter/blob/stable/packages/flutter_tools/lib/src/android/java.dart#L45-L54 for Java.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/os/operating_system_interface.dart
+++ b/packages/shorebird_cli/lib/src/os/operating_system_interface.dart
@@ -2,25 +2,24 @@ import 'dart:io';
 
 import 'package:mason_logger/mason_logger.dart';
 import 'package:scoped/scoped.dart';
+import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/process.dart';
 
 /// A reference to a [OperatingSystemInterface] instance.
-final osRef = create(() => OperatingSystemInterface.instance);
+final osRef = create(OperatingSystemInterface.new);
 
 /// The [OperatingSystemInterface] instance available in the current zone.
 OperatingSystemInterface get os => read(osRef);
 
+/// {@template operating_system_interface}
+/// A wrapper around operating system specific functionality.
+/// {@endtemplate}
 abstract class OperatingSystemInterface {
-  /// Returns the first instance of [executableName] found on the PATH.
-  ///
-  /// This is the equivalent of the `which` command on Linux and macOS and
-  /// `where.exe` on Windows.
-  String? which(String executableName);
-
-  static OperatingSystemInterface get instance {
-    if (Platform.isWindows) {
+  /// {@macro operating_system_interface}
+  factory OperatingSystemInterface() {
+    if (platform.isWindows) {
       return _WindowsOperatingSystemInterface();
-    } else if (Platform.isMacOS || Platform.isLinux) {
+    } else if (platform.isMacOS || platform.isLinux) {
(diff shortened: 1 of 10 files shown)


packages/shorebird_cli/lib/src/os/operating_system_interface.dart
```diff
@@ -0,0 +1,65 @@
...
+final osRef = create(OperatingSystemInterface.new);
+
+/// The [OperatingSystemInterface] instance available in the current zone.
+OperatingSystemInterface get os => read(osRef);
+// coverage:ignore-end
+
+/// {@template operating_system_interface}
+/// A wrapper around operating system specific functionality.
+/// {@endtemplate}
+abstract class OperatingSystemInterface {
```
felangel, line 20: Can we rename this to `OperatingSystem`?

bryanoltman: I'm not crazy about that, as I would expect `OperatingSystem` to represent…well, an operating system, and have properties like "name" (e.g., "Linux"), "version", etc.

bryanoltman: `flutter_tools` calls their version of this `OperatingSystemUtils`, which seems about the same level of fine to me as `OperatingSystemInterface` (although I think I might like "interface" better, because we're "interfacing" with the OS, and util is not especially descriptive in any context)

felangel: Ah I see when I read `OperatingSystemInterface` I thought of interface as in a class that must be implemented. I think it's a little confusing to the ref `os` then we should be consistent imo 🤷

packages/shorebird_cli/lib/src/os/operating_system_interface.dart
```diff
@@ -0,0 +1,65 @@
...
+import 'package:shorebird_cli/src/platform.dart';
+import 'package:shorebird_cli/src/process.dart';
+
+// TODO(bryanoltman): remove this once os is used.
+// coverage:ignore-start
+/// A reference to a [OperatingSystemInterface] instance.
+final osRef = create(OperatingSystemInterface.new);
+
+/// The [OperatingSystemInterface] instance available in the current zone.
+OperatingSystemInterface get os => read(osRef);
```
felangel: seems weird to call this `os` when the class isn't called `OperatingSystem` imo

felangel (approved):

packages/shorebird_cli/lib/src/os/operating_system_interface.dart
bryanoltman: Fixed

packages/shorebird_cli/lib/src/os/operating_system_interface.dart
bryanoltman: Updated the name of `os` to `osInterface`

PR #1416, opened 2023-10-19
chore(code_push_protocol): remove `tosAcceptedAt` from `User`

felangel:
## Description

- Since this will be enforced strictly by the backend we don't need to expose this field on the public `User` object. If a user must accept new terms the backend will return a response to indicate so.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [X] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/test/src/os/operating_system_interface_test.dart
+++ b/packages/shorebird_cli/test/src/os/operating_system_interface_test.dart
@@ -119,7 +119,9 @@ void main() {
 
           test('returns null', () {
             expect(
-                runWithOverrides(() => osInterface.which('shorebird')), isNull);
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/test/src/os/operating_system_interface_test.dart
```diff
@@ -119,7 +119,9 @@ void main() {
 
           test('returns null', () {
             expect(
-                runWithOverrides(() => osInterface.which('shorebird')), isNull);
+              runWithOverrides(() => osInterface.which('shorebird')),
+              isNull,
```
felangel, line 123: fix missing trailing comma

eseidel (approved):

PR #1447, opened 2023-11-02
fix(shorebird_cli): don't run `flutter pub get` after build if flutter is not installed

### LINKED ISSUE #1430: "`flutter pub get` failed to run after the build completed" prints if Flutter is not installed
This is sometimes an issue on CI. We shouldn't attempt to run this command if Flutter is not installed.

bryanoltman:
## Description

Checks for the presence of flutter on the user's path before attempting to run `flutter pub get` after a build.

For context: `flutter pub get` is run after a build to reset `.dart_tool/package_config.json` to point to the system flutter install instead of the shorebird flutter install. This value is set when running a flutter build command or flutter pub get.

Fixes https://github.com/shorebirdtech/shorebird/issues/1430

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
@@ -5,6 +5,7 @@ import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/logger.dart';
+import 'package:shorebird_cli/src/os/operating_system_interface.dart';
 import 'package:shorebird_cli/src/process.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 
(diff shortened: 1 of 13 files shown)


packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
```diff
@@ -277,6 +278,12 @@ mixin ShorebirdBuildMixin on ShorebirdCommand {
   /// https://github.com/shorebirdtech/shorebird/issues/1101 for more info.
   Future<void> _systemFlutterPubGet() async {
     const executable = 'flutter';
+    if (osInterface.which(executable) == null) {
```
felangel, line 281: we really need to get rid of this mixin so we can avoid having to duplicate the tests 🤣

felangel (approved):

packages/shorebird_cli/lib/src/shorebird_build_mixin.dart
bryanoltman: Yeah, this PR is a case study re: why we shouldn't use mixins for any functionality we want to test :)

PR #1469, opened 2023-11-08
feat(shorebird_cli): add `ShorebirdLinkerArtifact` to cache

felangel:
## Description

- feat(shorebird_cli): add `ShorebirdLinkerArtifact` to cache
  - still need to modify build_engine to build and upload the shorebird_linker to our storage bucket

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -196,8 +196,9 @@ class ShorebirdLinkerArtifact extends CachedArtifact {
   @override
   List<String> get executables => ['shorebird_linker'];
 
+  /// The linker is only available for revisions that support mixed-mode.
   @override
-  bool get required => false; // Only supported for newer engine revisions.
+  bool get required => false;
 
   @override
   String get storageUrl {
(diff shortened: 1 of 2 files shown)


eseidel (approved):
It's kinda silly that we're buiding these (instead of shipping down the dart), but if it works, lgtm!

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -176,6 +187,33 @@ allowed to access $storageUrl.''',
...
+  ShorebirdLinkerArtifact({required super.cache, required super.platform});
+
+  @override
+  String get name => 'shorebird_linker';
+
+  @override
+  List<String> get executables => ['shorebird_linker'];
+
+  @override
+  bool get required => false; // Only supported for newer engine revisions.
```
bryanoltman: I can imagine future us asking: "newer than what?"

bryanoltman (approved):

packages/shorebird_cli/lib/src/cache.dart
felangel: Yeah good point I can reword to say this is only available in revisions that support mixed-mode

PR #1489, opened 2023-11-13
fix(shorebird_cli): `shorebird init` should not prompt when `--force` or running in CI

felangel:
## Description

- fix(shorebird_cli): `shorebird init` does not prompt when `--force` or running in CI

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -152,10 +152,14 @@ Please make sure you are running "shorebird init" from the root of your Flutter 
     final String appId;
     Map<String, String>? flavors;
     try {
-      final displayName = logger.prompt(
-        '${lightGreen.wrap('?')} How should we refer to this app?',
-        defaultValue: shorebirdEnv.getPubspecYaml()?.name,
-      );
+      final needsConfirmation = !force && !shorebirdEnv.isRunningOnCI;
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -152,10 +152,14 @@ Please make sure you are running "shorebird init" from the root of your Flutter
...
     Map<String, String>? flavors;
     try {
-      final displayName = logger.prompt(
-        '${lightGreen.wrap('?')} How should we refer to this app?',
-        defaultValue: shorebirdEnv.getPubspecYaml()?.name,
-      );
+      final needsConfirmation = !force && !shorebirdEnv.isRunningOnCI;
+      final pubspecName = shorebirdEnv.getPubspecYaml()!.name;
+      final displayName = needsConfirmation
+          ? logger.prompt(
```
eseidel, line 158: I suspect we eventually make a systemic fix to swap out the "prompt" callback to have a default that is either taken in the CI case, or if no default, throws.

felangel: Yup I just kept it consistent with how @bryanoltman was doing it in other places but I agree we can do a refactor to abstract the logger in a separate PR

PR #1498, opened 2023-11-15
fix(shorebird_cli): flutter validator should not warn if system Flutter does not exist

### LINKED ISSUE #337: fix: Shorebird flutter validation should not fail if a user does not have Flutter installed on their device
**Description**

As a user, I should be able to use Shorebird without having Dart/Flutter installed on my machine. Currently, shorebird validators check if the shorebird flutter is the same as the system flutter but if there is no system flutter the validation check should pass instead of throwing.

**Steps To Reproduce**

1. Remove `flutter` from system path
2. Run `shorebird build appbundle`
3. See error

**Expected Behavior**

No validation error should occur

felangel:
## Description

- fix(shorebird_cli): flutter validator does not warn if system Flutter does not exist (closes #337)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
+++ b/packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
@@ -15,6 +15,8 @@ class FlutterValidationException implements Exception {
   String toString() => 'FlutterValidationException: $message';
 }
 
+class CommandNotFoundException implements Exception {}
+
 class ShorebirdFlutterValidator extends Validator {
   ShorebirdFlutterValidator();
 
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/validators/shorebird_flutter_validator.dart
```diff
@@ -117,6 +121,8 @@ This can cause unexpected behavior if you are switching between the tools and th
           ? await shorebirdFlutter.getVersion()
           : await shorebirdFlutter.getSystemVersion();
     } on ProcessException catch (error) {
+      if (error.errorCode == 127) throw CommandNotFoundException();
```
bryanoltman, line 124: It's not on the [sysexits man page](https://man.freebsd.org/cgi/man.cgi?query=sysexits) linked to in the mason_logger source, but it looks like code 127 is used by `sh` and `bash` to indicate that a command was not found on the user's PATH (https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html). I was going to suggest just leaving a comment here about this, but adding this to mason_logger might not be a bad idea. Should I open a PR?

PR #1515, opened 2023-11-17
feat(shorebird_cli): validate storage.googleapis.com is pingable in doctor

### LINKED ISSUE #1506: Update "does shorebird work from my country" faq
The point is to explain to customers how to teach reachability of shorebird's servers.

Maybe instead we should include this as part of `shorebird doctor`?

Confusion: https://discord.com/channels/1030243211995791380/1174569116875628574/1174571036059115520

bryanoltman:
## Description

Adds a validator to verify that the user can ping `https://storage.googleapis.com`, which is required to upload release artifacts.

Fixes https://github.com/shorebirdtech/shorebird/issues/1506

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/storage_access_validator.dart
+++ b/packages/shorebird_cli/lib/src/validators/storage_access_validator.dart
@@ -1,0 +1,31 @@
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/platform.dart';
+import 'package:shorebird_cli/src/process.dart';
+import 'package:shorebird_cli/src/validators/validators.dart';
+
+class StorageAccessValidator extends Validator {
+  @override
+  String get description => 'Has access to storage.googleapis.com';
+
+  @override
+  Future<List<ValidationIssue>> validate() async {
+    final result = await process.run(
+      'ping',
+      [
+        // ping on Windows auto-terminates, but will go on indefinitely on
+        // other linux and mac unless we set a count. 2 was chosen arbitrarily.
+        if (!platform.isWindows) ...['-c', '2'],
+        'https://storage.googleapis.com',
+      ],
+    );
+    if (result.exitCode != ExitCode.success.code) {
+      return [
+        const ValidationIssue(
+          severity: ValidationIssueSeverity.error,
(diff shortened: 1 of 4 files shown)


eseidel (approved):
Felix is probably a better reviewer than I, but looks reasonable.  I'm not sure how long ping takes and thus how much longer this makes doctor, but seems like a good start!

bryanoltman:
Sure thing! With `-c 2`, this adds 1s to doctor. `ping`'s default interval is 1s, so there's the initial ping, which happens immediately, and then the second ping, which happens 1s later. The number 2 was chosen arbitrarily, and we could potentially get away with just one. Windows does 4 pings by default (https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/ping), which we could also change.

bryanoltman:
Despite the documentation, I'm unable to get a successful `ping` on our windows gha runner, so let's maybe hold off on merging this until I have a windows machine handy to test with.

bryanoltman:
Update: omitting the `https://` got this working, and also works on mac

eseidel:
I expect we'll end up pinging more things (and probably not with ping, but rather our own dart calls), but just getting a first thing working is and important first step!

bryanoltman:
Updated to only execute a single ping on all platforms. This check now takes about 60ms on my Windows laptop.

eseidel (approved):

packages/shorebird_cli/lib/src/validators/storage_access_validator.dart
```diff
@@ -0,0 +1,30 @@
...
+import 'package:shorebird_cli/src/process.dart';
+import 'package:shorebird_cli/src/validators/validators.dart';
+
+class StorageAccessValidator extends Validator {
+  @override
+  String get description => 'Has access to storage.googleapis.com';
+
+  @override
+  Future<List<ValidationIssue>> validate() async {
+    final result = await process.run(
```
felangel, line 12: Shouldn't we make an `http` request using `package:http` since that's what we'll be using when users try to upload/download files instead of `ping`?

felangel (approved):

PR #1516, opened 2023-11-20
feat(shorebird_cli): show upload progress for release and patch artifact uploads

bryanoltman:
## Description

Adds an `http.BaseClient` subclass to allow progress reporting for multipart uploads. 

WIP until tests are added.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_code_push_client/lib/src/request_progress.dart
```diff
@@ -0,0 +1,26 @@
+/// A function that reports download or upload progress.
+typedef ProgressCallback = void Function(DataTransferProgress progress);
```
felangel: Does this need to return a `DataTransferProgress` object or can it just be a `double`?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -412,6 +419,8 @@ ${archMetadata.arch} artifact already exists, continuing...''',
       }
     }
 
+    const baseLogMessage = 'Creating aab artifact';
```
felangel: Seems a bit misleading to call this `creating` because the upload will likely account for majority of the time spent.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -395,7 +396,13 @@ Please create a release using "shorebird release" and try again.
           platform: platform,
           hash: hash,
           canSideload: false,
+          onProgress: (progress) {
```
felangel: unrelated to this PR but would be nice to consolidate these two since the code looks almost identical.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -86,6 +189,7 @@ class CodePushClient {
     http.Client? httpClient,
     Uri? hostedUri,
   })  : _httpClient = _CodePushHttpClient(httpClient ?? http.Client()),
+        _uploadProgressClient = _UploadProgressHttpClient(),
```
felangel: Seems weird to have to inject a separate client for the uploads. If we go with this approach it might be nicer to just have a `ShorebirdHttpClient` which includes the retry logic, logging, and optional progress reporting.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -77,6 +79,107 @@ class _CodePushHttpClient extends http.BaseClient {
   }
 }
 
+/// Based on the [IOClient] from the `http` package. The primary difference is
+/// that this implementation uses `stream.addStream` instead of `stream.pipe`
+/// to allow for progress reporting.
+class _UploadProgressHttpClient extends http.BaseClient {
```
felangel: Rather than extending `BaseClient` couldn't we extend `MultipartRequest` and override `finalize` and expose an `onProgress` callback? It seems like it would be a lot simpler and less invasive. Then we wouldn't need to create a special http client instance we could just use our custom `MultipartRequest` with the existing `http.Client`.

packages/shorebird_code_push_client/lib/src/request_progress.dart
bryanoltman: It could be either. I thought it would be good to include the URL at least, as `progressStream` is per-client (not per-request), and this provides consumers a way to verify that the progress being reported is for the upload they care about (although this is not a problem we currently have), and once the URL was in there, it seemed like the other info was nice to have. 🤷

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: Yep, log messages need to be updated

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: Agreed

packages/shorebird_code_push_client/lib/src/code_push_client.dart
bryanoltman: Agreed that it's simpler, but that was the first thing I tried, and the stream progress very quickly completes (goes from 0% to 100% with a bunch of intermediate values) and then hangs while the actual upload happens. I believe the progress we see in the `MultipartRequest` subclass is the file being read, not the upload happening.

bryanoltman: For posterity, the way I'm testing this is:

1. `shorebird release android`
2. At the release confirmation prompt, turn on the network link condition. I'm using LTE speed.
3. Type 'y'
4. Observe

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: (will do before this is out of a draft state)

packages/shorebird_code_push_client/lib/src/code_push_client.dart
bryanoltman: Yeah, I agree. I just explored moving this into the `shorebird_cli` package, but was thwarted because 1) this wraps a `dart:io` HttpClient, not a `package:http` `http.Client`, meaning it has to be the innermost wrapped class (with LoggingClient and RetryingClient around it), which in turn means we don't have a ready way to access the APIs it offers that aren't on `http.BaseClient` and 2) `CodePushClient` needs to be aware of the types that are used in the progress callback. We could do away with the second problem if we just communicated progress through a primitive type, but I don't have a clean solution to the first issue offhand.

lhmzhou (approved):
LGTM! Will you be squashing your commits before merge?

bryanoltman:
@lhmzhou yep, we always do!

@felangel one thing I just realized is that, because we're using a different http client for uploads, we lose the logging we currently have when POSTing to the storage bucket URLs. Looking to see if there's a non-awful way to recover that now.

PR #1524, opened 2023-11-21
feat(shorebird_cli): support local engine artifacts

### LINKED ISSUE #1518: Teach `shorebird` about local engine assets
Right now `shorebird` always uses assets from its cache or the flutter cache, even when --local-engine is specified.

This is wrong and manifests in a few places, including when using `gen_snapshot` for generating the elf/vmcode files for patching on iOS if you're using --local-engine you'll end up with snapshot mismatch errors if you have an edited version of Dart in your local engine relative to what was the most recent shorebird release.

felangel:
## Description

- feat(shorebird_cli): support local engine artifacts
  - closes #1518 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/flutter_artifacts.dart
+++ b/packages/shorebird_cli/lib/src/flutter_artifacts.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: one_member_abstracts
+
 import 'dart:io';
 
 import 'package:path/path.dart' as p;
@@ -16,7 +18,7 @@ enum FlutterArtifact {
 
 /// A reference to a [FlutterArtifacts] instance.
 final flutterArtifactsRef = create<FlutterArtifacts>(
-  () => const FlutterCachedArtifacts(),
+  FlutterCachedArtifacts.new,
 );
 
 /// The [FlutterArtifacts] instance available in the current zone.
@@ -26,9 +28,6 @@ FlutterArtifacts get flutterArtifacts => read(flutterArtifactsRef);
 /// A class that provides access to Flutter artifacts.
 /// {@endtemplate}
 abstract class FlutterArtifacts {
-  /// {@macro flutter_artifacts}
-  const FlutterArtifacts();
-
   /// Returns the path to the given [artifact].
   String getArtifactPath({required FlutterArtifact artifact});
 }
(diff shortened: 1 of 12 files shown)


eseidel (approved):

packages/shorebird_cli/lib/src/flutter_artifacts.dart
```diff
@@ -0,0 +1,124 @@
...
+
+  File get _genSnapshotFile {
+    return File(
+      p.join(
+        shorebirdEnv.flutterDirectory.path,
+        'bin',
+        'cache',
+        'artifacts',
+        'engine',
+        'ios-release',
```
bryanoltman, line 62: This section of directories is repeated – might be nice to extract to a shared place. The executable names could also be made a property of the `FlutterArtifact` enum.

packages/shorebird_cli/lib/src/flutter_artifacts.dart
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }
+
+  File get _genSnapshotFile {
+    return File(
+      p.join(
+        engineConfig.localEngineSrcPath!,
+        'out',
+        'ios_release',
+        'clang_x64',
```
bryanoltman, line 106: Same here re: path to artifacts and artifact executable name

bryanoltman (approved):
LGTM, just a few small improvement suggestions

packages/shorebird_cli/lib/src/flutter_artifacts.dart
felangel: I didn't want to over-optimize here because I wanted to keep the local engine artifacts completely decoupled from the flutter cached artifacts since technically there's no reason they need to have the same path segments and/or executable names 🤷

felangel: Will merge as is for now but lmk if you feel strongly and I'm happy to refactor in a subsequent PR

bryanoltman: Nope, just an idea for cleanup. Not hurting anything as it is :)

PR #1576, opened 2023-12-17
feat: enable linker by default

eseidel:
This renames --use-linker to --mix and makes it on by default.

This isn't quite complete though as it needs to check which
version of Flutter the release you're patching used and 'mix' or
not depending on the version.  Old versions shouldn't mix, all new
versions should.

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -205,6 +205,14 @@ class AotToolsArtifact extends CachedArtifact {
   bool get required => false;
 
   @override
+  Directory get location => Directory(
+        p.join(
+          cache.getArtifactDirectory(name).path,
+          shorebirdEnv.shorebirdEngineRevision,
+        ),
+      );
+
+  @override
   String get storageUrl {
     var artifactName = 'aot-tools-';
     if (platform.isMacOS) {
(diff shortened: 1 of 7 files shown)


eseidel:
I think this mostly needs a set of Flutter versions which should default to --no-mix and then can land?

eseidel:
Hmm, not sure this is working, or it captured some of our un-released branches?

```
base_app_snapshot_instructions or base_app_snapshot_data is set but no link table is available, ignoring base snapshot.
```
When I ran with my local build just now.

bryanoltman:
> Hmm, not sure this is working, or it captured some of our un-released branches?

Capturing some of the discord conversation, the `patch` command uses the linker if the current engine revision is one we don't know about. Because we haven't released an engine revision with a working linker ("working" meaning "produces apps that don't crash"), none of the known engine revs will use the linker.

Updated to always use the linker if a local engine is being used.

eseidel:
Ooooh. That makes sense.

eseidel:
Hmm. But doesn’t the rev of flutter on main include aot_tools?  Even if not
working I would expect our latest rev to have a linker included.

bryanoltman:
It includes aot_tools, but because that linker produces patches that crash, it seemed wrong to ever use it

eseidel:
Makes sense. Was just making sure we understood the mechanism. Thanks!

felangel (approved):

PR #1596, opened 2023-12-21
fix(shorebird_cli): use aot-tools.dill for linking if available

### LINKED ISSUE #1595: fix: shorebird doesn't work on mac x64 machines
**Description**

Just got started with Shorebird today. `shorebird release iOS-alpha` worked fine, but `shorebird patch iOS-alpha` gave an error: Failed to link AOT files: ProcessException: Bad CPU type in executable.

See screenshots for the entire response and result from shorebird doctor.

**Screenshots**

repsonse
doctor

bryanoltman:
## Description

Because the precompiled `aot-tools` executable will only work on architectures that match the `dart` executable used to build it, and because we build dart on arm64 macs, the `aot-tools` executable does not work on x64 macs. This change looks for an aot-tools kernel file (`.dill`) and uses the `dart` packaged with shorebird's flutter (which was used to produce the kernel file) to run it.

Fixes https://github.com/shorebirdtech/shorebird/issues/1595

Related to but not dependent on https://github.com/shorebirdtech/_build_engine/pull/56

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -137,12 +137,14 @@ abstract class CachedArtifact {
 
   String get storageUrl;
 
-  String get executable;
+  String get fileName;
+
+  bool get isExecutable;
 
   bool get required => true;
 
   Future<void> extractArtifact(http.ByteStream stream, String outputPath) {
-    final file = File(p.join(outputPath, executable))
+    final file = File(p.join(outputPath, fileName))
       ..createSync(recursive: true);
     return stream.pipe(file.openWrite());
   }
@@ -180,10 +182,10 @@ allowed to access $storageUrl.''',
 
     await extractArtifact(response.stream, location.path);
 
-    if (!platform.isWindows) {
+    if (!platform.isWindows && isExecutable) {
       final result = await process.start(
         'chmod',
-        ['+x', p.join(location.path, executable)],
+        ['+x', p.join(location.path, fileName)],
       );
       await result.exitCode;
     }
@@ -199,7 +201,10 @@ class AotToolsArtifact extends CachedArtifact {
   // Not technically an executable, but this is where the file should end up and
   // is how it will be consumed.
   @override
-  String get executable => 'aot-tools.dill';
+  String get fileName => 'aot-tools.dill';
+
+  @override
+  bool get isExecutable => false;
 
   /// The aot-tools are only available for revisions that support mixed-mode.
   @override
@@ -225,7 +230,10 @@ class PatchArtifact extends CachedArtifact {
   String get name => 'patch';
 
   @override
-  String get executable => 'patch';
+  String get fileName => 'patch';
+
+  @override
+  bool get isExecutable => true;
 
   @override
   Future<void> extractArtifact(
@@ -260,7 +268,10 @@ class BundleToolArtifact extends CachedArtifact {
   String get name => 'bundletool.jar';
 
   @override
-  String get executable => 'bundletool.jar';
+  String get fileName => 'bundletool.jar';
+
+  @override
+  bool get isExecutable => false;
 
   @override
   String get storageUrl {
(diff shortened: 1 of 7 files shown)


packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -179,9 +180,7 @@ allowed to access $storageUrl.''',
 
     await extractArtifact(response.stream, location.path);
 
-    if (platform.isWindows) return;
-
-    for (final executable in executables) {
+    if (!platform.isWindows) {
       final result = await process.start(
         'chmod',
         ['+x', p.join(location.path, executable)],
```
eseidel: lol.  So I guess we're making the .dill executable? :)

eseidel (approved):

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: ¯\_(ツ)_/¯ seemed fine, but I added an `isExecutable` property to `CachedArtifact` so only those are now made to be executable.

eseidel: Yeah was fine, mostly amused.

eseidel (approved):

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -198,7 +199,10 @@ class AotToolsArtifact extends CachedArtifact {
   String get name => 'aot-tools';
```
felangel, line 199: Should we rename this to executableName to avoid confusion? I don't think it's obvious what the difference between name and fileName is at first glance.

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -198,7 +199,10 @@ class AotToolsArtifact extends CachedArtifact {
...
 
   @override
-  List<String> get executables => ['aot-tools'];
+  String get fileName => 'aot-tools.dill';
```
felangel, line 202: Why couldn't we change `name` to be `aot-tools.dill` and avoid introducing a new `fileName`?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -213,18 +217,8 @@ class AotToolsArtifact extends CachedArtifact {
...
-    } else if (platform.isLinux) {
-      artifactName += 'linux-x64';
-    } else if (platform.isWindows) {
-      artifactName += 'windows-x64';
-    }
-
-    return '${cache.storageBaseUrl}/${cache.storageBucket}/shorebird/${shorebirdEnv.shorebirdEngineRevision}/$artifactName';
-  }
+  String get storageUrl =>
+      '${cache.storageBaseUrl}/${cache.storageBucket}/shorebird/${shorebirdEnv.shorebirdEngineRevision}/aot-tools.dill';
```
felangel, line 221: Could this use `fileName` instead of the string literal `aot-tools.dill`?

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Because the artifact path is `[name]/[fileName]` (or, in the case of aot-tools, `[name]/[flutterRev]/[fileName]`). We could use fileName for both, but this felt cleaner.

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: We could ¯\_(ツ)_/¯

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: I can add some comments. I changed `executables` to `executable` (because our artifacts only ever have one file) to `fileName` (because our artifacts aren't always executable).

PR #1612, opened 2024-01-03
fix(shorebird_cli): don't copy directory before zipping

bryanoltman:
## Description

I'm not sure why we were copying this in the first place, but this fixes the issue described in https://github.com/shorebirdtech/shorebird/issues/1610.

Fixes https://github.com/shorebirdtech/shorebird/issues/1610

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [x] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #1613, opened 2024-01-03
chore: track pubspec.lock files

bryanoltman:
## Description

This will let us have a better idea of when our deps change, which may make it easier to track down the causes of bugs. Motivated by https://github.com/shorebirdtech/shorebird/pull/1612.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


bryanoltman:
Note: I didn't do this in `package:scoped` because there's a chance that might end up on pub.

felangel (commented):
Thoughts on only doing this for the CLI since we probably don't want to lock deps for packages like redis_client, shorebird_code_push etc.

https://dart.dev/guides/libraries/private-files

felangel (approved):

bryanoltman:
> Thoughts on only doing this for the CLI since we probably don't want to lock deps for packages like redis_client, shorebird_code_push etc.
> 
> https://dart.dev/guides/libraries/private-files

Yep, makes sense. Wanted to do this to be thorough, but I imagine we'll catch everything that changes as a transitive dependency in shorebird_cli's pubspec.lock.

PR #1616, opened 2024-01-04
refactor(shorebird_cli): throw error if incomplete engine config args are provided

bryanoltman:
## Description

Moves `EngineConfig` to its own file and checks that all values (engine, engine src, and host release) are provided if any of them are provided (i.e., they must all either be null or non-null). 

This also removes the `engineConfig` arg from ShorebirdProcess as it wasn't being used.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/shorebird_engine_config.dart
```diff
@@ -0,0 +1,39 @@
...
+    this.localEngineSrcPath,
+    this.localEngine,
+    this.localEngineHost,
+  }) {
+    final args = [localEngineSrcPath, localEngine, localEngineHost];
+    final allArgsAreNull = args.every((arg) => arg == null);
+    final allArgsAreNotNull = args.every((arg) => arg != null);
+    if (!allArgsAreNull && !allArgsAreNotNull) {
+      // Only some args were provided, this is invalid.
+      throw ArgumentError(
```
felangel: looks like we're missing a test for this?

packages/shorebird_cli/lib/src/shorebird_engine_config.dart
```diff
@@ -0,0 +1,39 @@
...
+        localEngine = null,
+        localEngineHost = null;
+
+  final String? localEngineSrcPath;
+  final String? localEngine;
+  final String? localEngineHost;
+
+  @override
+  String toString() {
+    return '''EngineConfig(localEngineSrcPath: $localEngineSrcPath, localEngine: $localEngine, localEngineHost: $localEngineHost)''';
```
felangel: same here

packages/shorebird_cli/lib/src/shorebird_engine_config.dart
```diff
@@ -0,0 +1,39 @@
...
+class EngineConfig {
+  EngineConfig({
+    this.localEngineSrcPath,
+    this.localEngine,
+    this.localEngineHost,
+  }) {
+    final args = [localEngineSrcPath, localEngine, localEngineHost];
+    final allArgsAreNull = args.every((arg) => arg == null);
+    final allArgsAreNotNull = args.every((arg) => arg != null);
+    if (!allArgsAreNull && !allArgsAreNotNull) {
```
felangel: Rather than doing this can we just have a two constructors? One that accepts no args and one that requires all of them? That we we retain a const constructor and you get errors are compile time rather than at run time.

felangel (approved):

packages/shorebird_cli/lib/src/shorebird_engine_config.dart
bryanoltman: Done

packages/shorebird_cli/lib/src/shorebird_engine_config.dart
bryanoltman: Done

packages/shorebird_cli/lib/src/shorebird_engine_config.dart
bryanoltman: I think there will always be the possibility of a runtime error (we can't know ahead of time what arguments the user will provide), but I've updated this to move the check out of EngineConfig and into the arg parsing code so it's a bit cleaner.

packages/shorebird_cli/lib/src/engine_config.dart
```diff
@@ -0,0 +1,29 @@
...
+final engineConfigRef = create(() => const EngineConfig.empty());
+
+// The [EngineConfig] instance available in the current zone.
+EngineConfig get engineConfig => read(engineConfigRef);
+
+class EngineConfig {
+  const EngineConfig({
+    this.localEngineSrcPath,
+    this.localEngine,
+    this.localEngineHost,
```
felangel:
```suggestion
    required this.localEngineSrcPath,
    required this.localEngine,
    required this.localEngineHost,
```

PR #1677, opened 2024-01-31
chore: Create dependabot.yml

eseidel:
Not sure if this is correct, but figured might as well give it a try.


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pub" # See documentation for possible values
+    directory: "/" # Location of package manifests
```
felangel: I think you need to explicitly include all the directories. See https://github.com/VeryGoodOpenSource/dart_frog/blob/main/.github/dependabot.yaml for an example.

felangel:
https://github.com/dependabot/dependabot-core/issues/2178

felangel (approved):

PR #1679, opened 2024-01-31
fix: update iOS warning to say "beta" and only show on patching.

eseidel:
Also updated release notes and version for 0.24.1.


felangel (approved):
LGTM 🚢  :shipit:

PR #1680, opened 2024-02-01
chore(shorebird_cli)!: remove `alpha` suffix from `ios` commands

felangel:
## Description

- chore(shorebird_cli): remove `alpha` suffix from `ios` commands

## Type of Change

- [x] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [X] 🗑️ Chore


eseidel (commented):
Do you think it's worth aliasing *-alpha?  Otherwise we will break people's CI scripts I suspect?

eseidel (approved):
Maybe we just bump the version to 0.25.0 and move on.

bryanoltman (approved):
We should probably do a docs pass as well

felangel:
> Maybe we just bump the version to 0.25.0 and move on.

Yeah went back and forth on it. Happy to add an alias for backward compat.

felangel:
> We should probably do a docs pass as well

Yup was planning to do that next.

felangel:
> Maybe we just bump the version to 0.25.0 and move on.

That’s what I was leaning toward but happy to adjust and add an alias for backward compat but we’ll break at some point anyway and if we bump the version to 0.25.0 I think it’s fine 🤷‍♂️

eseidel:
Let’s add an alias. Otherwise this has to match the shorebird setup
revision exactly too, or?

felangel:
> Let’s add an alias. Otherwise this has to match the shorebird setup revision exactly too, or?
> […](#)

Setup shorebird always uses the latest stable version of shorebird by default.

PR #1688, opened 2024-02-01
fix: Don't support releasing to known-bad versions of Flutter

eseidel:
We now know that Shorebird releases of Flutter 3.16.4-3.16.7 had
crashers on iOS.  This prevents users from releasing with iOS
to those versions and thus hitting those crashes.


eseidel (commented):
lgtm, but since I wrote parts of this, should get @felangel to review. :)

felangel (approved):

PR #1695, opened 2024-02-02
feat(shorebird_cli): print dsym download link after ios release

bryanoltman:
## Description

If a Flutter.dSYM.zip is available for the given engine rev, print a download link at the end of the release process.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


eseidel (changes requested):
I think 95% of our customers don't need this.  I think we should find some other place to put this.  Maybe in ci-only output, or verbose mode or links from the console or something?

PR #1707, opened 2024-02-06
fix: download.shorebird.dev symbol downloads not working

eseidel:
Turns out we were not pointing our manifest to symbols, nor
had we listed symbols as an explicitly allowed url.

Fixed both, and wrote a test which validates that any url in
our generated manifest is a url that the server actually knows
how to handle.


felangel (approved):

packages/artifact_proxy/test/artifact_proxy_test.dart
```diff
@@ -142,5 +146,32 @@ void main() {
       expect(response.headers['content-type'], equals('text/html'));
       expect(response.readAsString(), completion(contains('Shorebird')));
     });
+
+    test('generate_manifest matches config', () async {
+      // Make a temp directory, run generate_manifest, parse the yaml
+      // and make sure all urls are handled.
+      const engineRevision = '8b89f8bd9fc6982aa9c4557fd0e5e89db1ff9986';
+      final result = Process.runSync('/bin/sh', [
+        'tool/generate_manifest.sh',
```
felangel, line 156:
```suggestion
        p.join('tool', 'generate_manifest.sh'),
```

PR #1749, opened 2024-02-23
chore: change User.authProvider to User.jwtIssuer

bryanoltman:
## Description

Because JWT issuers can vary between Azure tenants, we're recording the full issuer instead of simply that it was Microsoft.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -361,7 +361,7 @@ extension OauthValues on oauth2.AuthProvider {
       case MicrosoftAuthProvider:
         return oauth2.ClientId(
           /// Shorebird CLI's OAuth 2.0 identifier for Azure/Entra.
-          'e389e829-42b2-47ab-8868-581475f90313',
+          '0ff83897-ec85-4642-a250-48d5f595137c',
         );
     }
 
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -22,13 +22,13 @@ final authRef = create(Auth.new);
 Auth get auth => read(authRef);
 
 /// Matches 'https://accounts.google.com' exactly.
-final googleJwtIssuerRegexp = RegExp(r'https:\/\/accounts\.google\.com');
+final googleJwtIssuerRegexp = RegExp(r'^https:\/\/accounts\.google\.com$');
 
 /// Matches lines like
 /// https://login.microsoftonline.com/{tenant-id}/v2.0/.well-known/openid-configuration
 /// Captures the tenant ID as a group.
 final microsoftJwtIssuerRegexp = RegExp(
-  r'https:\/\/login\.microsoftonline\.com\/([\w-]+)\/v2\.0',
+  r'^https:\/\/login\.microsoftonline\.com\/([\w-]+)\/v2\.0$',
 );
 
 typedef ObtainAccessCredentials = Future<oauth2.AccessCredentials> Function(
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -21,15 +21,13 @@ final authRef = create(Auth.new);
 // The [Auth] instance available in the current zone.
 Auth get auth => read(authRef);
 
-/// Matches 'https://accounts.google.com' exactly.
-final googleJwtIssuerRegexp = RegExp(r'^https:\/\/accounts\.google\.com$');
+/// The JWT issuer field for Google-issued JWTs.
+const googleJwtIssuer = 'https://accounts.google.com';
 
-/// Matches lines like
-/// https://login.microsoftonline.com/{tenant-id}/v2.0/.well-known/openid-configuration
-/// Captures the tenant ID as a group.
-final microsoftJwtIssuerRegexp = RegExp(
-  r'^https:\/\/login\.microsoftonline\.com\/([\w-]+)\/v2\.0$',
-);
+/// Microsoft-issued JWTs are of the form
+/// https://login.microsoftonline.com/{tenant-id}/v2.0. We don't care about the
+/// tenant ID, so we just match the prefix.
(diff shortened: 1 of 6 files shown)


packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -21,6 +21,16 @@ final authRef = create(Auth.new);
 // The [Auth] instance available in the current zone.
 Auth get auth => read(authRef);
 
+/// Matches 'https://accounts.google.com' exactly.
+final googleJwtIssuerRegexp = RegExp(r'https:\/\/accounts\.google\.com');
```
eseidel: To match it exactly, you'd need $ and ^ I think to mark start and end?

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -21,6 +21,16 @@ final authRef = create(Auth.new);
...
+
+/// Matches lines like
+/// https://login.microsoftonline.com/{tenant-id}/v2.0/.well-known/openid-configuration
+/// Captures the tenant ID as a group.
+final microsoftJwtIssuerRegexp = RegExp(
+  r'https:\/\/login\.microsoftonline\.com\/([\w-]+)\/v2\.0',
```
eseidel: Same comment re: start/end.  I'm not sure. how important that is?  Would an attacker get anything by sending us a jwt issued from something similar-but-not-identical to Google/Microsoft? e.g.

http://haxor.com/https://google.com/

packages/shorebird_cli/test/src/auth/auth_test.dart
```diff
@@ -109,7 +112,7 @@ void main() {
 
       group('when issuer is accounts.google.com', () {
         setUp(() {
-          when(() => payload.iss).thenReturn('https://accounts.google.com');
+          when(() => payload.iss).thenReturn(googleJwtIssuer);
```
eseidel, line 115: I'm glad to see the continuing move towards constants.

packages/shorebird_cli/test/src/auth/auth_test.dart
```diff
@@ -142,7 +145,11 @@ void main() {
     const idToken =
         '''eyJhbGciOiJIUzI1NiIsImtpZCI6IjEyMzQiLCJ0eXAiOiJKV1QifQ.eyJpc3MiOiJodHRwczovL2FjY291bnRzLmdvb2dsZS5jb20iLCJhenAiOiI1MjMzMDIyMzMyOTMtZWlhNWFudG0wdGd2ZWsyNDB0NDZvcmN0a3RpYWJyZWsuYXBwcy5nb29nbGV1c2VyY29udGVudC5jb20iLCJhdWQiOiI1MjMzMDIyMzMyOTMtZWlhNWFudG0wdGd2ZWsyNDB0NDZvcmN0a3RpYWJyZWsuYXBwcy5nb29nbGV1c2VyY29udGVudC5jb20iLCJzdWIiOiIxMjM0NSIsImhkIjoic2hvcmViaXJkLmRldiIsImVtYWlsIjoidGVzdEBlbWFpbC5jb20iLCJlbWFpbF92ZXJpZmllZCI6dHJ1ZSwiaWF0IjoxMjM0LCJleHAiOjY3ODl9.MYbITALvKsGYTYjw1o7AQ0ObkqRWVBSr9cFYJrvA46g''';
     const email = 'test@email.com';
-    const user = User(id: 42, email: email, authProvider: AuthProvider.google);
+    const user = User(
+      id: 42,
+      email: email,
+      jwtIssuer: googleJwtIssuer,
```
eseidel, line 151: We're probably going to end up with these being enums and just getting the .toJson value out of them as a url for the db, rather than using strings inside our models. 🤷

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -320,10 +330,10 @@ class UserNotFoundException implements Exception {
 
 extension OauthAuthProvider on Jwt {
   oauth2.AuthProvider get authProvider {
-    if (payload.iss.startsWith('https://login.microsoftonline.com')) {
-      return MicrosoftAuthProvider();
-    } else if (payload.iss == 'https://accounts.google.com') {
+    if (googleJwtIssuerRegexp.hasMatch(payload.iss)) {
```
eseidel: Instead of having this thing do the iss -> auth provider, you might consider having a middle step which is iss string -> enum -> auth-provider.  Then the string -> enum/object could hide all the regexp goop.  🤷

eseidel (approved):
My only concern is that this has moved from testing "startsWith" to testing "contains" which may not be what you want. I'm not sure you want to use regexp here (since you dont' seem to ever used the captured value?).

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: Good call, fixed

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: I don't think they would. I've updated to no longer use regex and to do an exact match for google (because the issuer is always `https://accounts.google.com`) and to use a prefix match for microsoft (whose issuers are of the form `https://login.microsoftonline.com/{tenant-id}/v2.0`.

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: Regex goop removed, although I'm not sure that removing it addresses this comment.

I think our VC convo got cut short as you were explaining this, but I'm not sure I understand how adding an intermediate step would make this cleaner.

packages/shorebird_cli/test/src/auth/auth_test.dart
bryanoltman: I don't think it's required that the `iss` value needs to be a well-formed URL (see https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1).

PR #1758, opened 2024-02-27
fix(shorebird_cli): get MS auth token, add token provider env var for CI

bryanoltman:
## Description

Adds `SHOREBIRD_TOKEN_PROVIDER` environment variable requirement for CI login. This lets us know which auth provider we should use to obtain a JWT.

This also adds the `offline_access` scope to our Azure scopes to allow us to obtain a refresh token.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


eseidel:
Could we just embed this in the token?

felangel:
> Could we just embed this in the token?

Yeah I agree, we could base64 encode the provider + refresh token 🤷

bryanoltman:
Created a CiToken type whose base64-representation will be the SHOREBIRD_TOKEN env var

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -248,9 +263,15 @@ class Auth {
...
 
   void _loadCredentials() {
-    final token = platform.environment['SHOREBIRD_TOKEN'];
-    if (token != null) {
-      _token = token;
+    final base64Token = platform.environment[shorebirdTokenEnvVar];
+    if (base64Token != null) {
+      try {
+        _token = CiToken.fromBase64(base64Token);
+      } catch (_) {
```
eseidel, line 270:
```suggestion
      } catch (_) {
      // TODO(bryanoltman): Remove this legacy behavior after July 2024 or next major release.
      logger.warn('Token failed to parse, assuming an older Google token and re-trying.');
      _token = CiToken(base64Token, googleAuth);
```

eseidel: Not sure if it's worth it, but we could do something like this. 🤷

eseidel (approved):
Consider not forcing existing customers using CI to re-login.  otherwise lgtm.

packages/shorebird_cli/lib/src/auth/auth.dart
bryanoltman: Good call, done. Added more context to the message, along with what steps the user could take to resolve the warning.

PR #1785, opened 2024-03-11
fix: only build patch artifacts once if release version is specified and flutter version is not current

### LINKED ISSUE #1784: Patch builds fail if app cannot build with latest version of Flutter
1. A user created a release with Flutter version A
2. Shorebird updated to Flutter version B
3. The user's app does not build with Flutter version A due to dependency version conflicts
4. User cannot patch their release because our patch commands build the app with Flutter version B to determine the release version, even if a release version is provided as an argument to the patch command.

See https://discord.com/channels/1030243211995791380/1215306109137584128/1216795988237222100 for context.

bryanoltman:
## Description

Updates `patch ios` and `patch android` commands to use the `release-version` parameter (if specified) instead of always building the patch artifact to determine the release version.

Fixes https://github.com/shorebirdtech/shorebird/issues/1784

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
@@ -112,19 +112,11 @@ If this option is not provided, the version number will be determined from the p
     final shorebirdYaml = shorebirdEnv.getShorebirdYaml()!;
     final appId = shorebirdYaml.getAppId(flavor: flavor);
     final app = await codePushClientWrapper.getApp(appId: appId);
-
     final originalFlutterRevision = shorebirdEnv.flutterRevision;
-    final buildProgress = logger.progress('Building patch');
-    try {
-      await buildAppBundle(flavor: flavor, target: target);
-      buildProgress.complete();
-    } on ProcessException catch (error) {
-      buildProgress.fail('Failed to build: ${error.message}');
-      return ExitCode.software.code;
-    }
+
+    var hasBuiltWithLatestFlutter = false;
 
     final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
-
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -112,19 +112,11 @@ If this option is not provided, the version number will be determined from the p
...
-    try {
-      await buildAppBundle(flavor: flavor, target: target);
-      buildProgress.complete();
-    } on ProcessException catch (error) {
-      buildProgress.fail('Failed to build: ${error.message}');
-      return ExitCode.software.code;
-    }
 
-    final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
+    var hasBuiltWithLatestFlutter = false;
```
eseidel, line 117: Is this just "haveBuilt"?

eseidel: Or maybe "haveBuiltWithActiveFlutter"?  Which just gets set to "false" whenever we switch?  And then the switch and not-switch cases can share build logic?

eseidel: I see, they can't actually share due to the finally clause in the switching case.  Maybe that all should be moved into a "Build with Fltuter version" which checks if the asked for version is the same as active?

eseidel (approved):
I got a bit confused by the "latest flutter" part of the variable name, but otherwise looks reasonable.  We definitely need to come in and clean up this code more.

felangel (approved):

PR #1790, opened 2024-03-13
fix(shorebird_cli): report correct flutter rev when creating release

bryanoltman:
## Description

Fixes a bug where we would record the incorrect Flutter revision for releases created with the non-default Flutter revision.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_aar_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_aar_command.dart
@@ -105,7 +105,7 @@ make smaller updates to your app.
       );
     }
 
-    var flutterRevision = shorebirdEnv.flutterRevision;
+    var flutterRevisionForRelease = shorebirdEnv.flutterRevision;
     if (flutterVersion != null) {
       final String? revision;
       try {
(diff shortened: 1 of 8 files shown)


packages/shorebird_cli/lib/src/commands/release/release_aar_command.dart
```diff
@@ -202,7 +202,7 @@ ${summary.join('\n')}
       release = await codePushClientWrapper.createRelease(
         appId: appId,
         version: releaseVersion,
-        flutterRevision: shorebirdEnv.flutterRevision,
+        flutterRevision: flutterRevision,
```
eseidel:
```suggestion
        // Intentionally passing flutterRevision rather than shorebirdEnv.flutterRevision
        // since we've already reverted shorebirdEnv back to the latest flutter at this point.
        flutterRevision: flutterRevision,
```

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -257,7 +257,7 @@ ${summary.join('\n')}
       release = await codePushClientWrapper.createRelease(
         appId: appId,
         version: releaseVersion,
-        flutterRevision: shorebirdEnv.flutterRevision,
+        flutterRevision: flutterRevision,
```
eseidel:
```suggestion
        // Intentionally passing flutterRevision rather than shorebirdEnv.flutterRevision
        // since we've already reverted shorebirdEnv back to the latest flutter at this point.
        flutterRevision: flutterRevision,
```

eseidel (approved):

packages/shorebird_cli/lib/src/commands/release/release_aar_command.dart
bryanoltman: Updated the variable name to make it clear that we're passing the release flutter version

PR #1796, opened 2024-03-15
fix(shorebird_cli): run all release and patch commands with a scoped shorebirdEnv

### LINKED ISSUE #1794: fix: shorebird patch fails when switching flutter revisions
**Description**

Running `shorebird patch ios` for a release that was built using an older flutter revision fails due to:

```
[WARN] iOS support is beta. Some apps may run slower after patching.
See ]8;;https://docs.shorebird.dev/status\https://docs.shorebird.dev/status]8;;\ for more information.
⠋ Fetching apps...
✓ Fetching apps (0.3s)
⠋ Building patch...
✓ Building patch (190.2s)
Detected release version 2.0.0+02000
⠋ Fetching releases...
✓ Fetching releases (99ms)

The release you are trying to patch was built with a different version of Flutter.

Release Flutter Revision: 771d07b2cf97cf107bae6eeedcf41bdc9db772fa
Current Flutter Revision: 5b9d29d67adb059103beefb65710ee3dabae2f85

⠋ Building patch...⠋ Switching to Flutter revision 771d07b2cf97cf107bae6eeedcf41bdc9db772fa...
✓ Switching to Flutter revision 771d07b2cf97cf107bae6eeedcf41bdc9db772fa (7.4s)
⠋ Switching back to original Flutter revision 5b9d29d67adb059103beefb65710ee3dabae2f85...
✓ Switching back to original Flutter revision 5b9d29d67adb059103beefb65710ee3dabae2f85 (4ms)

✗ Exception: Failed to create snapshot: Can't load Kernel binary: Invalid kernel binary format version.
 (163.1s)


Build failed :|
Step 7 script `🚀 Shorebird Patch` exited with status code 70
```

I suspect the issue is we're not using the correct Flutter revision when executing `gen_snapshot` because `ShorebirdArtifacts` uses `ShorebirdEnv` to resolve the artifact and in this case we're building using a flutter revision override which isn't propagating when building the snapshot.

`buildElfAotSnapshot` likely also needs to take a `flutterRevision` and do the correct lookup similar to how `buildIpa` was modified to require a specific `flutterRevision`.

Tagging @bryanoltman since it's related to #1790 (which looks like it was an incomplete fix).

bryanoltman:
## Description

Updates all patch and release commands to run build operations with a scoped `shorebirdEnv`, which is used to determine the current Flutter revision.

The diff is ugly, but broadly, this change simply creates a new `shorebirdEnv` using `copyWith` once we know which Flutter revision we're using to create the build, and wraps all build operations in `runScoped` with that shorebirdEnv as an override.

Fixes https://github.com/shorebirdtech/shorebird/issues/1794

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
@@ -140,6 +140,12 @@ Use `shorebird flutter versions list` to list available versions.
       flutterRevisionForRelease = revision;
     }
 
+    final flutterInstallProgress = logger.progress(
+      'Installing Flutter $flutterRevisionForRelease',
+    );
+    await shorebirdFlutter.installRevision(revision: flutterRevisionForRelease);
+    flutterInstallProgress.complete();
+
     final releaseFlutterShorebirdEnv = shorebirdEnv.copyWith(
       flutterRevisionOverride: flutterRevisionForRelease,
     );
--- a/packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
+import 'package:scoped/scoped.dart';
 import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
 import 'package:shorebird_cli/src/command.dart';
 import 'package:shorebird_cli/src/config/shorebird_yaml.dart';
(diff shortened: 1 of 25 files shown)


packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -139,162 +140,165 @@ Use `shorebird flutter versions list` to list available versions.
...
-    if (switchFlutterRevision) {
-      await shorebirdFlutter.useRevision(revision: flutterRevisionForRelease);
-    }
-
-    final flutterVersionString = await shorebirdFlutter.getVersionAndRevision();
-
-    final buildProgress = logger.progress(
-      'Building release with Flutter $flutterVersionString',
+    final flutterInstallProgress = logger.progress(
+      'Installing Flutter $flutterRevisionForRelease',
```
felangel: So do we always show this log even when the revision of flutter is already installed?

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -139,162 +140,165 @@ Use `shorebird flutter versions list` to list available versions.
...
     );
+    await shorebirdFlutter.installRevision(revision: flutterRevisionForRelease);
+    flutterInstallProgress.complete();
```
felangel: Minor but if `installRevision` throws, then the progress will never complete which will look a bit weird to users.

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -139,162 +140,165 @@ Use `shorebird flutter versions list` to list available versions.
...
-        flutterRevisionForRelease != originalFlutterRevision;
-
-    if (switchFlutterRevision) {
-      await shorebirdFlutter.useRevision(revision: flutterRevisionForRelease);
-    }
-
-    final flutterVersionString = await shorebirdFlutter.getVersionAndRevision();
-
-    final buildProgress = logger.progress(
-      'Building release with Flutter $flutterVersionString',
```
felangel, line 153: It was nice to have this log framed as `"Building with Flutter <revision>"` instead of `installing Flutter` since we don't always need to install a revision of Flutter right? Most times, the revision they use will be default version that is already installed. I think it would be nice to keep the old copy wdyt?

packages/shorebird_cli/test/src/commands/release/release_aar_command_test.dart
```diff
@@ -336,22 +351,26 @@ $exception''',
...
-            'and reverts to original flutter version', () async {
+        test('uses specified flutter version to build', () async {
           setUpProjectRootArtifacts();
+          when(
+            () => shorebirdProcess.run(
+              any(),
+              any(that: containsAll(['build', 'aar'])),
+              runInShell: any(named: 'runInShell'),
+            ),
+          ).thenAnswer((invocation) async {
```
felangel:
```suggestion
          ).thenAnswer((_) async {
```

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -136,6 +136,16 @@ void main() {
       shorebirdValidator = MockShorebirdValidator();
 
       when(() => shorebirdEnv.getShorebirdYaml()).thenReturn(shorebirdYaml);
+      when(
+        () => shorebirdEnv.copyWith(
+          flutterRevisionOverride: any(named: 'flutterRevisionOverride'),
+        ),
+      ).thenAnswer((invocation) {
```
felangel, line 143:
```suggestion
      ).thenAnswer((_) {
```

packages/shorebird_cli/test/src/commands/release/release_ios_command_test.dart
```diff
@@ -211,9 +211,24 @@ flutter:
       ).thenReturn(projectRoot);
       when(() => shorebirdEnv.flutterRevision).thenReturn(flutterRevision);
       when(() => shorebirdEnv.isRunningOnCI).thenReturn(false);
+      when(
+        () => shorebirdEnv.copyWith(
+          flutterRevisionOverride: any(named: 'flutterRevisionOverride'),
+        ),
+      ).thenAnswer((invocation) {
```
felangel, line 218:
```suggestion
      ).thenAnswer((_) {
```

felangel (approved):

packages/shorebird_cli/test/src/commands/release/release_aar_command_test.dart
bryanoltman: Done

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
bryanoltman: We print both "installing" and "building", the diff is just bad.

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
bryanoltman: Yep, but I'll fix that and add error handling for install failures. One sec.

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
bryanoltman: On it

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
bryanoltman: Updated

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
bryanoltman: Fixed

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -38,7 +38,7 @@ class ShorebirdFlutter {
     final version = await getVersionString(revision: revision);
 
     final installProgress =
-        logger.progress('Installing Flutter revision $version ($revision)');
+        logger.progress('Installing Flutter $version ($revision)');
```
felangel: minor but maybe we can truncate the revision to just the first 10 characters?

bryanoltman: Done

PR #1801, opened 2024-03-18
feat(shorebird_cli): replace `force` flag with `allow-native-diffs` and `allow-asset-diffs`

bryanoltman:
## Description

Removes the `force` flag on all release and patch commands.

Adds `allow-native-diffs` and `allow-asset-diffs` on patch commands to ignore warnings that native code changes or asset changes were detected between the patch and release builds. 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
```diff
@@ -52,9 +52,17 @@ of the Android app that is using this module.''',
...
+        help: '''
+Patch even if native code diffs are detected.
+NOTE: this is **not** recommended.''',
+        negatable: false,
+      )
+      ..addFlag(
+        'allow-asset-diffs',
+        help: '''
+Patch even if asset diffs are detected
+NOTE: this is **not** recommended.''',
```
eseidel: This one is less scary than the native one.  We might want to give context on why or link to docs?

eseidel (commented):
I guess my only question is if we should offer any sort of transition period for people, or if we should just bump our minor version number and break folks who are using --force.

packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
```diff
@@ -88,13 +99,17 @@ of the Android app that is using this module.''',
     }
 
     final force = results['force'] == true;
-    final dryRun = results['dry-run'] == true;
-
-    if (force && dryRun) {
-      logger.err('Cannot use both --force and --dry-run.');
+    if (force) {
+      logger
+        ..err(PatchCommand.forceDeprecationErrorMessage)
```
felangel, line 104: Isn't this technically a breaking change and not a deprecation if we're changing the behavior of `--force`? Shouldn't we not exit to maintain backward compatibility here?

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -14,6 +15,33 @@ class PatchCommand extends ShorebirdCommand {
...
+  static const forceDeprecationErrorMessage =
+      'The --force flag has been deprecated';
+
+  static final forceDeprecationExplanation =
+      'If your app has known safe native code or asset changes, you can use '
+      'the ${cyan.wrap('--allow-native-diffs')} or '
+      '${cyan.wrap('--allow-asset-diffs')} '
+      'flags. We do not recommend using these flags unless you are '
+      '${styleBold.wrap('absolutely sure')} that the changes are safe. '
+      'Note: the --force flag is not required for use in CI environments.';
```
felangel:
```suggestion
      'Note: the ${cyan.wrap('--force flag')} is not required for use in CI environments.';
```

felangel (commented):
LGTM except if we're deprecating the `--force` flag we shouldn't break compatibility in this change and should instead preserve the behavior until the next major release imo.

packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
bryanoltman: Discussed offline. Resolving.

packages/shorebird_cli/lib/src/commands/patch/patch_aar_command.dart
bryanoltman: Added more info

PR #1810, opened 2024-03-22
feat: use new linker pipeline on iOS

bryanoltman:
Teach `shorebird patch ios` how to use the new linker pipeline when available.


eseidel:
Is this what I need to test the new object pool linker? (I presume so).

bryanoltman:
Yep - this just adds the gen snapshot path and path to the kernel file to the aot-tools link command call

packages/shorebird_cli/lib/src/executables/aot_tools.dart
```diff
@@ -109,20 +111,39 @@ class AotTools {
...
   }
 
+  Future<bool> _linkerUsesGenSnapshot() async {
+    final version = await _getVersion();
+    return version != null && version >= Version(0, 0, 1);
+  }
+
+  Future<Version?> _getVersion() async {
+    final result = await _exec(['--version']);
+    if (result.exitCode != 0) {
```
felangel:
```suggestion
    if (result.exitCode != ExitCode.success.code) {
```
nit

felangel (approved):

PR #1823, opened 2024-03-26
chore(shorebird_code_push_protocol): Add metadata messages

bryanoltman:
Adds metadata classes for recording information about release and patch command runs. Will be used to more effectively diagnose release and patch issues and to better understand how shorebird is being used.

PR #1824, opened 2024-03-26
feat(shorebird_cli): record release and patch command metadata

bryanoltman:
Updates the patch and release commands to send metadata introduced in https://github.com/shorebirdtech/shorebird/pull/1823 when creating patches and releases.


packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -245,16 +246,14 @@ class CodePushClient {
...
   }) async {
     final request = CreatePatchRequest(
       releaseId: releaseId,
       wasForced: false,
-      hasAssetChanges: hasAssetChanges,
-      hasNativeChanges: hasNativeChanges,
-      // TODO(bryanoltman): add metadata
-      metadata: null,
+      hasAssetChanges: metadata.hasAssetChanges,
+      hasNativeChanges: metadata.hasNativeChanges,
```
felangel, line 255: Why not remove the redundant `hasAssetChanges` and `hasNativeChanges` here since they are part of the metadata?

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -284,16 +286,27 @@ ${summary.join('\n')}
...
           status: ReleaseStatus.active,
+          metadata: UpdateReleaseMetadata(
+            releasePlatform: releasePlatform,
+            flutterVersionOverride: flutterVersion,
+            generatedApks: generateApk,
+            environment: BuildEnvironmentMetadata(
+              operatingSystem: platform.operatingSystem,
+              operatingSystemVersion: platform.operatingSystemVersion,
+              shorebirdVersion: packageVersion,
+              xcodeVersion: null,
```
felangel, line 307: I suspect we'll want to add AndroidStudio version as well?

packages/shorebird_code_push_client/lib/src/code_push_client.dart
bryanoltman: Can do now that the backend change is rolled out. Updating

bryanoltman: Actually, the server still expects this. Will remove once we roll out a new version of the CLI that includes this PR

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
bryanoltman: Yep, although we don't need it immediately. Will file an issue to add.

eseidel (approved):

eseidel:
I didn't read that closely, seems totally reasonable.

Makes me wonder if Metadata wont' want something looser/lighter weight like just a map. 🤷

PR #1845, opened 2024-04-01
fix(shorebird_cli): don't prompt users for confirmation if we can't accept input via stdin

### LINKED ISSUE #1705: fix: `shorebird` hangs when run under `fastlane`
"""
I included the shorebird command in my fastlane pipeline for ease of use and this is where it never continues after it packages everything up. I then tried running the command outside of fastlane (thinking it might be some directory issue, it occured in the past with different tools), and that's when I found out why it never finished via fastlane. Shorebird asks you if you want to create a new release ("Would you like to continue?"), and that question is not visible if you run the command via fastlane. So just had to add the --force command for the release build that I was doing via fastlane and everything works. 😂
"""

1. We should detect this case (fastlane is pretty common, or was at one point).  And presumably we're headless in that environment (don't have a TERM?)
2. We should discourage use of --force and instead some sort of --no-confirm flag for situations like this we can't detect.

bryanoltman:
## Description

If running in a terminal-less environment, Shorebird should not prompt for user confirmation.

Fixes https://github.com/shorebirdtech/shorebird/issues/1705

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -154,7 +154,7 @@ Please make sure you are running "shorebird init" from within your Flutter proje
     final String appId;
     Map<String, String>? flavors;
     try {
-      final needsConfirmation = !force && !shorebirdEnv.isRunningOnCI;
+      final needsConfirmation = !force && shorebirdEnv.canAcceptUserInput;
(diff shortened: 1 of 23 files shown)


eseidel (approved):
lgtm.  We may still want to consider a "--no-confirm" or "--non-interactive" or whatever the pattern is.

PR #1849, opened 2024-04-02
feat(shorebird_cli): suggest running with verbose on command failure

### LINKED ISSUE #1848: feat: prompt to re-run command with `--verbose` flag on failure
When a command fails, we should prompt a user to re-run the command with the --verbose flag to help them diagnose the issue.

bryanoltman:
## Description

If a command fails with a `software` exit code (what we conventionally use for exceptions and general "something went wrong" errors), suggest that the user re-run the command with `--verbose` if they aren't already using that flag.

Fixes https://github.com/shorebirdtech/shorebird/issues/1848

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/command_runner.dart
+++ b/packages/shorebird_cli/lib/src/command_runner.dart
@@ -181,7 +181,7 @@ ${lightCyan.wrap('shorebird release android -- --no-pub lib/main.dart')}''';
     }
 
     // Run the command or show version
-    final int? exitCode;
+    int? exitCode;
     if (topLevelResults['version'] == true) {
       final flutterVersion = await _tryGetFlutterVersion();
       final shorebirdFlutterPrefix = StringBuffer('Flutter');
@@ -205,7 +205,22 @@ Run ${lightCyan.wrap('shorebird upgrade')} to upgrade.''');
       }
       exitCode = ExitCode.success.code;
     } else {
-      exitCode = await super.runCommand(topLevelResults);
+      try {
+        exitCode = await super.runCommand(topLevelResults);
+      } catch (error, stackTrace) {
+        logger
+          ..err('$error')
+          ..info('$stackTrace');
+        exitCode = ExitCode.software.code;
+      }
+    }
+
+    if (exitCode == ExitCode.software.code && logger.level != Level.verbose) {
+      logger.info(
+        '''
+
+If you aren't sure why this command failed, re-run with the ${lightCyan.wrap('--verbose')} flag to see more information.''',
+      );
     }
 
     return exitCode;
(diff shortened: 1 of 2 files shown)


erickzanardo (approved):

packages/shorebird_cli/test/src/command_runner_test.dart
```diff
@@ -274,6 +271,35 @@ Run ${lightCyan.wrap('shorebird upgrade')} to upgrade.'''),
...
     });
 
+    group('on command failure', () {
+      group('when running with --verbose', () {
+        setUp(() {
+          when(() => logger.level).thenReturn(Level.verbose);
+        });
+
+        test('does not suggest running with --verbose', () async {
+          // This will fail due to missing scoped overrides.
```
felangel: What do you mean by this comment?

packages/shorebird_cli/test/src/command_runner_test.dart
```diff
@@ -274,6 +271,35 @@ Run ${lightCyan.wrap('shorebird upgrade')} to upgrade.'''),
...
+          await runWithOverrides(
+            () => commandRunner.run(['release', 'android', '--verbose']),
+          );
+          verifyNever(() => logger.info(any(that: contains('--verbose'))));
+        });
+      });
+
+      group('when running without --verbose', () {
+        test('suggests using --verbose flag', () async {
+          // This will fail due to missing scoped overrides.
```
felangel: Same here I'm not sure I understand what the comment means.

packages/shorebird_cli/lib/src/command_runner.dart
```diff
@@ -181,7 +181,7 @@ ${lightCyan.wrap('shorebird release android -- --no-pub lib/main.dart')}''';
     }
 
     // Run the command or show version
-    final int? exitCode;
+    int? exitCode;
```
felangel, line 184: why did we have to make this non-final?

felangel (approved):
LGTM just left a few minor questions

packages/shorebird_cli/lib/src/command_runner.dart
bryanoltman: Mostly because the analyzer isn't smart enough to detect that `exitCode = await super.runCommand(topLevelResults);` is the only statement in the `try` block. The analyzer thinks there's a chance that it will get assigned a value and that the `catch` block (where it's assigned a value) may also execute.

packages/shorebird_cli/test/src/command_runner_test.dart
bryanoltman: Improved this. The release android command being invoked below will fail because `runWithOverrides` does not provide the scoped dependencies required.

packages/shorebird_cli/test/src/command_runner_test.dart
bryanoltman: Same as above

PR #1868, opened 2024-04-04
chore(shorebird_code_push_client): support semantically correct response codes

bryanoltman:
## Description

In some create/delete functions, we check for status code 200 (OK) when a 201 (created) or a 204 (no content) would be more appropriate. This PR adjusts the code push client to accept both, which will allow us to correct this server-side.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore


erickzanardo (approved):
LGTM

Just wonder about unit tests that could be updated to verify for the new behaviour.

felangel (approved):
LGTM in addition to unit tests we should probably leave TODOs to remove the old, deprecated response codes in the future so they don't linger for a long time.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -140,7 +140,8 @@ class CodePushClient {
     final response = await _httpClient.send(request);
     final body = await response.stream.bytesToString();
 
-    if (response.statusCode != HttpStatus.ok) {
+    if (response.statusCode != HttpStatus.ok &&
```
felangel: Might also be nice to have an extension on HttpStatus like `isCreatedOrOk` to eliminate some of the duplication and then we can leave a todo on the extension method to remove it once we've rolled out the backend changes and done a force upgrade in the future.

bryanoltman: I'd be curious to hear your thoughts re: having an even more general `isSuccess` extension, something like:

```dart
bool get isSuccess => statusCode >= 200 && statusCode < 300;
```

bryanoltman: Or maybe `< 400`, depending on our feelings about redirects.

bryanoltman: Went ahead and did this. It's a quite a bit less rigid than the previous implementation but I don't think we're losing anything in terms of correctness.

erickzanardo (approved):

PR #1885, opened 2024-04-09
test(shorebird_cli): increase timeout for `extractZip` large test

felangel:
## Description

- tests(shorebird_cli): increase timeout for `extractZip` large test

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [X] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/test/src/artifact_manager_test.dart
+++ b/packages/shorebird_cli/test/src/artifact_manager_test.dart
@@ -220,9 +220,11 @@ void main() {
           );
           expect(outDir.listSync(recursive: true), hasLength(1));
           expect(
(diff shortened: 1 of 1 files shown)


eseidel:
It seems like this test will be dependent on the file system's reading pattern. e.g. if it used a memory file system all the read might come back in one go and thus never be truly async and fail?

felangel:
> It seems like this test will be dependent on the file system's reading pattern. e.g. if it used a memory file system all the read might come back in one go and thus never be truly async and fail?

yeah the idea behind this test was to be more of an e2e test but given how slow/flaky it appears to be might be better to refactor this code to make it mockable since we really want to just make sure the input stream isn't closed before the zip has been extracted.

PR #1906, opened 2024-04-16
fix(shorebird_cli): flaky windows failure due to `pkg:archive`

### LINKED ISSUE #1902: tests: flaky windows preview test
Example Run: https://github.com/shorebirdtech/shorebird/actions/runs/8698822023/job/23856415749

```
test\src\commands\preview_command_test.dart: PreviewCommand android setChannelOnAab when channel is not set when target channel is not production sets shorebird.yaml channel to target channel (failed)
  PathAccessException: Deletion failed, path = 'C:\Users\RUNNER~1\AppData\Local\Temp\8ae70a41' (OS Error: The process cannot access the file because it is being used by another process.
  , errno = 32)
  dart:io                                                         FileSystemEntity.deleteSync
  package:shorebird_cli/src/commands/preview_command.dart 518:15  PreviewCommand.setChannelOnAab.<fn>
```

felangel:
## Description

- fix(shorebird_cli): flaky windows failure due to `pkg:archive` (closes #1902)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/pubspec.yaml
+++ b/packages/shorebird_cli/pubspec.yaml
@@ -9,7 +9,15 @@ environment:
   sdk: ">=3.0.0 <4.0.0"
 
 dependencies:
-  archive: ^3.3.6
+  # We're using a specific commit because the latest published version (3.4.10)
+  # does not include a sync close API for ZipFileEncoder.
+  # https://github.com/brendan-duncan/archive/commit/94bf62410e9a073451bda0d851a07e8b8863bde4
+  # This was causing a flaky test on windows when using `shorebird preview`
+  # https://github.com/shorebirdtech/shorebird/issues/1902
+  archive:
+    git:
+      url: https://github.com/brendan-duncan/archive
+      ref: 6b5bf69f76164c7dad164555491e18d5be683785
   args: ^2.3.1
   barbecue: ^0.5.0
   checked_yaml: ^2.0.2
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/pubspec.yaml
```diff
@@ -9,7 +9,15 @@ environment:
...
 dependencies:
-  archive: ^3.3.6
+  # We're using a specific commit because the latest published version (3.4.10)
+  # does not include a sync close API for ZipFileEncoder.
+  # https://github.com/brendan-duncan/archive/commit/94bf62410e9a073451bda0d851a07e8b8863bde4
+  # This was causing a flaky test on windows when using `shorebird preview`
+  # https://github.com/shorebirdtech/shorebird/issues/1902
+  archive:
+    git:
+      url: https://github.com/brendan-duncan/archive
```
bryanoltman, line 20: Maybe leave a TODO here noting that we should switch away from a specific commit once a release containing the API we need is published

bryanoltman (approved):

packages/shorebird_cli/pubspec.yaml
erickzanardo, line 20: Nit since this might be exaggerated concerns: Usually when I stumble in cases like this, I tend to fork the repo under a org that I have control, just to be safe that this commit will not vanish.

I know the odds of that happening is not much though, but it don't hurt to be overzealous haha

erickzanardo (approved):

packages/shorebird_cli/pubspec.yaml
felangel: Yeah I thought about that but it seemed like overkill. Ideally we'll move away from this as soon as the next version is published.

PR #1907, opened 2024-04-16
feat(shorebird_cli): add `ShorebirdYamlAssetValidator`

### LINKED ISSUE #1888: feat: shorebird doctor --fix should fix the pubspec.yaml file
**Description**

When running `shorebird doctor --fix` in a project where has a `shorebird.yaml` file, but don't have all the changes needed (shorebird.yaml file on the pubspec.yaml and android internet permission), the command just adds the android permission, it should also change the pubspec.yaml to add the shorebird.yaml file to the assets.

**Requirements**

- [ ] add the shorebird.yaml file to the assets.

borjandev:
## Description

Includes a validator, doctor, and doctor fix integration for :

```
flutter:
  assets:
    - shorebird.yaml
```

Attempts to handle most common, as well as not so common edge cases, includes tests.

Resolves https://github.com/shorebirdtech/shorebird/issues/1888

Screenshot 2024-04-16 at 19 07 35

-----

Screenshot 2024-04-16 at 19 08 02

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/doctor.dart
```diff
@@ -28,6 +28,7 @@ class Doctor {
     ShorebirdFlutterValidator(),
     AndroidInternetPermissionValidator(),
     StorageAccessValidator(),
+    ShorebirdYAMLAssetValidator(),
```
bryanoltman: Nit: YAML -> Yaml

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,108 @@
+import 'dart:io';
+import 'package:path/path.dart' as p;
+import 'package:shorebird_cli/src/shorebird_env.dart';
+import 'package:shorebird_cli/src/validators/validators.dart';
+import 'package:yaml/yaml.dart';
+import 'package:yaml_edit/yaml_edit.dart';
+
+class ShorebirdYAMLAssetValidator extends Validator {
```
bryanoltman: Needs doc comments throughout

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,108 @@
...
+    if (root == null) return null;
+    return File(p.join(root.path, 'pubspec.yaml'));
+  }
+
+  bool _pubspecYamlHasShorebirdAsset(YamlMap pubspecYaml) {
+    final flutterSection = pubspecYaml['flutter'];
+    if (flutterSection is! YamlMap) return false;
+
+    final assetsSection = flutterSection['assets'];
+    if (assetsSection is! YamlList) return false;
```
bryanoltman: These should be checking for null rather than using `is!`

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
```diff
@@ -0,0 +1,168 @@
...
+          ShorebirdYAMLAssetValidator().validate,
+        );
+        expect(results, hasLength(1));
+        expect(
+          results.first,
+          equals(
+            ValidationIssue(
+              severity: ValidationIssueSeverity.error,
+              message:
+                  '${p.join('pubspec.yaml')} is missing shorebird.yaml in assets',
```
bryanoltman: I don't think the `p.join` is necessary here with only one item.

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
```diff
@@ -0,0 +1,168 @@
...
+        var results = await runWithOverrides(
+          ShorebirdYAMLAssetValidator().validate,
+        );
+        expect(results, hasLength(1));
+        expect(results.first.fix, isNotNull);
+        await runWithOverrides(() => results.first.fix!());
+        results = await runWithOverrides(
+          ShorebirdYAMLAssetValidator().validate,
+        );
+        expect(results, isEmpty);
```
bryanoltman: This should validate the "after" state of the pubspec file to make sure it actually made the expected change

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
```diff
@@ -0,0 +1,168 @@
...
+      });
```
bryanoltman: Same here re: validating that the pubspec file was updated.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    final pubspecContent = pubspecYamlFile.readAsStringSync();
+    final pubspecYaml = loadYaml(pubspecContent) as YamlMap;
+
+    if (!_pubspecYamlHasShorebirdAsset(pubspecYaml)) {
+      return [
+        ValidationIssue(
+          severity: ValidationIssueSeverity.error,
+          message: '$pubspecYamlPath is missing shorebird.yaml in assets',
```
felangel:
```suggestion
          message: 'No shorebird.yaml found in pubspec.yaml assets.',
```

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    return [];
+  }
+
+  File? get _pubspecYamlFile {
+    final root = shorebirdEnv.getFlutterProjectRoot();
+    if (root == null) return null;
+    return File(p.join(root.path, 'pubspec.yaml'));
+  }
```
felangel: Can't we use `shorebirdEnv.getPubspecYaml` instead? It'll handle finding and parsing the pubspec contents.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,108 @@
...
+    final flutterSection = pubspecYaml['flutter'];
+    if (flutterSection is! YamlMap) return false;
+
+    final assetsSection = flutterSection['assets'];
+    if (assetsSection is! YamlList) return false;
+
+    return assetsSection.any((asset) => asset == 'shorebird.yaml');
+  }
+
+  void _addShorebirdAssetToFile(File pubspecYamlFile) {
```
felangel: This looks like it's a duplicate of the `_addShorebirdYamlToProject` in `init_command.dart`. We should refactor so that the logic isn't duplicated.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,108 @@
...
+  @override
+  String get incorrectContextMessage => '''
+The ${_pubspecYamlFile?.path ?? 'pubspec.yaml'} file does not exist.
+The command you are running must be run within a Flutter app project.''';
+
+  @override
+  Future<List<ValidationIssue>> validate() async {
+    final pubspecYamlFile = _pubspecYamlFile;
+    if (pubspecYamlFile == null || !pubspecYamlFile.existsSync()) {
+      return [
```
felangel: Isn't this code unreachable since `canRunInCurrentContext` would return false if there's no `pubspec.yaml`?

felangel (changes requested):
Thanks for the contribution! I left some comments, let me know what you think.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
borjandev: Simplified the logic a bit in the latest commit

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
borjandev: I tried this but was somehow making the tests fail, I suspect it's related to the way that the tests write the mock pubspec.yaml file

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
borjandev: I reworked the logic a bit in the latest commits

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
borjandev: I found an even better function ` _addShorebirdYamlToPubspecAssets` from the `InitCommand` clsas. 

I made it static and public so that I can tap into that logic instead, it greatly reduced the code.

Is it ok to make that function static and public like that @felangel ?

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
borjandev: I think this still works because of `await runWithOverrides(() => results.first.fix!());` ? 

Since if I comment that line out, the test fails.

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
borjandev: I think this still works because of `await runWithOverrides(() => results.first.fix!());` ? 

Since if I comment that line out, the test fails.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
borjandev: I tried a different approach and I think it works now

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
borjandev: I reworked the logic, should be better now

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
borjandev: I reworked the logic, should be better now

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -315,7 +315,7 @@ app_id:
     return ShorebirdYaml(appId: appId);
   }
 
-  void _addShorebirdYamlToPubspecAssets(File pubspecFile) {
+  static void addShorebirdYamlToPubspecAssets(File pubspecFile) {
```
bryanoltman: If this is being used outside of this command, it should probably live somewhere else (possibly in ShorebirdEnv), and be referenced by this command and the new validator.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,46 @@
...
+    if (!canRunInCurrentContext()) {
+      return [
+        const ValidationIssue(
+          severity: ValidationIssueSeverity.error,
+          message: 'No pubspec.yaml file found',
+        ),
+      ];
+    }
+
+    if (!shorebirdEnv.pubspecContainsShorebirdYaml) {
```
bryanoltman: This could be an early return (i.e., `if (shorebirdEnv.pubspecContainsShorebirdYaml) return [];`)

bryanoltman (commented):
A couple minor comments, but otherwise looking good!

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -210,7 +209,7 @@ Please make sure you are running "shorebird init" from within your Flutter proje
     );
 
     if (!shorebirdEnv.pubspecContainsShorebirdYaml) {
-      _addShorebirdYamlToPubspecAssets(
+      ShorebirdEnv.addShorebirdYamlToPubspecAssets(
```
bryanoltman: It doesn't need to be static – an instance method on shorebirdev would mean that we don't need to pass the pubspec file as an argument

borjandev: @bryanoltman in that case I believe it would conflict with the existing way that the `InitCommand` tests are executed as well as the `ShorebirdYamlAssetValidator` tests since `shorebirdEnv = MockShorebirdEnv();` ?

Implementing your suggestion here would likely require us to do something like?

```
when(() => shorebirdEnv.addShorebirdYamlToPubspecAssets()).thenReturn(
   () {
        // The actual function logic?
    },
  );
```
Most of the other shorebirdEnv methods make more sense to be mocked out as above but this one has more logic in it, writes to a file, and has a void return type, meaning that the above usage of `.thenReturn` wouldn't even make sense in this case.

Making it static avoids these issues.

Do you have some ideas on how to address the issues above if we were to not make it static anymore?

bryanoltman: Yep! ShorebirdEnv is a dependency of both the InitCommand and ShorebirdYamlAssetValidator. For unit tests, we assume dependencies do what they say. That is, a unit test for the InitCommand should only be testing the code in the InitCommand, and can assume that, as long as `shorebirdEnv. addShorebirdYamlToPubspecAssets()` was called, it's doing the right thing. We would ensure that `shorebirdEnv. addShorebirdYamlToPubspecAssets()` does what it should in the ShorebirdEnv unit tests.

borjandev: I took a stab at it @bryanoltman 

Is something like this what you had in mind? 

Might still need one more ShorebirdEnv unit test with this approach, but I want to confirm the direction first

packages/shorebird_cli/lib/src/shorebird_env.dart
```diff
@@ -242,7 +242,10 @@ class ShorebirdEnv {
       platform.environment.containsKey('TF_BUILD');
 
   /// Adds `shorebird.yaml` to the `assets` section of the `pubspec.yaml` file
-  static void addShorebirdYamlToPubspecAssets(File pubspecFile) {
+  void get addShorebirdYamlToPubspecAssets {
```
bryanoltman: I'm not sure why this is a getter – it has side effects and should be a function

borjandev: Changed it to a function

packages/shorebird_cli/lib/src/shorebird_env.dart
```diff
@@ -238,4 +240,35 @@ class ShorebirdEnv {
       // https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
       ||
       platform.environment.containsKey('TF_BUILD');
+
+  /// Adds `shorebird.yaml` to the `assets` section of the `pubspec.yaml` file
```
bryanoltman:
```suggestion
  /// Adds shorebird.yaml to the assets section of the pubspec.yaml file.
```

packages/shorebird_cli/test/src/commands/init_command_test.dart
bryanoltman, line 1: I think you'll probably need to move the tests that validate the pubspec changes to the ShorebirdEnv tests and just verify that `addShorebirdYamlToPubspecAssets` is called from the init command.

borjandev: I like this approach, tried it in the latest changes

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
       });
     });
+
+    group('addShorebirdYamlToPubspecAssets', () {
+      const pubspecContents = '''
+name: test
+version: 1.0.0
+environment:
+ sdk: ">=2.19.0 <3.0.0"''';
+      test('creates flutter.assets and adds shorebird.yaml', () {
```
bryanoltman:
```suggestion
 sdk: ">=2.19.0 <3.0.0"''';
 
      test('creates flutter.assets and adds shorebird.yaml', () {
```

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+          pubspecFile.readAsStringSync(),
+          equals('''
+$pubspecContents
+flutter:
+ assets:
+   - shorebird.yaml
+'''),
+        );
+      });
+      test('creates assets and adds shorebird.yaml (empty)', () {
```
bryanoltman:
```suggestion
      });

      test('creates assets and adds shorebird.yaml when flutter entry is empty', () {
```

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+          pubspecFile.readAsStringSync(),
+          equals('''
+$pubspecContents
+flutter:
+ assets:
+   - shorebird.yaml
+'''),
+        );
+      });
+      test('creates assets and adds shorebird.yaml (non-empty)', () {
```
bryanoltman:
```suggestion
      });

      test('creates assets and adds shorebird.yaml when flutter entry is present', () {
```

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+          equals('''
+$pubspecContents
+flutter:
+ assets:
+  - shorebird.yaml
+ uses-material-design: true
+'''),
+        );
+      });
+      test('adds shorebird.yaml to assets', () {
```
bryanoltman:
```suggestion
      });

      test('adds shorebird.yaml to assets', () {
```

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+          ..createSync()
+          ..writeAsStringSync('''
+$pubspecContents
+flutter:
+ assets:
+  - some/asset.txt
+''');
+        IOOverrides.runZoned(
+          () => runWithOverrides(
+            () => shorebirdEnv.addShorebirdYamlToPubspecAssets(),
```
bryanoltman: I think this can be a tear-off (ex: `runWithOverrides(shorebirdEnv.addShorebirdYamlToPubspecAssets)`)

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+        final pubspecFile = File(p.join(tempDir.path, 'pubspec.yaml'))
+          ..createSync()
+          ..writeAsStringSync('''
+$pubspecContents
+flutter:
+ uses-material-design: true
+''');
+        IOOverrides.runZoned(
+          () => runWithOverrides(
+            () => shorebirdEnv.addShorebirdYamlToPubspecAssets(),
```
bryanoltman: I think this can be a tear-off (ex: `runWithOverrides(shorebirdEnv.addShorebirdYamlToPubspecAssets)`)

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+        final tempDir = Directory.systemTemp.createTempSync();
+        final pubspecFile = File(p.join(tempDir.path, 'pubspec.yaml'))
+          ..createSync()
+          ..writeAsStringSync('''
+$pubspecContents
+flutter:
+''');
+        IOOverrides.runZoned(
+          () => runWithOverrides(
+            () => shorebirdEnv.addShorebirdYamlToPubspecAssets(),
```
bryanoltman: I think this can be a tear-off (ex: `runWithOverrides(shorebirdEnv.addShorebirdYamlToPubspecAssets)`)

packages/shorebird_cli/test/src/shorebird_env_test.dart
```diff
@@ -717,5 +717,111 @@ base_url: https://example.com''');
...
+environment:
+ sdk: ">=2.19.0 <3.0.0"''';
+      test('creates flutter.assets and adds shorebird.yaml', () {
+        final tempDir = Directory.systemTemp.createTempSync();
+        final pubspecFile = File(p.join(tempDir.path, 'pubspec.yaml'))
+          ..createSync()
+          ..writeAsStringSync(pubspecContents);
+        IOOverrides.runZoned(
+          () => runWithOverrides(
+            () => shorebirdEnv.addShorebirdYamlToPubspecAssets(),
```
bryanoltman: I think this can be a tear-off (ex: `runWithOverrides(shorebirdEnv.addShorebirdYamlToPubspecAssets)`)

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
```diff
@@ -0,0 +1,213 @@
...
+      writePubspecToPath(pubspecWithoutShorebirdAsset, projectRoot.path);
+
+      when(() => shorebirdEnv.getFlutterProjectRoot()).thenReturn(projectRoot);
+      when(() => shorebirdEnv.getPubspecYamlFile(cwd: any(named: 'cwd')))
+          .thenReturn(pubspecYamlFile);
+    });
+
+    setUpAll(() {
+      registerFallbackValue(MockDirectory());
+    });
```
bryanoltman: We usually put `setUpAll` before `setUp`

packages/shorebird_cli/test/src/validators/shorebird_yaml_asset_validator_test.dart
```diff
@@ -0,0 +1,213 @@
...
+      'returns successful result if pubspec.yaml has shorebird.yaml in assets',
+      () async {
+        when(() => shorebirdEnv.hasPubspecYaml).thenReturn(true);
+        when(() => shorebirdEnv.pubspecContainsShorebirdYaml).thenReturn(true);
+        final results = await runWithOverrides(
+          ShorebirdYamlAssetValidator().validate,
+        );
+        expect(results.map((res) => res.severity), isEmpty);
+      },
+    );
```
bryanoltman: This test and several below should be in a `validate` group

bryanoltman (approved):

packages/shorebird_cli/lib/src/shorebird_env.dart
```diff
@@ -238,4 +240,35 @@ class ShorebirdEnv {
       // https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml
       ||
       platform.environment.containsKey('TF_BUILD');
+
+  /// Adds shorebird.yaml to the assets section of the pubspec.yaml file.
+  void addShorebirdYamlToPubspecAssets() {
```
felangel: Minor but this doesn't feel like it belongs in `ShorebirdEnv`. I think it would make more sense to split this out into it's own class (e.g `PubspecEditor`) since `ShorebirdEnv` should mainly be a way to query information about the current shorebird environment. This is a mutation operation which can depend on the `ShorebirdEnv` but doesn't really belong on `ShorebirdEnv` imo.

packages/shorebird_cli/lib/src/validators/shorebird_yaml_asset_validator.dart
```diff
@@ -0,0 +1,40 @@
...
+  @override
+  Future<List<ValidationIssue>> validate() async {
+    if (!canRunInCurrentContext()) {
+      return [
+        const ValidationIssue(
+          severity: ValidationIssueSeverity.error,
+          message: 'No pubspec.yaml file found',
+        ),
+      ];
+    }
```
felangel, line 27: Minor but @bryanoltman shouldn't this be baked into `Validator`? Seems unnecessary to have to manually throw if validate is called when `canRunInCurrentContext` returns false. Wdyt?

bryanoltman: Good call. Part of the contract for Validators (maybe it could be more explicit) is that `canRunInCurrentContext` will be respected by the Validator owners

bryanoltman (approved):

felangel (approved):
LGTM thanks so much for the contribution! 💙

packages/shorebird_cli/lib/src/shorebird_env.dart
felangel: Refactored 👍

felangel (changes requested):
Going to clean up one more test before merging.

felangel (approved):

PR #1920, opened 2024-04-17
feat: add filename to CreateReleaseArtifactRequest

bryanoltman:
## Description

Sends a filename in the `CreateReleaseArtifactRequest`, which the server will use when writing to Google Cloud Storage. This will allow users to download release artifacts from the console with the correct extension (they are currently all have the `.vmcode` extension).

Part of https://github.com/shorebirdtech/_console/issues/158

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/messages/create_release_artifact/create_release_artifact_request.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/create_release_artifact/create_release_artifact_request.dart
@@ -18,6 +18,7 @@ class CreateReleaseArtifactRequest {
     required this.hash,
     required this.size,
     required this.canSideload,
+    required this.filename,
   });
 
   /// Converts a Map<String, dynamic> to a [CreateReleaseArtifactRequest]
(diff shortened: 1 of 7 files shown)


felangel:
We'll have to be careful about how this affects the rest of the system because the patch check endpoint (just as an example) returns a download url which includes the file name (.vmcode)

packages/shorebird_code_push_protocol/lib/src/messages/create_release_artifact/create_release_artifact_request.dart
```diff
@@ -36,6 +37,10 @@ class CreateReleaseArtifactRequest {
   /// The hash of the artifact.
   final String hash;
 
+  /// The name of the file.
+  /// TODO(NEXT_API): make this non-nullable.
```
felangel:
```suggestion
  /// TODO(bryanoltman): make this non-nullable in the next major release.
```

felangel (approved):
The change on the CLI looks fine but I think we need to be very careful before rolling it out on the backend because we'll need to fix several sites that relied on the previous file names.

bryanoltman:
> we need to be very careful before rolling it out on the backend because we'll need to fix several sites that relied on the previous file names

Searching for `vmcode` in the _shorebird repo (including the shorebird submodule), I'm only seeing this for patch artifacts, which are untouched by this change. LMK if I'm missing a site.

packages/shorebird_code_push_protocol/lib/src/messages/create_release_artifact/create_release_artifact_request.dart
bryanoltman: Since we've got an impending version bump, I'm just going to go ahead and make this non-optional with the assumption that we won't release a version between now and the update.

PR #1943, opened 2024-04-19
fix(shorebird_cli): fix release android target_platform with local_engine

bryanoltman:
## Description

Because the default value for `shorebird release android`'s `target-platform` option is `android-arm,android-arm64,android-x64` and because a local engine build targets only one of these, this PR updates the command to only use the target-platform args included in `AndroidArch.availableAndroidArchs`. An alternative solution would be to require `--target-platform` to be specified for local engine builds.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:collection/collection.dart';
 import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:scoped/scoped.dart';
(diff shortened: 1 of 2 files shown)


eseidel (approved):
I don't really understand the change, but so long as it's not breaking users, I'm not at all concerned about us breaking local-engine (since we can always fix that for ourselves easily).

felangel (approved):

bryanoltman:
Closing this in favor of requiring a target platform being specified for local engine builds (e.g., `shorebird release android --local-engine-src-path=$HOME/shorebirdtech/engine/src --local-engine=android_release_arm64 --local-engine-host=host_release --target-platform=android-arm64`)

PR #1952, opened 2024-04-22
chore(shorebird_cli): use upstream googleapis_auth package

### LINKED ISSUE #1944: fix: Because shorebird_cli depends on googleapis_auth from git which doesn't exist
**Description**

A clear and concise description of what the bug is when install shorebird in windows 10 by 

```
Set-ExecutionPolicy RemoteSigned -scope CurrentUser # Needed to execute remote scripts
iwr -UseBasicParsing 'https://raw.githubusercontent.com/shorebirdtech/install/main/install.ps1'|iex
```

Got error 

Because shorebird_cli depends on googleapis_auth from git which doesn't exist
  (Could not find git ref '647f96d5ffdae63e41aeea091c34acfffff533e7' (fatal: bad
  object 647f96d5ffdae63e41aeea091c34acfffff533e7)),

**Steps To Reproduce**

**Expected Behavior**

**Screenshots**

image

**Additional Context**

bryanoltman:
## Description

Fixes https://github.com/shorebirdtech/shorebird/issues/1944

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/pubspec.yaml
+++ b/packages/shorebird_cli/pubspec.yaml
@@ -26,11 +26,7 @@ dependencies:
   cli_util: ^0.4.1
   collection: ^1.18.0
   crypto: ^3.0.3
-  googleapis_auth:
-    git:
-      url: https://github.com/shorebirdtech/googleapis.dart.git
-      path: googleapis_auth
-      ref: 647f96d5ffdae63e41aeea091c34acfffff533e7
+  googleapis_auth: ^1.6.0
   http: ^1.2.1
   intl: ^0.19.0
   io: ^1.0.4
--- a/packages/shorebird_cli/pubspec.lock
+++ b/packages/shorebird_cli/pubspec.lock
@@ -286,19 +286,18 @@ packages:
     dependency: transitive
     description:
       name: google_identity_services_web
-      sha256: "000b7a31e1fa17ee04b6c0553a2b2ea18f9f9352e4dcc0c9fcc785cf10f2484e"
+      sha256: "9482364c9f8b7bd36902572ebc3a7c2b5c8ee57a9c93e6eb5099c1a9ec5265d8"
       url: "https://pub.dev"
     source: hosted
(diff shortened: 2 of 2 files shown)


eseidel:
Does this bump our minimum Dart version?

felangel (approved):

packages/shorebird_cli/pubspec.yaml
```diff
@@ -26,11 +26,7 @@ dependencies:
   cli_util: ^0.4.1
   collection: ^1.18.0
   crypto: ^3.0.3
-  googleapis_auth:
-    git:
-      url: https://github.com/shorebirdtech/googleapis.dart.git
-      path: googleapis_auth
-      ref: 647f96d5ffdae63e41aeea091c34acfffff533e7
+  googleapis_auth: ^1.6.0
```
felangel, line 29: Don't we also need to bump the min dart sdk version?

felangel: nevermind just checked and it looks like googleapis_auth min is 3.0.0 so we're good 👍

bryanoltman: I think there may a dependency of that project that requires a higher version. Checking with an older Dart version now.

bryanoltman: It looks like several of our dependencies require newer versions of Dart than our pubspec currently says it requires. `http 1.2.1` requires Dart 3.3.0, for example.

eseidel:
We already pull our own Dart.  I'm not sure what call we're making that's not using it:
https://github.com/shorebirdtech/shorebird/blob/main/bin/shorebird
https://github.com/shorebirdtech/shorebird/blob/main/third_party/flutter/bin/internal/shared.sh

eseidel:
Yeah, we have our own Dart, we should just fix cases where we're using the wrong one.

PR #1958, opened 2024-04-23
fix(shorebird_cli): update cache artifacts on every command run, log progress

### LINKED ISSUE #1899: fix: `detecting release version` hangs for a long time (downloading assets)
This has confused many customers.  We don't explain that we're downloading Flutter assets.

One fix might be to just run `flutter precache` in its own separate step.

See 
https://discord.com/channels/1030243211995791380/1229355551868321874
as on example.

bryanoltman:
## Description

Updates `shorebird_cli` to ensure all cache artifacts are up-to-date before running any command and prints the update progress, if any. This ensures:

1. We will always know that cache artifacts are up-to-date in any command.
2. Cache artifacts are updated in a consistent and predictable place.
3. Users have better visibility into what the CLI is doing (see the fixed issue).

Fixes https://github.com/shorebirdtech/shorebird/issues/1899

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -58,9 +58,9 @@ class Cache {
         continue;
       }
 
-      final progress = logger.progress('Downloading ${artifact.name}');
+      final progress = logger.progress('Downloading ${artifact.name} artifact');
       await artifact.update();
-      progress.complete('Downloaded ${artifact.name}');
+      progress.complete('Downloaded ${artifact.name} artifact');
     }
   }
 
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -1,6 +1,7 @@
 import 'dart:io' hide Platform;
 
 import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 import 'package:platform/platform.dart';
 import 'package:scoped/scoped.dart';
@@ -49,15 +50,17 @@ class Cache {
     registerArtifact(AotToolsExeArtifact(cache: this, platform: platform));
   }
 
-  void registerArtifact(CachedArtifact artifact) => _artifacts.add(artifact);
+  void registerArtifact(CachedArtifact artifact) => artifacts.add(artifact);
 
   Future<void> updateAll() async {
-    for (final artifact in _artifacts) {
+    for (final artifact in artifacts) {
       if (await artifact.isUpToDate()) {
         continue;
       }
 
+      final progress = logger.progress('Downloading ${artifact.name}');
       await artifact.update();
+      progress.complete('Downloaded ${artifact.name}');
     }
   }
 
@@ -98,7 +101,8 @@ class Cache {
     );
   }
 
-  final List<CachedArtifact> _artifacts = [];
+  @visibleForTesting
+  final List<CachedArtifact> artifacts = [];
 
   String get storageBaseUrl => 'https://storage.googleapis.com';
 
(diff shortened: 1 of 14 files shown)


third_party/flutter/bin/internal/shared.sh
```diff
@@ -164,8 +164,12 @@ function upgrade_shorebird () (
       mv "$SNAPSHOT_PATH" "$SNAPSHOT_PATH_OLD"
     fi
 
-    # Compile...
-    $DART_PATH --verbosity=error --disable-dart-dev --snapshot="$SNAPSHOT_PATH" --snapshot-kind="app-jit" --packages="$SHOREBIRD_CLI_DIR/.dart_tool/package_config.json" --no-enable-mirrors "$SCRIPT_PATH" > /dev/null
+    # Compile our snapshot.
+    # We invoke `$SNAPSHOT_PATH completion` to trigger the "completion" command, which
+    # avoids executing as much of our code as possible. We do this because running
+    # the script here (instead of from the compiled snapshot) invalidates a lot of
+    # assumptions we make about the cwd in the shorebird_cli tool.
```
eseidel, line 171: Is there no way to compile w/o running?

bryanoltman: There quite possibly is, but adding the `completion` argument here seemed like a much smaller and safer change than updating how we compile our tool. Happy to revisit if you feel it's worth further investigation.

bryanoltman: (the logic here is pulled directly from https://github.com/flutter/flutter/blob/master/bin/internal/shared.sh)

felangel (approved):

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -207,6 +211,14 @@ class AotToolsDillArtifact extends CachedArtifact {
...
 /// For a few revisions in Dec 2023, we distributed aot-tools as an executable.
 /// Should be removed sometime after June 2024.
+///
+/// The change to use a .dill file was made in
+/// https://github.com/shorebirdtech/_build_engine/commit/babbc37d93e7a2f36e62787e47eee5a3b5458901
+/// The Flutter versions that use this are:
+///  - 3.13.9 (a3d5f7c614aa1cc4d6cb1506e74fd1c81678e68e)
+///  - 3.16.3 (b9b23902966504a9778f4c07e3a3487fa84dcb2a)
+///  - 3.16.4 (7e92b034c5dddb727cf5e802c23cddd39b325a7f)
+///  - 3.16.5 (4e8a7c746ae6f10951f3e676f10b82b21d7300a5)
```
eseidel, line 221: Future me says thank you. :)

third_party/flutter/bin/internal/shared.sh
eseidel: sg, thanks.

packages/shorebird_cli/lib/src/command_runner.dart
```diff
@@ -206,6 +207,9 @@ Run ${lightCyan.wrap('shorebird upgrade')} to upgrade.''');
       exitCode = ExitCode.success.code;
     } else {
       try {
+        // Ensure all cached artifacts are up-to-date before running the
+        // command.
```
eseidel, line 215:
```suggestion
        // Most commands need shorebird artifacts, so we ensure the cache
        // is up to date for all commands (even ones which don't need it).  We
        // could make a Command subclass and move the cache onto that and
        // only update in that case, but it didn't seem worth it at the time.
        // Ensure all cached artifacts are up-to-date before running the
        // command.
```

eseidel (approved):

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: 🙏

PR #1960, opened 2024-04-23
fix(shorebird_cli): handle multi-dimensional flavors on android

### LINKED ISSUE #1830: fix: flavors with multiple dimensions cause release command to not find aab
With flavors across multiple dimensions, our release command cannot find the generated aab.

For example:

```gradle
    flavorDimensions "track", "country"
    productFlavors {
        internal {
            dimension "track"
            applicationIdSuffix ".internal"
            manifestPlaceholders = [applicationLabel: "[Internal] Shorebird Example"]
        }
        stable {
            dimension "track"
            manifestPlaceholders = [applicationLabel: "Shorebird Example"]
        }
        global {
            applicationIdSuffix ".gl"
            dimension "country"
            manifestPlaceholders = [applicationLabel: "Shorebird Example"]
        }
        local {
            applicationIdSuffix ".lcl"
            dimension "country"
            manifestPlaceholders = [applicationLabel: "Shorebird Example"]
        }
    }
```

With this, running `shorebird release android --flavor stableGlobal` generates an aab at `build/app/outputs/bundle/stableGlobalRelease/app-stable-global-release.aab`, but our tool fails with:

```
✗ Exception: Failed to extract version name from app bundle: [BT:1.15.6] Error: File '/Users/bryanoltman/shorebirdtech/samples/flavors/build/app/outputs/bundle/stableGlobalRelease/app-stableGlobal-release.aab' was not found.
java.lang.IllegalArgumentException: File '/Users/bryanoltman/shorebirdtech/samples/flavors/build/app/outputs/bundle/stableGlobalRelease/app-stableGlobal-release.aab' was not found.
```

Changing `stable` and `global` to `stableFlavor` and `globalFlavor` yields:

`✗ Failed to build: Gradle build failed to produce an .aab file. It's likely that this file was generated under /Users/bryanoltman/shorebirdtech/samples/flavors/build, but the tool couldn't find it.`

The output:

```
⑆ l /Users/bryanoltman/shorebirdtech/samples/flavors/build/app/outputs/bundle/stableFlavorGlobalFlavorRelease
total 50104
drwxr-xr-x@ 3 bryanoltman  staff    96B Apr 22 12:09 .
drwxr-xr-x@ 7 bryanoltman  staff   224B Apr 22 12:09 ..
-rw-r--r--@ 1 bryanoltman  staff    24M Apr 22 12:09 app-stableFlavor-globalFlavor-release.aab
```

_Originally posted by @AbhijithKonnayil in https://github.com/shorebirdtech/shorebird/issues/515#issuecomment-2022555249_

erickzanardo:
## Description

Our CLI would get confused when using multi dimensional flavors on android. This PR fixes that.

Fixes #1830 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -195,36 +195,31 @@ Use `shorebird flutter versions list` to list available versions.
...
 
-        final bundleDirPath = p.join(
-          projectRoot.path,
-          'build',
-          'app',
-          'outputs',
-          'bundle',
-        );
+        String? bundlePath;
+        String? apkPath;
```
eseidel: Do these need to be nullable?

Dart supports fancy:
```
final Foo foo;
try {
   foo = thing();
} catch () {
   logger.info('failed');
   return;
}
```
It will do flow analysis to know that the finals are set during the try or not used after.

eseidel: Ah, nevermind. I see they can get explicitly set to null.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
+import 'package:path/path.dart' as p;
+import 'package:scoped/scoped.dart';
+import 'package:shorebird_cli/src/command.dart';
+import 'package:shorebird_cli/src/third_party/flutter_tools/lib/flutter_tools.dart';
+
+/// Throw when multiple artifacts are found in the build directory.
```
eseidel: Will that happen when using Android Studio and building different flavors/configs between multiple runs?  I wonder if older .aab's from other configs being in the build dir can cause these failures to happen?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+  @override
+  String toString() {
+    return 'Multiple artifacts found in $buildDir: '
+        '${foundArtifacts.map((e) => e.path)}';
+  }
+}
+
+extension on String {
+  String get artifactId => replaceAll(RegExp(r'\W'), '').toLowerCase();
+}
```
eseidel, line 66: I believe this makes it anonymous and thus local to this file, correct?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+
+final shorebirdAndroidArtifactsRef = create(ShorebirdAndroidArtifacts.new);
+
+ShorebirdAndroidArtifacts get shorebirdAndroidArtifacts =>
+    read(shorebirdAndroidArtifactsRef);
```
eseidel, line 71: @felangel is our style to use globals for things like this?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+
+/// Mixin on [ShorebirdCommand] which exposes methods
+// to find the artifacts generated for android
+class ShorebirdAndroidArtifacts {
+  /// Find the artifact in the build directory.
+  String? findArtifact({
+    required String artifactName,
+    required String buildDir,
```
eseidel: Should this be a Directory type?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+
+ShorebirdAndroidArtifacts get shorebirdAndroidArtifacts =>
+    read(shorebirdAndroidArtifactsRef);
+
+/// Mixin on [ShorebirdCommand] which exposes methods
+// to find the artifacts generated for android
+class ShorebirdAndroidArtifacts {
+  /// Find the artifact in the build directory.
+  String? findArtifact({
+    required String artifactName,
```
eseidel, line 78: Should this be an enum or File?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+
+    final allFiles = directory.listSync();
+    final candidates = allFiles.where((file) {
+      final fileName = p.basename(file.path);
+      return fileName.artifactId == artifactId;
+    }).toList();
+
+    if (candidates.isEmpty) {
+      return null;
+    } else if (candidates.length > 1) {
```
eseidel: No need for else after return.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+      throw MultipleArtifactsFoundException(
+        buildDir: buildDir,
+        foundArtifacts: candidates,
+      );
+    } else {
```
eseidel: No need for else after return.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+      return candidates.first.path;
+    }
+  }
+
+  /// Find the app bundle in the build directory.
+  String? findAppBundle({
+    required String projectPath,
```
eseidel: Should this be a File?  (I don't know what the CLI style is.  I recall @bryanoltman mentioning he tried to use File instead of String for most paths?)

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
felangel: We can still have the top-level variables be non-nullable and only set the in the try block when we have successfully determined a bundlePath and apkPath.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
felangel:
```suggestion
/// Thrown when multiple artifacts are found in the build directory.
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
felangel: For scoped instances, yes. This allows us to use a single instance of `ShorebirdAndroidArtifacts` throughout while also being able to override it in tests for mocking.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
felangel: Can we change the signature of this method to return a non-nullable `File` and throw an ArtifactNotFound exception in the case where an artifact isn't found?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
felangel: Same thing here regarding changing the signature to return a non-nullable `File` and throwing if an artifact is not found.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: I should never happen as far I can tell, this was more of a "just to be sure" checking that Bryan suggested while we were discussing the solution.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: Exactly!

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+final shorebirdAndroidArtifactsRef = create(ShorebirdAndroidArtifacts.new);
+
+ShorebirdAndroidArtifacts get shorebirdAndroidArtifacts =>
+    read(shorebirdAndroidArtifactsRef);
+
+/// Mixin on [ShorebirdCommand] which exposes methods
+// to find the artifacts generated for android
+class ShorebirdAndroidArtifacts {
+  /// Find the artifact in the build directory.
+  String? findArtifact({
```
felangel: Should this be private so that it forces callers to use the `findAppBundle` and `findAppApk` APIs?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
felangel: It would be nice to add a comment here explaining what this is doing and why we're using this method to compute an artifact id.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+    final artifactName =
+        flavor == null ? 'app-release.aab' : 'app-$flavor-release.aab';
+
+    return findArtifact(
+      buildDir: buildDir,
+      artifactName: artifactName,
+    );
+  }
+
+  /// Find the apk in the build directory.
```
felangel:
```suggestion
  /// Find the apk in the provided [projectPath].
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
felangel: Yeah I'm also concerned about this case. Have we filed an issue upstream for this bug?

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
felangel: Better yet, see my comments about adjusting the signatures to return non-nullable types.

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,282 @@
...
+      expect(
+        shorebirdAndroidArtifacts.findApk(
+          projectPath: projectPath.path,
+          flavor: null,
+        ),
+        isNull,
+      );
+    });
+
+    test('find the app bundle without any flavors', () {
```
felangel:
```suggestion
    test('finds the app bundle without any flavors', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,282 @@
...
+    test('returns null when cannot find the aab', () {
+      expect(
+        shorebirdAndroidArtifacts.findAppBundle(
+          projectPath: projectPath.path,
+          flavor: null,
+        ),
+        isNull,
+      );
+    });
+    test('returns null when cannot find the apk', () {
```
felangel:
```suggestion
    });

    test('returns null when an apk does not exist', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,282 @@
...
+  group(ShorebirdAndroidArtifacts, () {
+    late Directory projectPath;
+    late ShorebirdAndroidArtifacts shorebirdAndroidArtifacts;
+
+    setUp(() {
+      projectPath = Directory.systemTemp.createTempSync();
+      shorebirdAndroidArtifacts = ShorebirdAndroidArtifacts();
+    });
+
+    test('returns null when cannot find the aab', () {
```
felangel:
```suggestion
    test('returns null when an aab does not exist', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,282 @@
...
+
+      expect(
+        shorebirdAndroidArtifacts.findAppBundle(
+          projectPath: projectPath.path,
+          flavor: null,
+        ),
+        equals(artifact.path),
+      );
+    });
+    test('find the apk bundle without any flavors', () {
```
felangel:
```suggestion
    });

    test('finds the apk without any flavors', () {
```

Also might be nice to group these tests like:

```dart
group('when apk exists', () {
  group('when apk contains flavors', () {
    
  });

  group('when apk does not contain flavors', () {

  });
});

group('when apk does not exist', () {

});
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: Added, lmk if it is comprehensible

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -13,22 +13,22 @@ void main() {
...
+        () => shorebirdAndroidArtifacts.findAppBundle(
           projectPath: projectPath.path,
           flavor: null,
         ),
-        isNull,
+        throwsA(isA<ArtifactNotFoundException>()),
       );
     });
-    test('returns null when cannot find the apk', () {
+    test('throws ArtifactNotFoundException when cannot find the apk', () {
```
bryanoltman:
```suggestion

    test('throws ArtifactNotFoundException when cannot find the apk', () {
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
bryanoltman: Did you push the commit that adds the comment? I'm not seeing it on my end. Something like "Removes all non-word characters (i.e., `[^a-zA-Z0-9_]`) from the string" would be great.

bryanoltman: Also mentioning the "why" is great too – I feel like we've had more cases of "why did we do that again?" this week than usual.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,110 @@
...
+    // Remove all non characters and digits from the artifact name.
+    final artifactId = artifactName.artifactId;
+
+    final directory = Directory(buildDir);
+    if (!directory.existsSync()) {
+      return null;
+    }
+
+    final allFiles = directory.listSync();
+    final candidates = allFiles.where((file) {
```
bryanoltman: Maybe `aabArtifactCandidates`?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
bryanoltman: Yep, if this represents a Directory, it's better to use the semantically richer type.

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -274,6 +304,19 @@ void main() {
...
+      //    projectPath: any(named: 'projectPath'),
+      //    flavor: any(named: 'flavor'),
+      //  ),
+      //).thenReturn('app-release.aab');
+      //when(
+      //  () => shorebirdAndroidArtifacts.findApk(
+      //    projectPath: any(named: 'projectPath'),
+      //    flavor: any(named: 'flavor'),
+      //  ),
+      //).thenReturn('app-release.apk');
```
erickzanardo:
```suggestion
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: @bryanoltman I pushed yes! I can see the comment in my PR, can you check again please?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+  final String buildDir;
+
+  @override
+  String toString() {
+    return 'Artifact $artifactName not found in $buildDir';
+  }
+}
+
+/// When building android artifacts, gradlew name these artifacts in a non
+/// consistent way.
```
bryanoltman:
```suggestion
/// When building android artifacts, gradlew name these artifacts in an
/// inconsistent way.
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+///  ...bundle/fullFlavorCamelCase/app-flavor1-flavor2-release.aab
+///
+/// The pattern follows for APKs.
+///
+/// The only thing that is consistent is the artifactId which is the name of the
+/// artifact will follow always the same order of name of the flavors.
+///
+/// So we create the idea of an identifier, which consistis in the file name
+/// without any special character and in lower case, that allows use to easily
+/// search in the build folder what was the artifact that was generated by the
```
bryanoltman:
```suggestion
/// To get around this, we create an identifier for the artifact
/// that is the file name lowercased and without any non-word characters.
/// This allows use to reliably find the artifact generated by the
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
bryanoltman: Yep, see it now. Added some suggestions 👍

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: Not yet. Was focusing on our PR first, will open now that it seems that I addressed all/most of the comments

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -195,43 +195,33 @@ Use `shorebird flutter versions list` to list available versions.
...
-        final apkPath = flavor != null
-            ? p.join(
-                apkDirPath,
-                flavor,
-                'release',
-                'app-$flavor-release.apk',
-              )
-            : p.join(apkDirPath, 'release', 'app-release.apk');
+        final File bundleFile;
+        final File apkFile;
```
felangel:
```suggestion
        final File apkFile, bundleFile;
```

nit 🤷

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+  final String artifactName;
+  final String buildDir;
+
+  @override
+  String toString() {
+    return 'Artifact $artifactName not found in $buildDir';
+  }
+}
+
+/// When building android artifacts, gradlew name these artifacts in an
```
felangel:
```suggestion
/// When building android artifacts, gradlew names these artifacts in an
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: This method handles both aabs and apks, so `aabArtifactCandidates` would be wrong.

Gonna change to `artifactCandidates`

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+///
+///  With multi dimensional flavors
+///  ...bundle/fullFlavorCamelCase/app-flavor1-flavor2-release.aab
+///
+/// The pattern follows for APKs.
+///
+/// The only thing that is consistent is the artifactId which is the name of the
+/// artifact will follow always the same order of name of the flavors.
+///
+/// To get around this, we create an identifier for the artifact
```
felangel, line 60: Is there an upstream bug in Flutter tracking this? If so, can we link to it and add a TODO to revisit this workaround when the upstream bug is fixed?

erickzanardo: I don't think that the upstream bug will make us no need this routine.

The bug that I gound in flutter tool is that if you try `flutter run --flavor developmentPlayStore`, whereas `PlayStore` is a flavor of its own, the tool will break the execution since it will not find the the apk.

If we look on the sourcecode of the flutter tool, we notice that they also do a routine it just don't include a logic to get that case, while the logic that we wrote on shorebirtd does.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: Filled: https://github.com/flutter/flutter/issues/147261

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -195,43 +195,32 @@ Use `shorebird flutter versions list` to list available versions.
...
-            : p.join(bundleDirPath, 'release', 'app-release.aab');
-        final apkPath = flavor != null
-            ? p.join(
-                apkDirPath,
-                flavor,
-                'release',
-                'app-$flavor-release.apk',
-              )
-            : p.join(apkDirPath, 'release', 'app-release.apk');
+        final File apkFile, bundleFile;
```
bryanoltman: I think we generally define one variable per line, even if they're of the same type.

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+/// that is the file name lowercased and without any non-word characters.
+/// This allows use to reliably find the artifact generated by the
```
bryanoltman:
```suggestion
/// This allows us to reliably find the artifact generated by the
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+      'flutter-apk',
+    );
+
+    final artifaceName =
+        flavor == null ? 'app-release.apk' : 'app-$flavor-release.apk';
+
+    return _findArtifact(
+      directory: Directory(buildDir),
+      artifactName: artifaceName,
+    );
```
bryanoltman:
```suggestion
    final artifactName =
        flavor == null ? 'app-release.apk' : 'app-$flavor-release.apk';

    return _findArtifact(
      directory: Directory(buildDir),
      artifactName: artifactName,
    );
```

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
...
+    required Directory directory,
+  }) {
+    // Remove all non characters and digits from the artifact name.
+    final artifactId = artifactName.artifactId;
+
+    if (!directory.existsSync()) {
+      throw ArtifactNotFoundException(
+        artifactName: artifactName,
+        buildDir: directory.path,
+      );
```
bryanoltman, line 88: Should we throw the same type of exception in both the "we got a bad directory" case and the "we didn't find any artifacts in a good directory" case? It could be that we don't care, but just checking in case we do.

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,71 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
       expect(exitCode, equals(ExitCode.success.code));
       verifyNever(() => logger.confirm(any()));
     });
+
+    test('error when the app bundle cannot be found ', () async {
```
bryanoltman:
```suggestion
    test('error when the app bundle cannot be found', () async {
```

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,71 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
...
+        ),
+      );
+      final exitCode = await runWithOverrides(command.run);
+      verify(
+        () => logger.err('Artifact app-release.aab not found in buildDir'),
+      ).called(1);
+      expect(exitCode, ExitCode.software.code);
+    });
+
+    test('error when the apk cannot be found ', () async {
```
bryanoltman:
```suggestion
    test('error when the apk cannot be found', () async {
```

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,71 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
...
+      shorebirdAndroidArtifacts = MockShorebirdAndroidArtifacts();
+      when(
+        () => shorebirdAndroidArtifacts.findAppBundle(
+          project: any(named: 'project'),
+          flavor: any(named: 'flavor'),
+        ),
+      ).thenReturn(File('app-release.aab'));
```
bryanoltman: Is this override here because we try to find an aab before looking for an apk?

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,71 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
...
+        ),
+      );
+      final exitCode = await runWithOverrides(command.run);
+      verify(
+        () => logger.err('Artifact app-release.apk not found in buildDir'),
+      ).called(1);
+      expect(exitCode, ExitCode.software.code);
+    });
+
+    test('error when multiple artifacts are found ', () async {
```
bryanoltman:
```suggestion
    test('error when multiple artifacts are found', () async {
```

packages/shorebird_cli/test/src/mocks.dart
```diff
@@ -149,3 +150,6 @@ class MockStdin extends Mock implements Stdin {}
 class MockValidator extends Mock implements Validator {}
 
 class MockXcodeBuild extends Mock implements XcodeBuild {}
+
+class MockShorebirdAndroidArtifacts extends Mock
+    implements ShorebirdAndroidArtifacts {}
```
bryanoltman: The Mocks in this file are alphabetical, this should be higher up

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          expect(
+            () => shorebirdAndroidArtifacts.findApk(
+              project: project,
+              flavor: null,
+            ),
+            throwsA(isA<ArtifactNotFoundException>()),
+          );
+        });
+      });
+      group('when build folder exists but not the file', () {
```
bryanoltman:
```suggestion
      });

      group('when build folder exists but not the file', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          ).createSync(recursive: true);
+          expect(
+            () => shorebirdAndroidArtifacts.findAppBundle(
+              project: project,
+              flavor: null,
+            ),
+            throwsA(isA<ArtifactNotFoundException>()),
+          );
+        });
+        test('throws ArtifactNotFoundException for apks', () {
```
bryanoltman:
```suggestion
        });

        test('throws ArtifactNotFoundException for apks', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          shorebirdAndroidArtifacts
+              .findAppBundle(
+                project: project,
+                flavor: null,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+      test('finds the apk', () {
```
bryanoltman:
```suggestion
      });

      test('finds the apk', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+              .findAppBundle(
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+      test('finds the apk', () {
+        final artifactPath = path.join(
```
bryanoltman:
```suggestion
      });

      test('finds the apk', () {
        final artifactPath = path.join(
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          shorebirdAndroidArtifacts
+              .findAppBundle(
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+      test('finds the apk with multi dimensional flavors', () {
```
bryanoltman:
```suggestion
      });

      test('finds the apk with multi dimensional flavors', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          shorebirdAndroidArtifacts
+              .findAppBundle(
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+      test('finds the apk', () {
```
bryanoltman:
```suggestion
      });

      test('finds the apk', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+
+        expect(
+          () => shorebirdAndroidArtifacts.findAppBundle(
+            project: project,
+            flavor: flavor,
+          ),
+          throwsA(isA<MultipleArtifactsFoundException>()),
+        );
+      });
+      test('throws when finding multiple apk files', () {
```
bryanoltman:
```suggestion
      });

      test('throws when finding multiple apk files', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          'build',
+          'app',
+          'outputs',
+          'bundle',
+          'internalRelease',
+          'app-internal-release.aab',
+        );
+        final artifact = File(
+          path.join(project.path, artifactPath),
+        )..createSync(recursive: true);
```
bryanoltman, line 140: This could be in the group's setUp

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+        test('throws ArtifactNotFoundException for aabs', () {
+          expect(
+            () => shorebirdAndroidArtifacts.findAppBundle(
+              project: project,
+              flavor: null,
+            ),
+            throwsA(isA<ArtifactNotFoundException>()),
+          );
+        });
+        test('throws ArtifactNotFoundException for apks', () {
```
bryanoltman:
```suggestion
        });

        test('throws ArtifactNotFoundException for apks', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          Directory(
+            path.join(
+              project.path,
+              'build',
+              'app',
+              'outputs',
+              'bundle',
+              'release',
+            ),
+          ).createSync(recursive: true);
```
bryanoltman: This could be in the group's setUp (instead of in the test itself)

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,347 @@
...
+          'build',
+          'app',
+          'outputs',
+          'bundle',
+          'release',
+          'app-release.aab',
+        );
+        final artifact = File(
+          path.join(project.path, artifactPath),
+        )..createSync(recursive: true);
```
bryanoltman, line 91: Same here re: being in a setUp function instead of in the test

bryanoltman (approved):
A bunch of small stylistic comments, but overall LGTM

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
felangel: My fault, I suggested a one liner. I'm happy either way we should probably enforce one or the other via lint rules.

erickzanardo: Agree with the lint rule 😅

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: oh, big typo, good catch

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
erickzanardo: Exactly, so the `findAppBundle` cannot fail, otherwise the test will not fail because of the reason that we want it to fail

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
erickzanardo: Screenshot 2024-04-23 at 19 11 44

the lint actually complain, I wonder why the CI didn't catch this?

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: I feel it is not worth differentiating since the action is the same.

Unless we can think in any different action for each case?

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
erickzanardo: I am unsure, each of the tests uses a different path since one is about aabs and the other about apks, which lives in different places. So I fail to see the benefit.

Or maybe I am missing something?

packages/shorebird_cli/bin/shorebird.dart
```diff
@@ -61,6 +62,7 @@ Future<void> main(List<String> args) async {
         shorebirdValidatorRef,
         shorebirdVersionRef,
         xcodeBuildRef,
+        shorebirdAndroidArtifactsRef,
```
felangel: Can we keep this list sorted alphabetically?

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -195,43 +195,33 @@ Use `shorebird flutter versions list` to list available versions.
...
-                'app-$flavor-release.apk',
-              )
-            : p.join(apkDirPath, 'release', 'app-release.apk');
+        final File apkFile;
+        final File bundleFile;
+        try {
+          bundleFile = shorebirdAndroidArtifacts.findAppBundle(
+            project: projectRoot,
+            flavor: flavor,
+          );
```
felangel:
```suggestion
        final File apkFile;
        final File aabFile;
        try {
          aabFile = shorebirdAndroidArtifacts.findAab(
            project: projectRoot,
            flavor: flavor,
          );
```

Minor but for consistency might want to rename?

packages/shorebird_cli/bin/shorebird.dart
erickzanardo: I wish there were a linter that could check those things :*(

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -195,43 +195,33 @@ Use `shorebird flutter versions list` to list available versions.
...
+        final File aabFile;
+        try {
+          aabFile = shorebirdAndroidArtifacts.findAppBundle(
+            project: projectRoot,
+            flavor: flavor,
+          );
+          apkFile = shorebirdAndroidArtifacts.findApk(
+            project: projectRoot,
+            flavor: flavor,
+          );
```
bryanoltman: Missed this the first time, but this reads like it will always fail unless `generateApk` is true.

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
bryanoltman: 👍  I'd maybe leave a small comment so that's clear

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
bryanoltman: In this case, it's probably not necessary.

packages/shorebird_cli/lib/src/commands/release/release_android_command.dart
```diff
@@ -195,43 +195,35 @@ Use `shorebird flutter versions list` to list available versions.
...
-                apkDirPath,
-                flavor,
-                'release',
-                'app-$flavor-release.apk',
-              )
-            : p.join(apkDirPath, 'release', 'app-release.apk');
+        late final File apkFile;
+        final File aabFile;
+        try {
+          aabFile = shorebirdAndroidArtifacts.findAppBundle(
```
felangel:
```suggestion
          aabFile = shorebirdAndroidArtifacts.findAab(
```

minor but might want to rename for consistency

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
```diff
@@ -0,0 +1,157 @@
+import 'package:path/path.dart' as p;
+import 'package:scoped/scoped.dart';
+import 'package:shorebird_cli/src/command.dart';
+import 'package:shorebird_cli/src/third_party/flutter_tools/lib/flutter_tools.dart';
```
felangel: What is this import used for?

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,72 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
       expect(exitCode, equals(ExitCode.success.code));
       verifyNever(() => logger.confirm(any()));
     });
+
+    test('error when the app bundle cannot be found', () async {
```
felangel:
```suggestion
    test('errors when the app bundle cannot be found', () async {
```

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,72 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
...
+        ),
+      );
+      final exitCode = await runWithOverrides(command.run);
+      verify(
+        () => logger.err('Artifact app-release.aab not found in buildDir'),
+      ).called(1);
+      expect(exitCode, ExitCode.software.code);
+    });
+
+    test('error when the apk cannot be found', () async {
```
felangel:
```suggestion
    test('errors when the apk cannot be found', () async {
```

packages/shorebird_cli/test/src/commands/release/release_android_command_test.dart
```diff
@@ -888,5 +920,72 @@ Note: ${lightCyan.wrap('shorebird patch android --flavor=$flavor --target=$targe
...
+        ),
+      );
+      final exitCode = await runWithOverrides(command.run);
+      verify(
+        () => logger.err('Artifact app-release.apk not found in buildDir'),
+      ).called(1);
+      expect(exitCode, ExitCode.software.code);
+    });
+
+    test('error when multiple artifacts are found', () async {
```
felangel:
```suggestion
    test('errors when multiple artifacts are found', () async {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+  group(ShorebirdAndroidArtifacts, () {
+    late Directory project;
+    late ShorebirdAndroidArtifacts shorebirdAndroidArtifacts;
+
+    setUp(() {
+      project = Directory.systemTemp.createTempSync();
+      shorebirdAndroidArtifacts = ShorebirdAndroidArtifacts();
+    });
+
+    group('when cannot find artifacts', () {
```
felangel: Is this group necessary?

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+    });
+
+    group('when with multi dimensional flavors', () {
```
felangel:
```suggestion
    group('when using multi-dimensional flavors', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+                project: project,
+                flavor: null,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+    });
+
+    group('with one dimensional flavor', () {
```
felangel:
```suggestion
    group('when using single-dimensional flavor', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+              project: project,
+              flavor: null,
+            ),
+            throwsA(isA<ArtifactNotFoundException>()),
+          );
+        });
+      });
+    });
+
+    group('with no flavors', () {
```
felangel:
```suggestion
    group('when using no flavors', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+    });
+
+    group('when with multi dimensional flavors', () {
+      test('finds the app bundle with multi dimensional flavors', () {
```
felangel:
```suggestion
      test('finds the app bundle', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+              .findAppBundle(
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+
+      test('finds the apk with multi dimensional flavors', () {
```
felangel:
```suggestion
      test('finds the apk', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+    });
+
+    group('when with multi dimensional flavors and mult word flavor name', () {
```
felangel:
```suggestion
    group('when using multi-dimensional flavors and multi-word flavor name', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+    });
+
+    group('when finding multiple files', () {
+      test('throws when finding multiple aab files', () {
```
felangel:
```suggestion
      test('throws MultipleArtifactsFoundException when looking for aab', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+                project: project,
+                flavor: flavor,
+              )
+              .path,
+          equals(artifact.path),
+        );
+      });
+    });
+
+    group('when finding multiple files', () {
```
felangel:
```suggestion
    group('when multiple files are found', () {
```

packages/shorebird_cli/test/src/shorebird_android_artifacts_test.dart
```diff
@@ -0,0 +1,355 @@
...
+        expect(
+          () => shorebirdAndroidArtifacts.findAppBundle(
+            project: project,
+            flavor: flavor,
+          ),
+          throwsA(isA<MultipleArtifactsFoundException>()),
+        );
+      });
+
+      test('throws when finding multiple apk files', () {
```
felangel:
```suggestion
      test('throws MultipleArtifactsFoundException when looking for apk', () {
```

felangel (approved):
LGTM just left a bunch of small recommendations

packages/shorebird_cli/lib/src/shorebird_android_artifacts.dart
erickzanardo: apparently it exposes `dart:io` classes and I might auto selecting it, changed to `dart:io`

PR #1966, opened 2024-04-24
fix: handle multi dimensional flavors on patch android command

erickzanardo:
## Description

Change the patch command to handle multi dimensional flavors on `patch android` command.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
@@ -19,6 +19,7 @@ import 'package:shorebird_cli/src/logger.dart';
 import 'package:shorebird_cli/src/patch_diff_checker.dart';
 import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/platform/platform.dart';
+import 'package:shorebird_cli/src/shorebird_android_artifacts.dart';
 import 'package:shorebird_cli/src/shorebird_build_mixin.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 import 'package:shorebird_cli/src/shorebird_flutter.dart';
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -118,108 +119,118 @@ If this option is not provided, the version number will be determined from the p
...
-      releaseVersion = argReleaseVersion;
-    } else {
-      logger.detail('No release version provided. Determining from bundle.');
-      final flutterVersionString =
-          await shorebirdFlutter.getVersionAndRevision();
-      final buildProgress = logger.progress(
-        'Building patch with Flutter $flutterVersionString',
-      );
+
+    return runScoped(() async {
```
bryanoltman, line 123: I'm not clear why we're adding an extra runScoped here – it doesn't look like we're overriding any dependencies?

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -118,276 +119,295 @@ If this option is not provided, the version number will be determined from the p
...
-      },
-    );
+      } on ArtifactNotFoundException catch (error) {
+        logger.err(error.toString());
+        return ExitCode.software.code;
+      } on MultipleArtifactsFoundException catch (error) {
+        logger.err(error.toString());
+        return ExitCode.software.code;
+      }
+    });
```
bryanoltman, line 411: I think you may still be working on this, but if this logic stays in the patch command, it should probably be more local to the build commands, even if it results in some duplicated code.

erickzanardo: yeah, I uploaded this PR, which I guess made patch command working. But I am working in parallel in a better version of it, that implements this as we discussed internally

erickzanardo:
closing in favor of #1968

PR #1970, opened 2024-04-25
docs: add long string guidance to Dart style guide

bryanoltman:
## Description

Adds a style guide entry about string literals that extend past our line length limit.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/style_guides/dart.md
+++ b/style_guides/dart.md
@@ -53,3 +53,17 @@ group('myFunction', () {
   });
 });
 ```
+
+## DO use multiline string literals for long strings
+
+This makes the string easier to read, and reduces the chance of accidentally
+including too few or too many spaces in the string.
+
+```dart
+// BAD
+final longString = 'This is a string longer than the eighty character '
+  'limit that our lint rules enforce.';
+
+// GOOD
+final longString = '''This is a string longer than the eighty character limit that our lint rules enforce.'''
+```
--- a/style_guides/dart.md
+++ b/style_guides/dart.md
@@ -54,7 +54,7 @@ group('myFunction', () {
 });
 ```
 
-## DO use multiline string literals for long strings
+## PREFER to use multiline string literals for long strings
 
 This makes the string easier to read, and reduces the chance of accidentally
 including too few or too many spaces in the string.


style_guides/dart.md
```diff
@@ -53,3 +53,17 @@ group('myFunction', () {
   });
 });
 ```
+
+## DO use multiline string literals for long strings
+
+This makes the string easier to read, and reduces the chance of accidentally
+including too few or too many spaces in the string.
```
erickzanardo, line 60: I wonder how is this true? If you don't add the correct white space, the linter complains doesn't it?

What I don't like about using the `'''` is they need to be aligned to the beginning of the line, this looks weird in the editor when the string is declared in a block with identantion

felangel (approved):

style_guides/dart.md
bryanoltman: > I wonder how is this true? If you don't add the correct white space, the linter complains doesn't it?

Not always. I'm not sure how it decides when to flag this, but this:

```dart
  @override
  String toString() {
    return 'Multiple artifacts found in $buildDir:'
        '${foundArtifacts.map((e) => e.path)}';
  }
```

does not trigger a lint issue, but

```dart
return 'Artifact $artifactName not' 'found in $buildDir';
```

does.

bryanoltman: I don't know that I feel super strongly about this (or maybe it should be a PREFER instead of a DO?)

FWIW, multiline strings only need to move to a new line if they are actually multiline. `final foo = '''not a long string''';` is totally valid (although a single quote would make more sense in that case).

I've found the multiline string approach to be useful in cases like:

```dart
            final fixAppliedMessage =
                '''($numIssuesFixed fix${numIssuesFixed == 1 ? '' : 'es'} applied)''';
```

where, due to being indented several levels, a string just _barely_ goes over the 80 char limit, and splitting it across multiple lines would be awkward and make it harder to read.

erickzanardo: oh, I see, interesting. I guess I like the `PREFER` over `DO` suggestion. Either way I am also not super strong about either way :D

erickzanardo (approved):

PR #1989, opened 2024-04-29
feat: adding hash signature field to patches

erickzanardo:
## Description

Adds the `hash_signature` field to `CreatePatchArtifactRequest` in anticipation to the code signing feature.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/messages/create_patch_artifact/create_patch_artifact_request.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/create_patch_artifact/create_patch_artifact_request.dart
@@ -34,6 +34,10 @@ class CreatePatchArtifactRequest {
   final String hash;
 
   /// The signature of the [hash].
+  ///
+  /// Patch code signing is an opt in feature, introduced later in the life of
+  /// the product, so when this field is null, the patch does not uses code
+  /// signing.
   final String? hashSignature;
 
   /// The size of the artifact in bytes.
--- a/packages/shorebird_code_push_protocol/lib/src/messages/create_patch_artifact/create_patch_artifact_request.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/create_patch_artifact/create_patch_artifact_request.dart
@@ -14,6 +14,7 @@ class CreatePatchArtifactRequest {
     required this.platform,
     required this.hash,
     required this.size,
+    this.hashSignature,
   });
 
   /// Converts a Map<String, dynamic> to a [CreatePatchArtifactRequest]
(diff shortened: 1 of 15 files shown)


packages/shorebird_code_push_protocol/lib/src/models/release_artifact.g.dart
```diff
@@ -14,13 +14,13 @@ ReleaseArtifact _$ReleaseArtifactFromJson(Map<String, dynamic> json) =>
       json,
       ($checkedConvert) {
         final val = ReleaseArtifact(
-          id: $checkedConvert('id', (v) => v as int),
-          releaseId: $checkedConvert('release_id', (v) => v as int),
+          id: $checkedConvert('id', (v) => (v as num).toInt()),
```
felangel: Were all of these other .g.dart changes intentional? Are you using a newer version of `build_runner`?

packages/shorebird_code_push_protocol/test/src/messages/create_patch_artifact/create_patch_artifact_request_test.dart
```diff
@@ -9,6 +9,7 @@ void main() {
         platform: ReleasePlatform.android,
         hash: '1234',
         size: 9876,
+        hashSignature: 'asd',
```
felangel: It would be good to leave this test as it was before (all required fields only) and instead to add a new test where we add the optional `hashSignature` to reinforce the fact that this change is not breaking.

packages/shorebird_code_push_protocol/lib/src/messages/create_patch_artifact/create_patch_artifact_request.dart
```diff
@@ -32,6 +33,9 @@ class CreatePatchArtifactRequest {
   /// The hash of the artifact.
   final String hash;
 
+  /// The signature of the [hash].
+  final String? hashSignature;
```
felangel, line 41: Might be nice to explain why this field is nullable and what happens if it's null vs not null in the doc comment above.

felangel (commented):
Overall lgtm just left a few minor comments.

packages/shorebird_code_push_protocol/lib/src/models/release_artifact.g.dart
erickzanardo: the build runner version is configured in the project pubspec, but we don't have the lock file under version control, so I am pretty sure I ended up using a newer version than the last time that the generated files were created.

maybe it would be better if I extract these changes to its own separate PR?

felangel: Yeah I think it's fine either way just wanted to double check.

erickzanardo: extracted to its own PR: https://github.com/shorebirdtech/shorebird/pull/1994

felangel (approved):

PR #1999, opened 2024-04-30
feat: adding hidden argument for patch code signing

erickzanardo:
## Description

This PR adds a hidden flag to the CLI that hook it up to signing a patch hash. The signing code is not final yet, it is just a mechanism that will allow us to move forward with the full feature, that is why it is hidden.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -300,10 +305,24 @@ Looked in:
               releaseArtifactPath: releaseArtifactPath.value,
               patchArtifactPath: patchArtifactPath,
             );
+
+            String? hashSignature;
+            if (codeSignPrivateKey != null) {
+              final result = Process.runSync(
```
erickzanardo: Needs to investigate why this is returning empty, it works when ran in a shell. Or maybe should just go ahead and use a dart code to do the signing 🤷‍♂️

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -78,6 +78,10 @@ If this option is not provided, the version number will be determined from the p
         'staging',
         negatable: false,
         help: 'Whether to publish the patch to the staging environment.',
+      )
+      ..addOption(
+        'codesign-private-key',
+        hide: true,
```
eseidel: I don't think the final product will ever work this way (where it takes a private key and does the signing).

I could imagine it taking a key store (which is what gradle does).

But I think more likely it's going to take some sort of option (probably as part of shorebird.yaml), of a way to delegate to either a web-based signing service, or a command-line based one.

It's also possible that we will want to split our patch creation into two phases.  One which creates the artifacts, and a second which uploads them.  That's probably somewhat error prone (might be a non-trivial amount of state that has to be carefully passed to the "upload" phase), but may end up needed, depending on how long we expect signing to take (it might require human intervention).

Another possibility would be for us to do an entire `patch` release and have it in an "unsigned" state (similar to the current not-live state) and have it need to transition to a signed state before it can be made live?

eseidel: One could even imagine that we want patches to go through some sort of state machine on the server.  Where they start out in some "new" state, transition to "uploaded", "signed", "live", etc. 🤷

eseidel: Regardless of my comments above, if this unblocks demos, you should go forward with it. Just know that this isn't likely to be what you end up with for an architecture.

erickzanardo: Oh, for sure, this option in the CLI is just a "less hardecoded" way for us to do the the demo for step 1 and move forward with the other parts of the full feature

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -78,6 +79,22 @@ If this option is not provided, the version number will be determined from the p
...
+        help: r'''
+When true will sign allow the hash of the patch to be signed.
+
+The CLI will look for a script called sign_hash in the current directory, which
+is expected to receive the generated hash, and print the signed hash to stdout.
+
+A naive implementation of sign_hash in BASH might look like this:
+
+```bash
+echo $(echo "$1" | openssl dgst -sha256 -sign ./private.pem /dev/stdin | base64)
```
erickzanardo, line 94: The idea behind this approach is that we have flexibility for now in the phases where we are testing the new feature.

By hooking our CLI in a generic script that its protocol is simply to receive the hash and print the signature in the stdout we test generating the signature from the `openssl` command, and we can also test hooking up a signature service, where we would just replace the script to be one that integrates with the signature service that we want to test, etc.

Basically I tried to think in a way that will unblock us to test different approaches for the patch signing, without needing to keep us changing the CLI.

I am not attached to this approach though, so if you feel there are any issues, or if it seems too orthodox, let me know and I am happy to change it.

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
erickzanardo: Solving given the discussion (both here and the offline one) and since I proposed a different approach in the, now ready to review, PR

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -41,6 +42,9 @@ class PatchArtifactBundle {
   /// The artifact hash.
   final String hash;
 
+  /// The signature of the [hash].
+  final String? hashSignature;
```
felangel: Minor but since `hash` and `hashSignature` are super closely related it might make sense to have a separate object for `Hash` like:

```dart
class Hash {
  final String value;
  final String? signature;
}
```

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
```diff
@@ -78,6 +79,22 @@ If this option is not provided, the version number will be determined from the p
         'staging',
         negatable: false,
         help: 'Whether to publish the patch to the staging environment.',
+      )
+      ..addFlag(
+        'sign',
+        help: r'''
+When true will sign allow the hash of the patch to be signed.
```
felangel, line 86: Do we need to ship this code? It seems like our users would not benefit from this so is there any harm to keeping this on a branch for our testing?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: I would call it `SignedHash` or `SignableHash` (or maybe even something like below), but this seems like a reasonable idea.

```dart
class ArtifactFingerprint {
  final String hash;
  final String? signature;
}
```

packages/shorebird_cli/lib/src/commands/patch/patch_android_command.dart
erickzanardo: I don't have a strong opinion. I foresee that part of the changes of this PR will not change. But I see no issues in keeping this open and use it from the branch while we are working on the patch sigining feature.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: We should also document that we expect this to be a base64-encoded string, or update the variable name to make this clear.

erickzanardo: Good idea, gonna refactor it

packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
```diff
@@ -1085,5 +1085,168 @@ flavors:
...
+
+        final bundles =
+            (captured.first as Map<Arch, PatchArtifactBundle>).values;
+        for (final bundle in bundles) {
+          expect(bundle.hashSignature, equals('Signature'));
+        }
+      });
+
+      test(
+        "don't break if no sign_hash command is found (do no harm)",
```
bryanoltman: I don't think this is actually what we want to do. If a user specifies that the patch should be signed, we should not allow the generation of an unsigned patch.

packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
```diff
@@ -1085,5 +1085,168 @@ flavors:
...
+          expect(exitCode, ExitCode.success.code);
+          verify(
+            () => logger.err(
+              '''no 'sign_hash' script found in the current folder, skipping signing. ''',
+            ),
+          ).called(3);
+        },
+      );
+
+      test("don't break if no sign_hash fails (do no harm)", () async {
```
bryanoltman: Same here

packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
bryanoltman: Speaking to the "do no harm" value, I think it would be far more harmful to generate the wrong patch without the user knowing than to not generate a patch

packages/shorebird_code_push_client/test/src/code_push_client_test.dart
```diff
@@ -185,15 +185,64 @@ void main() {
...
+        expect(request.fields.containsKey('hash_signature'), isFalse);
+
         expect(
           request.url,
           equals(v1('apps/$appId/patches/$patchId/artifacts')),
         );
         expect(request.hasHeaders(expectedHeaders), isTrue);
       });
 
+      group('when a hash signature is informed', () {
```
bryanoltman: "when a hash signature is provided" maybe? I'm not sure what "informed" means here

erickzanardo: Yeah, "provided" seems better! Maybe I have been using "informed" wrong in english. In portuguese, the literal translation of informed, which is "informado", is a direct alternative for provided when "the thing that were provided" is an information/data.

packages/shorebird_cli/test/src/commands/patch/patch_android_command_test.dart
erickzanardo: So, you mean that we should stop the command execution and return a non zero exit code?

bryanoltman: I think so. If the user specifically asks for a signed patch and we're unable to generate one, I would consider that to be an error.

erickzanardo: makes sense! Will change

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -20,6 +20,24 @@ import 'package:shorebird_cli/src/third_party/flutter_tools/lib/flutter_tools.da
 import 'package:shorebird_cli/src/version.dart';
 import 'package:shorebird_code_push_client/shorebird_code_push_client.dart';
 
+/// {@template patch_artifact_fingerprint}
+///
+/// Patch artifact fingerprint that we are about to upload.
+///
+/// {@endtemplate}
+class PatchArtifactFingerprint {
+  const PatchArtifactFingerprint({
```
eseidel, line 29: I think we're going to eventually want to just embed this in the patch, rather than storing it separately on the server.  This might be the right thing for now though.

erickzanardo:
@bryanoltman given all the chats and discussion we have been having on the path that the code sign is heading, I feel we are safe to close this POC right? I don't think there is much we could salvage from here

bryanoltman:
Yeah, after consolidating the patch and release commands, it might be tricky to update this PR

PR #1998, opened 2024-04-30
refactor(shorebird_cli): introduce ArtifactBuilder to move away from ShorebirdBuildMixin

bryanoltman:
## Description

As part of https://github.com/shorebirdtech/shorebird/issues/1988, we will need to remove our reliance on mixins.

This introduces an `ArtifactBuilder` class, which will replace `ShorebirdBuildMixin`. There are no plans to migrate the existing Release and Patch commands to use this, but the Build commands will be migrated, and the new Release and Patch commands will use this.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [x] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

PR #2000, opened 2024-04-30
refactor(shorebird_cli): move buildApk to ArtifactBuilder

bryanoltman:
## Description

Following up on https://github.com/shorebirdtech/shorebird/pull/1998, adds buildApk implementation to ArtifactBuilder.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore

### DIFF
--- a/packages/shorebird_cli/lib/src/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder.dart
@@ -124,4 +124,60 @@ Either run `flutter pub get` manually, or follow the steps in ${link(uri: Uri.pa
       );
     }
   }
+
+  Future<File> buildApk({
+    String? flavor,
+    String? target,
+    Iterable<Arch>? targetPlatforms,
+    bool splitPerAbi = false,
+    List<String> argResultsRest = const [],
+  }) async {
+    await _runShorebirdBuildCommand(() async {
+      const executable = 'flutter';
+      final targetPlatformArgs = targetPlatforms?.targetPlatformArg;
+      final arguments = [
+        'build',
+        'apk',
+        '--release',
+        if (flavor != null) '--flavor=$flavor',
+        if (target != null) '--target=$target',
+        if (targetPlatformArgs != null) '--target-platform=$targetPlatformArgs',
+        // TODO(bryanoltman): reintroduce coverage when we can support this.
+        // See https://github.com/shorebirdtech/shorebird/issues/1141.
+        // coverage:ignore-start
+        if (splitPerAbi) '--split-per-abi',
+        // coverage:ignore-end
+        ...argResultsRest,
+      ];
+
+      final result = await process.run(
+        executable,
+        arguments,
+        runInShell: true,
+      );
+
+      if (result.exitCode != ExitCode.success.code) {
+        throw BuildException(
+          'Failed to build: ${result.stderr}',
+        );
+      }
+    });
+    final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
+    try {
+      return shorebirdAndroidArtifacts.findApk(
+        project: projectRoot,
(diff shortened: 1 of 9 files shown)


packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -80,6 +80,62 @@ class ArtifactBuilder {
...
+        'apk',
+        '--release',
+        if (flavor != null) '--flavor=$flavor',
+        if (target != null) '--target=$target',
+        if (targetPlatformArgs != null) '--target-platform=$targetPlatformArgs',
+        // TODO(bryanoltman): reintroduce coverage when we can support this.
+        // See https://github.com/shorebirdtech/shorebird/issues/1141.
+        // coverage:ignore-start
+        if (splitPerAbi) '--split-per-abi',
+        // coverage:ignore-end
```
felangel, line 112: If we don't support this can we just remove the flag entirely?

packages/shorebird_cli/test/src/artifact_builder_test.dart
```diff
@@ -78,17 +78,80 @@ void main() {
...
-        ),
-      ).thenReturn(File('app-release.aab'));
 
       builder = ArtifactBuilder();
     });
 
+    Future<void> verifyCorrectFlutterPubGet(
+      Future<void> Function() testCall,
+    ) async {
+      group('when flutter is installed', () {
```
felangel, line 88: Seems strange to have groups and tests nested within a top-level function like this. Can we annotate this as `@isTest` instead and implement the standard test API?

felangel (commented):
Overall looks good but I think we should make `verifyCorrectFlutterPubGet` a proper `test` API if we're going to use it as such.

packages/shorebird_cli/test/src/artifact_builder_test.dart
bryanoltman: When you say "the standard test API", what do you mean exactly? It's not really a test, but a helper function that invokes runs test to verify shared functionality (that `flutter pub get` does or does not get called after a build method)

packages/shorebird_cli/test/src/artifact_builder_test.dart
```diff
@@ -195,117 +258,159 @@ void main() {
...
-                () => shorebirdProcess.run(
-                  'flutter',
-                  ['--no-version-check', 'pub', 'get', '--offline'],
-                  runInShell: any(named: 'runInShell'),
-                  useVendedFlutter: false,
-                ),
-              );
-            });
+            verifyCorrectFlutterPubGet(
+              () async => expectLater(
```
erickzanardo: Do we need these to be async if there is no await inside them?

erickzanardo (approved):
One minor comment, but looks good already

packages/shorebird_cli/test/src/artifact_builder_test.dart
felangel: I meant annotating it with `@isTest` and supplying a description, callback etc rather than nesting groups and tests within the function.

felangel: Resolving this based on our discussions 👍

felangel (approved):

packages/shorebird_cli/test/src/artifact_builder_test.dart
bryanoltman: Discussed offline, decided this is good as is

packages/shorebird_cli/test/src/artifact_builder_test.dart
bryanoltman: Good catch, removed throughout.

PR #2007, opened 2024-05-02
pipeline demo

bryanoltman:
Demo of the pipeline concept

### DIFF
--- a/packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
+++ b/packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
@@ -1,0 +1,18 @@
+import 'package:shorebird_cli/src/command_pipeline/command_pipelines.dart';
+
+class CommandPipeline {
+  CommandPipeline({
+    required this.steps,
+    PipelineContext? context,
+  }) : context = context ?? PipelineContext({});
+
+  final PipelineContext context;
+  final List<PipelineStep> steps;
+
+  Future<void> run() async {
+    var context = PipelineContext({});
+    for (final step in steps) {
+      context = await step.run(context);
+    }
+  }
+}
--- a/packages/shorebird_cli/lib/src/command_pipeline/pipeline_context.dart
+++ b/packages/shorebird_cli/lib/src/command_pipeline/pipeline_context.dart
@@ -1,0 +1,20 @@
+class PipelineContext {
+  PipelineContext(Map<String, Object> data) : _data = data;
+
+  final Map<String, Object> _data;
+
+  PipelineContext provide<T extends Object?>(T Function() create) {
+    return PipelineContext({..._data, '$T': create});
+  }
+
+  T read<T>() {
+    final value = _data['$T'];
+    if (value == null) {
+      throw StateError('''
+context.read<$T>() called with a context that does not contain a $T.
+''');
+    }
+
+    return (value as T Function())();
+  }
+}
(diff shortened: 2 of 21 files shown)


packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:shorebird_cli/src/command_pipeline/command_pipelines.dart';
+
+class CommandPipeline {
+  CommandPipeline({
+    required this.steps,
+    PipelineContext? context,
```
felangel, line 6: Do we ever want anyone to be able to inject a context? Is this just for testing?

packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
```diff
@@ -0,0 +1,18 @@
...
+  }) : context = context ?? PipelineContext({});
+
+  final PipelineContext context;
+  final List<PipelineStep> steps;
+
+  Future<void> run() async {
+    var context = PipelineContext({});
```
felangel, line 13: Shouldn't we be using the context member variable?

packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:shorebird_cli/src/command_pipeline/command_pipelines.dart';
+
+class CommandPipeline {
+  CommandPipeline({
+    required this.steps,
+    PipelineContext? context,
+  }) : context = context ?? PipelineContext({});
```
felangel, line 7:
```suggestion
  }) : this.context = context ?? PipelineContext({});
```

packages/shorebird_cli/lib/src/command_pipeline/pipeline_context.dart
```diff
@@ -0,0 +1,22 @@
+class PipelineContext {
+  PipelineContext() : _data = {};
+
+  PipelineContext._(Map<String, Object> data) : _data = data;
```
erickzanardo, line 4: Originally I thought on this being a class with defined attributes instead of a map.

That would mean that if a step needs to include a new attribute, they would need to change this class. But I assumed that this would not change that much and that would make this more compile safer?

Either way I also like the flexibility of a map brings, so no strong opinions

packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
bryanoltman: The idea is that this will be seeded by the release command, e.g., `final context = PipelineContext()..provide(() => results);`

packages/shorebird_cli/lib/src/command_pipeline/command_pipeline.dart
bryanoltman: We certainly should, yes :)

packages/shorebird_cli/lib/src/command_pipeline/pipeline_context.dart
bryanoltman: Can you give examples of what you expected the properties to be?

erickzanardo: Something like:

```dart
class PipelineContext {
  final Directory projectRoot;
  final String? releaseVersion;
  final bool bundlePatchSignKey;
}
```

Idk if all those attributes makes sense, but were the ones that came to mind while trying to think in an example haha

bryanoltman: Gotcha. I think I'm going to start with a map to mimic the behavior of dart frog and provider, but we'll see if I end up fighting that too much.

bryanoltman: Yeah, I think we'll want some other way to key than type – we have a lot of strings, and the way around that would be to wrap them in classes, which seems cumbersome

PR #2010, opened 2024-05-05
refactor(shorebird_cli): add iOS artifact lookup methods to ArtifactManager

bryanoltman:
## Description

Adds functions from ShorebirdArtifactsMixin to ArtifactManager for use with new release command.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -173,4 +176,88 @@ Failed to create diff (exit code ${result.exitCode}).
...
+  Directory? getIosAppDirectory({required Directory xcarchiveDirectory}) {
+    final applicationsDirectory = Directory(
+      p.join(
+        xcarchiveDirectory.path,
+        'Products',
+        'Applications',
+      ),
+    );
+
+    if (!applicationsDirectory.existsSync()) {
```
erickzanardo, line 212: Nit:

This method and the one above seems super similar, given a dir, it concatenate with additional path, check if it exist and return the first one with a extension right? Maybe it is worth extracting this to generic method?

erickzanardo (approved):
One minor suggestion, but LGTM!

felangel (approved):

packages/shorebird_cli/lib/src/artifact_manager.dart
bryanoltman: Agreed. If we find ourselves doing this another time, we should definitely do that.

PR #2030, opened 2024-05-07
refactor(shorebird_cli): introduce new patch command

bryanoltman:
## Description

Adds a new (hidden) patch command that currently supports patch `android` targets.

Usage:

```
shorebird patch-new -p android
```

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
```diff
@@ -0,0 +1,177 @@
...
+  @override
+  String get primaryReleaseArtifactArch => 'aab';
+
+  @override
+  ArchiveDiffer get archiveDiffer => AndroidArchiveDiffer();
+
+  @override
+  Future<void> assertArgsAreValid() async {
+    // Nothing to do here
+  }
```
eseidel: Should empty be the default impl?

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
```diff
@@ -0,0 +1,177 @@
...
+  Future<void> assertPreconditions() async {
+    try {
+      await shorebirdValidator.validatePreconditions(
+        checkUserIsAuthenticated: true,
+        checkShorebirdInitialized: true,
+        validators: doctor.androidCommandValidators,
+      );
+    } on PreconditionFailedException catch (e) {
+      exit(e.exitCode.code);
+    }
```
eseidel, line 45: Should the caller do this?

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
```diff
@@ -0,0 +1,177 @@
...
+      artifact.path,
+    );
+  }
+
+  @override
+  Future<File> patchArtifactForDiffCheck() async {
+    return shorebirdAndroidArtifacts.findAab(
+      project: projectRoot,
+      flavor: flavor,
+    );
```
eseidel: FYI, if we don't await the future here, this call won't show up in the call stack, only the call where it is actually awaited will (that may not matter).

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
bryanoltman: I think that's fair – there are fewer bad arg combinations in patching than in releasing

bryanoltman: Although we have to provide an implementation here because `Patcher` is abstract, so maybe a default implementation wouldn't offer much.

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
bryanoltman: The general contract in the corresponding release commands is that the functions themselves handle logging and exiting (if necessary) and should not throw.

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
bryanoltman: `findAab` doesn't return a future, so nothing to await

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
eseidel: I think an abstract class can still have default impls in Dart, no?

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
bryanoltman: Also turns out to not be needed. Deleted! :)

packages/shorebird_cli/lib/src/commands/patch_new/android_patcher.dart
bryanoltman: Yep, they can and they work how you'd think. I accidentally provided a default impl for the wrong function 🤦  Done.

erickzanardo (approved):

felangel (approved):

PR #2034, opened 2024-05-08
feat: adding local file logging to the cli

erickzanardo:
## Description

This PR proposes a wrapper logger on top of mason logger which will also log information into a file located in a `logs` folder inside the shorebird installation root.

This will allow us to better track errors in the CLI since we can ask for the error log from customers.

This PR don't implement the logger usage in any command yet though. I feel we can do that in follow ups if the proposal gets accepted.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore


packages/shorebird_cli/lib/src/shorebird_logger.dart
```diff
@@ -0,0 +1,107 @@
...
+      p.join(
+        shorebirdEnv.shorebirdRoot.path,
+        'logs',
+        '${timestamp}_$_logFileName',
+      ),
+    );
+
+    if (!file.parent.existsSync()) {
+      file.parent.createSync(recursive: true);
+    }
```
felangel:
```suggestion
    if (!file.existsSync()) {
      file.createSync(recursive: true);
    }
```

packages/shorebird_cli/lib/src/shorebird_logger.dart
```diff
@@ -0,0 +1,107 @@
...
+  /// {@macro shorebird_logger}
+  ShorebirdLogger({
+    Logger? logger,
+  }) : _logger = logger ?? Logger();
+
+  final Logger _logger;
+
+  static const _logFileName = 'shorebird.log';
+
+  late final File logFile = (() {
```
felangel: minor but couldn't this be initialized in the body of the constructor instead of having to wrap and call this anonymous function like this?

erickzanardo: For sure. I just did like that since I kinda like this approach to initialize "complex" variables. By doing so we can make sure the variable initialization stays "close" to where it is declared, so it is easy to spot.

Happy to change if this is a pattern that we would like to avoid.

bryanoltman (approved):

packages/shorebird_cli/lib/src/logger.dart
```diff
@@ -1,8 +1,110 @@
...
-Logger get logger => read(loggerRef);
+ShorebirdLogger get logger => read(loggerRef);
+
+class ShorebirdLogger extends Logger {
+  /// {@macro shorebird_logger}
+  ShorebirdLogger({
+    Logger? logger,
+  }) : _logger = logger ?? Logger();
+
+  final Logger _logger;
```
felangel: Since this extends `Logger` we don't need to inject an maintain an internal logger right?

erickzanardo: Right 🤡!

felangel (approved):

PR #2080, opened 2024-05-11
fix(shorebird_cli): make cache clear async

bryanoltman:
## Description

This allows the progress spinner to display while `shorebird cache clear` is doing its thing (currently, the command just hangs until the cache has been cleared, which can take several seconds).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -106,14 +106,11 @@ class Cache {
 
   Future<void> clear() async {
     final cacheDir = shorebirdCacheDirectory;
-    if (cacheDir.existsSync()) {
-      await cacheDir.delete(recursive: true);
-    }
-
     final logsDirectory = shorebirdEnv.logsDirectory;
-    if (logsDirectory.existsSync()) {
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -104,15 +104,15 @@ class Cache {
 
   String get storageBucket => 'download.shorebird.dev';
 
-  void clear() {
+  Future<void> clear() async {
     final cacheDir = shorebirdCacheDirectory;
     if (cacheDir.existsSync()) {
-      cacheDir.deleteSync(recursive: true);
+      await cacheDir.delete(recursive: true);
```
felangel: We can do them in parallel in this case right?

felangel (approved):

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Good call, done

PR #2087, opened 2024-05-13
feat(shorebird_cli): call `flutter precache` when installing new flutter rev

bryanoltman:
## Description

This provides better visibility into Flutter artifact downloads and why builds take as long as they do.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_flutter.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_flutter.dart
@@ -1,11 +1,13 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:scoped/scoped.dart';
 import 'package:shorebird_cli/src/executables/executables.dart';
 import 'package:shorebird_cli/src/logger.dart';
+import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 import 'package:shorebird_cli/src/shorebird_process.dart';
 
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -62,7 +71,24 @@ class ShorebirdFlutter {
...
+        ['precache', ...precacheArgs],
+        workingDirectory: targetDirectory.path,
+        runInShell: true,
+      );
+      precacheProgress.complete();
+    } catch (_) {
+      precacheProgress.fail('Failed to precache Flutter $version');
+      logger.info(
+        '''This is not a critical error, but your next build make take longer than usual.''',
+      );
```
felangel, line 90: minor but maybe we can consolidate these into a single log like:

```dart
precacheProgress.fail('Failed to precache Flutter $version -- the next build make take longer than usual. ');
```

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -62,7 +71,24 @@ class ShorebirdFlutter {
       rethrow;
     }
 
-    installProgress.complete();
+    final precacheProgress = logger.progress(
+      'Running ${lightCyan.wrap('flutter precache')}',
+    );
+
+    try {
+      await process.run(
```
felangel, line 79: What about initial install? Should we also do this as part of the install script for consistency?

felangel (approved):

packages/shorebird_cli/lib/src/shorebird_flutter.dart
bryanoltman: Good idea, filed https://github.com/shorebirdtech/shorebird/issues/2093

packages/shorebird_cli/lib/src/shorebird_flutter.dart
bryanoltman: I wanted to have a part of the message that wasn't a "failure" or "error" message to head off support requests when/if this fails (just to be extra clear that the thing that went wrong isn't going to cause shorebird to not work)

PR #2122, opened 2024-05-17
feat(shorebird_cli): always prompt for release version when patching

bryanoltman:
## Description

Updates the shorebird patch command to prompt for release version if not provided by the --release-version arg (unless on CI, where we maintain the current behavior).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -179,20 +179,33 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
     final app = await codePushClientWrapper.getApp(appId: appId);
 
     File? patchArtifact;
-    final String releaseVersion;
+    final Release release;
     if (results.wasParsed('release-version')) {
-      releaseVersion = results['release-version'] as String;
+      final releaseVersion = results['release-version'] as String;
+      release = await codePushClientWrapper.getRelease(
+        appId: appId,
+        releaseVersion: releaseVersion,
+      );
+    } else if (shorebirdEnv.canAcceptUserInput) {
+      release = await promptForRelease();
     } else {
+      logger.info(
+        '''Tip: make your patches build faster by specifying --release-version''',
+      );
       patchArtifact = await patcher.buildPatchArtifact();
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -179,20 +179,33 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
...
     File? patchArtifact;
-    final String releaseVersion;
+    final Release release;
     if (results.wasParsed('release-version')) {
-      releaseVersion = results['release-version'] as String;
+      final releaseVersion = results['release-version'] as String;
+      release = await codePushClientWrapper.getRelease(
+        appId: appId,
+        releaseVersion: releaseVersion,
+      );
```
eseidel, line 189: What is the behavior when this fails?

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -260,6 +273,17 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
     );
   }
 
+  Future<Release> promptForRelease() async {
+    final releases = await codePushClientWrapper.getReleases(
+      appId: appId,
+    );
+    return logger.chooseOne<Release>(
+      'Which release would you like to patch?',
+      choices: releases,
```
eseidel: Do these need to be displayed in a specific order?

eseidel (approved):

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
bryanoltman: `codePushClientWrapper.getRelease` logs an error and exits when it fails.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
bryanoltman: I believe the API returns them in reverse chronological order (most recently created first), but added a sort just in case.

PR #2130, opened 2024-05-20
feat: adding patch signing key path arg to release command

erickzanardo:
## Description

Adding a new argument to the release command, where the user can provide a path for a pem file key, which will be bundled in the release artifact, so patch signatures can be verified.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder.dart
@@ -46,7 +46,7 @@ class ArtifactBuilder {
     String? target,
     Iterable<Arch>? targetPlatforms,
     List<String> args = const [],
-    String? encodedPublicKey,
+    String? base64PublicKey,
   }) async {
     await _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
@@ -65,8 +65,8 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
-        environment: encodedPublicKey != null
-            ? {'SHOREBIRD_PUBLIC_KEY': encodedPublicKey}
+        environment: base64PublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': base64PublicKey}
             : null,
       );
 
@@ -102,7 +102,7 @@ class ArtifactBuilder {
     Iterable<Arch>? targetPlatforms,
     bool splitPerAbi = false,
     List<String> args = const [],
-    String? encodedPublicKey,
+    String? base64PublicKey,
   }) async {
     await _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
@@ -126,8 +126,8 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
-        environment: encodedPublicKey != null
-            ? {'SHOREBIRD_PUBLIC_KEY': encodedPublicKey}
+        environment: base64PublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': base64PublicKey}
             : null,
       );
 
--- a/packages/shorebird_cli/lib/src/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder.dart
@@ -46,7 +46,7 @@ class ArtifactBuilder {
     String? target,
     Iterable<Arch>? targetPlatforms,
     List<String> args = const [],
-    String? encodedPatchSigningPublicKey,
+    String? encodedPublicKey,
   }) async {
     await _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
@@ -65,8 +65,8 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
-        environment: encodedPatchSigningPublicKey != null
-            ? {'SHOREBIRD_PUBLIC_KEY': encodedPatchSigningPublicKey}
+        environment: encodedPublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': encodedPublicKey}
             : null,
       );
 
@@ -102,7 +102,7 @@ class ArtifactBuilder {
     Iterable<Arch>? targetPlatforms,
     bool splitPerAbi = false,
     List<String> args = const [],
-    String? encodedPatchSigningPublicKey,
+    String? encodedPublicKey,
   }) async {
     await _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
@@ -126,8 +126,8 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
-        environment: encodedPatchSigningPublicKey != null
-            ? {'SHOREBIRD_PUBLIC_KEY': encodedPatchSigningPublicKey}
+        environment: encodedPublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': encodedPublicKey}
             : null,
       );
 
--- a/packages/shorebird_cli/lib/src/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder.dart
@@ -46,6 +46,7 @@ class ArtifactBuilder {
     String? target,
     Iterable<Arch>? targetPlatforms,
     List<String> args = const [],
+    String? encodedPatchSigningPublicKey,
   }) async {
     await _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
@@ -64,6 +65,9 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
+        environment: encodedPatchSigningPublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': encodedPatchSigningPublicKey}
+            : null,
       );
 
       if (result.exitCode != ExitCode.success.code) {
@@ -98,6 +102,7 @@ class ArtifactBuilder {
     Iterable<Arch>? targetPlatforms,
     bool splitPerAbi = false,
     List<String> args = const [],
+    String? encodedPatchSigningPublicKey,
   }) async {
     await _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
@@ -121,6 +126,9 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
+        environment: encodedPatchSigningPublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': encodedPatchSigningPublicKey}
+            : null,
       );
 
       if (result.exitCode != ExitCode.success.code) {
--- a/packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
@@ -1,3 +1,5 @@
+import 'dart:convert';
+
 import 'package:mason_logger/mason_logger.dart';
 import 'package:shorebird_cli/src/artifact_builder.dart';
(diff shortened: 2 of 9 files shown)


packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
```diff
@@ -90,12 +92,25 @@ Please comment and upvote ${link(uri: Uri.parse('https://github.com/shorebirdtec
         .progress('Building app bundle with Flutter $flutterVersionString');
 
     final File aab;
+
+    final patchSigningKeyPath =
+        argResults['patch-signing-public-key-path'] as String?;
```
bryanoltman: I'd maybe shorten this—`public-key-path` perhaps?

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -126,8 +133,24 @@ of the iOS app that is using this module.''',
...
 
+  void _assertArgs() {
+    final patchSignKeyPath =
+        results['patch-signing-public-key-path'] as String?;
+    if (patchSignKeyPath != null) {
+      final file = File(patchSignKeyPath);
+      if (!file.existsSync()) {
+        logger.err(
+          'Received a patch signing key path, but no files exists for '
+          'it: $patchSignKeyPath',
```
bryanoltman:
```suggestion
'No file found at $patchSignKeyPath'
```

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -126,8 +133,24 @@ of the iOS app that is using this module.''',
   @override
   String get name => 'release';
 
+  void _assertArgs() {
```
bryanoltman: If we're doing arg validation both here and in the Releaser subclasses, how should we choose where to validate the arg? I don't think this is wrong, I just think we need to make it clearer.

erickzanardo: Yeah, I spent some time wondering about that as well. I guess ideally validation of the args should stay in a single place.

But this one in particular is an arg that will be used by multiple releasers, so ended up adding this validation here so we don't repeat ourselves.

Wdyt? Does this makes sense? Maybe we can think in a better way to not repeat ourselves and to organize this validation?

erickzanardo: I spent some time thinking on it and did some changes, I think I reached a good middle ground solution. Let me know what you think.

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -46,7 +46,7 @@ class ArtifactBuilder {
     String? target,
     Iterable<Arch>? targetPlatforms,
     List<String> args = const [],
-    String? encodedPatchSigningPublicKey,
+    String? encodedPublicKey,
```
bryanoltman: Don't forget to update the doc comment above to explain this. When/how is it used, how is it encoded (although maybe changing the name to `base64PublicKey` or something might remove the need for that), etc

packages/shorebird_cli/lib/src/commands/release/releaser.dart
```diff
@@ -71,4 +73,22 @@ abstract class Releaser {
...
+    final patchSignKeyPath = argResults['public-key-path'] as String?;
+    if (patchSignKeyPath != null) {
+      final file = File(patchSignKeyPath);
+      if (!file.existsSync()) {
+        logger.err(
+          'No file found at $patchSignKeyPath',
+        );
+        exit(ExitCode.software.code);
+      }
+    }
```
bryanoltman: Could we potentially make this a more generic? ex: "we parsed argument `name` that should point to an existing file, but no file is at that path"? Then we could just call that directly from the releaser's `assertArgsAreValid` and potentially use it elsewhere.

packages/shorebird_cli/lib/src/extensions/arg_results.dart
```diff
@@ -39,3 +41,17 @@ extension OptionFinder on ArgResults {
...
+
+/// Extension on [ArgResults] to provide file validation.
+extension FileValidation on ArgResults {
+  /// Checks if an option is a path that points to an existing file.
+  ///
+  /// This method will only return false when the argument with [name] is
+  /// provided and the file does not exist.
+  bool wasParsedAndFileExists(String name) {
+    final filePath = this[name] as String?;
+    if (filePath == null) return true;
```
bryanoltman: I'm not sure this is right—if `filePath` is null, that would mean it was not parsed, correct? ArgResults has a `wasParsed` method that might be useful here.

packages/shorebird_cli/lib/src/artifact_builder.dart
erickzanardo: I tried to come up with a good comment, but I had a hard time figuring out one. I guess just renaming, like you suggested, is already good

packages/shorebird_cli/lib/src/extensions/arg_results.dart
```diff
@@ -39,3 +41,15 @@ extension OptionFinder on ArgResults {
     return null;
   }
 }
+
+/// Extension on [ArgResults] to provide file related extensions.
+extension FileArgs on ArgResults {
+  /// Returns a [File] from the argument [name].
```
bryanoltman:
```suggestion
  /// Returns a [File] from the argument [name] or null if the argument was not provided.
```

packages/shorebird_cli/lib/src/extensions/file.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:io';
+
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/logger.dart';
+
+/// Extension methods for validating [File]s.
+extension FileValidations on File {
+  /// Asserts that the file exists.
+  ///
+  /// When it doesn't, logs an error and exits with [ExitCode.usage].
```
bryanoltman:
```suggestion
  /// Logs an error and exits with [ExitCode.usage] if this file does not exist.
```

bryanoltman (approved):

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -114,6 +114,13 @@ of the iOS app that is using this module.''',
         help: 'The target platform(s) for which the app is compiled.',
         defaultsTo: Arch.values.map((arch) => arch.targetPlatformCliArg),
         allowed: Arch.values.map((arch) => arch.targetPlatformCliArg),
+      )
+      ..addOption(
+        'public-key-path',
```
felangel, line 119: This is more of a general question but why would users need to specify the public key path to start? Wouldn't we just be using a Shorebird managed public/private key pair by default?

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -64,6 +65,9 @@ class ArtifactBuilder {
         executable,
         arguments,
         runInShell: true,
+        environment: base64PublicKey != null
+            ? {'SHOREBIRD_PUBLIC_KEY': base64PublicKey}
```
felangel, line 69: Since we're ultimately setting an environment variable couldn't this PR be simplified and check for the presence of this environment variable instead of adding a command-line option? Environment variables would then get forwarded to the subprocess.

packages/shorebird_cli/lib/src/commands/release/release_command.dart
bryanoltman: We decided yesterday that we'll be asking users to manage their own keys to start.

packages/shorebird_cli/lib/src/artifact_builder.dart
bryanoltman: It could, but having the user provide a path to a file as a CLI arg is more in-line with how our system currently operates. I think we'd ideally pass this to the flutter tool as an arg as well, but I believe it was easier to do via an env var (@erickzanardo could speak more on this part)

erickzanardo: Yeah, this is done this way since passing an argument to the code that bundles the shorebird.yaml file on android is not trivial.

That is done in the gradle layer (https://github.com/shorebirdtech/flutter/blob/shorebird/dev/packages/flutter_tools/gradle/src/main/groovy/flutter.groovy#L1337), and AFAIK, to pass an argument there, we would need to do multiple changes in the gradle tasks, I might be wrong on this since gradle is something I am still learning, but that is how we managed to work so far.

Happy to revisit this eventually, but I feel that in favor of shipping code signing soon, I guess this is good enough.

Let me know if any of you thinks otherwise

packages/shorebird_cli/test/src/commands/release/android_releaser_test.dart
```diff
@@ -211,6 +213,43 @@ void main() {
...
+            p.join(
+              Directory.systemTemp.createTempSync().path,
+              'public-key.pem',
+            ),
+          )..writeAsStringSync('public key');
+          when(() => argResults['public-key-path'])
+              .thenReturn(publicKeyFile.path);
+        });
+        test('returns normally', () async {
+          expect(
```
bryanoltman:
```suggestion
          )..writeAsStringSync('public key');
          when(() => argResults['public-key-path'])
              .thenReturn(publicKeyFile.path);
        });

        test('returns normally', () async {
          expect(
```

packages/shorebird_cli/test/src/commands/release/android_releaser_test.dart
```diff
@@ -211,6 +213,43 @@ void main() {
...
+        });
+      });
+
+      group('when a public key is provided but it does not exists', () {
+        setUp(() {
+          when(() => argResults['artifact']).thenReturn('apk');
+          when(() => argResults['public-key-path'])
+              .thenReturn('non-existing-key.pem');
+        });
+        test('logs and exits with usage err', () async {
```
bryanoltman:
```suggestion
              .thenReturn('non-existing-key.pem');
        });

        test('logs and exits with usage err', () async {
```

packages/shorebird_cli/lib/src/commands/release/release_command.dart
felangel: Okay just asked because it seems asking users to manage their own keys to start is both more work for us and more work for the users no? Either way feel free to ignore me as I haven't really been involved in the discussions around this work.

bryanoltman: Definitely not more work for us – it frees us from having to deal with a signing service, which is a significant chunk of work

packages/shorebird_cli/lib/src/artifact_builder.dart
felangel: Sounds good I only asked because it seemed like a lot more work to modify release commands to accept a command-line arg just to pass an environment variable to the flutter tool. Seemed much simpler to just provide the public key as an environment variable from the start since folks will likely be doing that anyway in CI environment via GitHub secrets manager etc.

packages/shorebird_cli/lib/src/commands/release/release_command.dart
felangel: Got it, I was thinking we could start by having a single shorebird public/private key pair which we use for all patches for the first p.o.c which shouldn't require a signing service but if we're starting off with per-app keys then makes sense 👍

PR #2133, opened 2024-05-21
feat: adding private key argument to patch command

erickzanardo:
## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/patch/aar_patcher.dart
```diff
@@ -46,6 +46,9 @@ class AarPatcher extends Patcher {
   @override
   ReleaseType get releaseType => ReleaseType.aar;
 
+  @override
+  Future<void> assertArgsAreValid() async {}
+
```
bryanoltman: I don't think this is needed

packages/shorebird_cli/lib/src/commands/patch/patcher.dart
```diff
@@ -67,7 +67,7 @@ https://docs.shorebird.dev/status#link-percentage-ios
   Future<void> assertPreconditions();
 
   /// Asserts that the combination arguments passed to the command are valid.
-  Future<void> assertArgsAreValid() async {}
+  Future<void> assertArgsAreValid();
```
bryanoltman: I'd revert this change. That way, we don't need to provide no-op implementations for the patchers that don't need them.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -92,6 +92,13 @@ of the iOS app that is using this module.''',
         abbr: 'n',
         negatable: false,
         help: 'Validate but do not upload the patch.',
+      )
+      ..addOption(
+        'private-key-path',
+        hide: true,
+        help: '''
+The path for a private key file that will be used to sign the patch artifact.
+''',
```
bryanoltman, line 101: We should (now or at least before this is announced) add more info about what format this file should be in, how to generate it, etc. Could be a link to docs.

packages/shorebird_cli/test/src/commands/patch/aar_patcher_test.dart
```diff
@@ -115,6 +115,12 @@ void main() {
       patcher = AarPatcher(argResults: argResults, flavor: null, target: null);
     });
 
+    group('assertArgsAreValid', () {
+      test('has no specific validations', () {
+        expect(patcher.assertArgsAreValid, returnsNormally);
+      });
+    });
```
bryanoltman: This could be removed too

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -436,6 +441,60 @@ Looked in:
...
+            expect(bundle.hashSignature, isNull);
+          }
+        });
+
+        group('when a private key is provided', () {
+          const testPrivateKeyContent = '''
+-----BEGIN PRIVATE KEY-----
+key
+-----END PRIVATE KEY-----
+''';
```
bryanoltman: I don't think we care about the content of the private key file, just that it exists, right? I think we could get away with writing an empty file, which would make the tests a bit clearer.

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -436,6 +441,60 @@ Looked in:
...
+          const mockSignature = 'mock-signature';
```
bryanoltman: This should be renamed – it's not a `Mock`

packages/shorebird_cli/test/src/commands/patch/ios_patcher_test.dart
```diff
@@ -138,6 +138,12 @@ void main() {
         );
       });
 
+      group('assertArgsAreValid', () {
+        test('has no specific validations', () {
+          expect(patcher.assertArgsAreValid, returnsNormally);
+        });
+      });
+
       group('archiveDiffer', () {
```
bryanoltman, line 147: This can be removed

packages/shorebird_cli/test/src/commands/patch/patcher_test.dart
```diff
@@ -72,4 +72,9 @@ class _TestPatcher extends Patcher {
 
   @override
   ReleaseType get releaseType => throw UnimplementedError();
+
+  @override
+  Future<void> assertArgsAreValid() {
+    throw UnimplementedError();
+  }
 }
```
bryanoltman: This can be removed

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -436,6 +441,60 @@ Looked in:
...
+            );
+
+            const expectedSignatures = [
+              mockSignature,
+              mockSignature,
+              mockSignature,
+            ];
+
+            final signatures =
+                result.values.map((bundle) => bundle.hashSignature).toList();
```
bryanoltman: Might be useful to verify that we're generating a signature per hash? This reads like we might be using the same signature for every hash, which would (in most cases) not be right.

packages/shorebird_cli/lib/src/commands/patch/patcher.dart
erickzanardo: Yeah, but I made that change because otherwise coverage would complain about the default (empty) implementation not being covered. Happy to revert it though if there is a better approach

bryanoltman (approved):

PR #2142, opened 2024-05-23
fix(shorebird_cli): don't warn about changed shorebird.yaml in patches

bryanoltman:
## Description

The new support for signed patches introduced in https://github.com/shorebirdtech/shorebird/pull/2133 will cause the patch command to warned about a changed shorebird_cli asset because the release build will contain a public key and the patch build will not. We should not warn users about this, as it is expected.

We could also do something more sophisticated if we are worried about users thinking they can make changes to shorebird.yaml in patches (e.g., changing auto update behavior).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/archive_analysis/archive_differ.dart
+++ b/packages/shorebird_cli/lib/src/archive_analysis/archive_differ.dart
@@ -16,6 +16,7 @@ abstract class ArchiveDiffer {
     'AssetManifest.bin',
     'AssetManifest.json',
     'NOTICES.Z',
+    'shorebird.yaml',
   };
 
   /// Whether there are asset differences between the archives that may cause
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/archive_analysis/archive_differ.dart
```diff
@@ -16,6 +16,7 @@ abstract class ArchiveDiffer {
     'AssetManifest.bin',
     'AssetManifest.json',
     'NOTICES.Z',
+    'shorebird.yaml',
```
eseidel, line 19: This is kinda a big hammer.  You might want to just make the patch shorebird.yaml look like the release on instead of doing this?

bryanoltman:
Closing this in favor of requiring users to provide a public key when patching.

PR #2143, opened 2024-05-23
feat: ios release with public key bundling

erickzanardo:
## Description

Adds the necessary logic to bundle the public key in an iOS release.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/release/ios_releaser.dart
```diff
@@ -98,13 +103,20 @@ class IosReleaser extends Releaser {
     final flutterVersionString = await shorebirdFlutter.getVersionAndRevision();
     final buildProgress =
         logger.progress('Building ipa with Flutter $flutterVersionString');
+    final publicKeyFile = argResults.file('public-key-path');
+
+    final base64PublicKey = publicKeyFile != null
+        ? base64Encode(publicKeyFile.readAsBytesSync())
+        : null;
```
bryanoltman: We should probably move this up to the Releaser/ReleaseCommand level, as this will be applicable for all platforms

erickzanardo: Agreed! I moved it to the `ArgResults` extension, that way we can share this logic in other commands too? Not sure if an extension is the most appropriate way, but I feel it kinda worked well, lmk what you think

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -34,6 +34,10 @@ final artifactBuilderRef = create(ArtifactBuilder.new);
 /// The [ArtifactBuilder] instance available in the current zone.
 ArtifactBuilder get artifactBuilder => read(artifactBuilderRef);
 
+extension on String {
+  Map<String, String> toPublicKeyEnv() => {'SHOREBIRD_PUBLIC_KEY': this};
+}
```
bryanoltman: Nice extension. I'd add a brief doc string explaining what this does/why it's used.

packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
```diff
@@ -63,7 +60,7 @@ class AndroidReleaser extends Releaser {
 
   @override
   Future<void> assertArgsAreValid() async {
-    argResults.file('public-key-path')?.assertExists();
+    argResults.assertPublicKey();
```
bryanoltman: This could maybe have a more verbose name. `assertPublicKey` suggests that this will fail if no public key is provided.

packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
```diff
@@ -96,10 +93,7 @@ Please comment and upvote ${link(uri: Uri.parse('https://github.com/shorebirdtec
 
     final File aab;
 
-    final publicKeyFile = argResults.file('public-key-path');
-    final base64PublicKey = publicKeyFile != null
-        ? base64Encode(publicKeyFile.readAsBytesSync())
-        : null;
+    final base64PublicKey = argResults.encodedPublicKey;
```
bryanoltman, line 96: Nice! I'd just inline this below in the call to buildAppBundle.

packages/shorebird_cli/test/src/commands/release/ios_releaser_test.dart
```diff
@@ -282,6 +284,54 @@ void main() {
...
         });
 
+        group('when a patch signing key path is provided', () {
+          late File patchSigningPublicKeyFile;
+
+          setUp(() {
+            patchSigningPublicKeyFile = File(
+              p.join(
+                Directory.systemTemp.createTempSync().path,
+                'patch-signing-public-key.pem',
```
bryanoltman: This doesn't affect the logic of the tests, but public keys are consumed in a `.der` format

packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
erickzanardo: In this instance I would rather keep the variable, since we might call a `buildApk` a couple of lines below, and we also pass the key there. With the variable we read the file just once.

bryanoltman: 👍  makes sense!

packages/shorebird_cli/lib/src/artifact_builder.dart
erickzanardo: Added, lmk if that is understandable

packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
erickzanardo: Makes sense, I have renamed, lmk what you think

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -34,6 +34,18 @@ final artifactBuilderRef = create(ArtifactBuilder.new);
 /// The [ArtifactBuilder] instance available in the current zone.
 ArtifactBuilder get artifactBuilder => read(artifactBuilderRef);
 
+extension on String {
+  /// Helper method to convert a String value that is supposed
+  /// to represent a public key, into a map of environment variables
+  /// containing itself.
```
bryanoltman:
```suggestion
  /// Converts this base64-encoded public key into the Map<String, String>:
  ///   {'SHOREBIRD_PUBLIC_KEY': this}
```

packages/shorebird_cli/test/src/commands/release/ios_releaser_test.dart
```diff
@@ -243,6 +245,43 @@ void main() {
...
+
+          test('returns normally', () async {
+            expect(
+              () => runWithOverrides(iosReleaser.assertArgsAreValid),
+              returnsNormally,
+            );
+          });
+        });
+
+        group('when a public key is provided but it does not exists', () {
```
bryanoltman:
```suggestion
        group('when the provided public key is a nonexistent file', () {
```

packages/shorebird_cli/test/src/commands/release/ios_releaser_test.dart
```diff
@@ -282,6 +321,54 @@ void main() {
...
+                flavor: any(named: 'flavor'),
+                target: any(named: 'target'),
+                args: any(named: 'args'),
+                base64PublicKey: any(named: 'base64PublicKey'),
+              ),
+            ).thenAnswer((_) async => File(''));
+          });
+
+          test(
+            'encodes the patch signing public key and forward it to buildAab',
```
bryanoltman:
```suggestion
            'encodes the patch signing public key and forward it to buildIpa',
```

bryanoltman (approved):
A couple small comments. LGTM!

PR #2144, opened 2024-05-23
feat: adding signing hash for ios patches

erickzanardo:
## Description

Adds signing logic to the ios patch command

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -9,12 +9,12 @@ import 'package:shorebird_cli/src/archive_analysis/archive_differ.dart';
 import 'package:shorebird_cli/src/artifact_builder.dart';
 import 'package:shorebird_cli/src/artifact_manager.dart';
 import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
+import 'package:shorebird_cli/src/code_signer.dart';
 import 'package:shorebird_cli/src/commands/patch/patcher.dart';
 import 'package:shorebird_cli/src/common_arguments.dart';
 import 'package:shorebird_cli/src/doctor.dart';
 import 'package:shorebird_cli/src/executables/executables.dart';
 import 'package:shorebird_cli/src/extensions/arg_results.dart';
-import 'package:shorebird_cli/src/extensions/string.dart';
 import 'package:shorebird_cli/src/logger.dart';
 import 'package:shorebird_cli/src/patch_diff_checker.dart';
 import 'package:shorebird_cli/src/platform.dart';
@@ -215,7 +215,12 @@ class IosPatcher extends Patcher {
     final patchFileSize = patchFile.statSync().size;
     final privateKeyFile = argResults.file(CommonArguments.privateKeyArgName);
     final hash = sha256.convert(patchBuildFile.readAsBytesSync()).toString();
-    final hashSignature = (hash, privateKeyFile).signature;
+    final hashSignature = privateKeyFile != null
+        ? codeSigner.sign(
+            message: hash,
+            privateKeyPemFile: privateKeyFile,
+          )
+        : null;
 
     return {
       Arch.arm64: PatchArtifactBundle(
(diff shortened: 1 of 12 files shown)


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -210,12 +212,21 @@ class IosPatcher extends Patcher {
     }
 
     final patchFileSize = patchFile.statSync().size;
+    final privateKeyFile = argResults.privateKeyFile;
+    final hash = sha256.convert(patchBuildFile.readAsBytesSync()).toString();
+    final hashSignature = privateKeyFile != null
```
erickzanardo, line 218: Need to think more on how to share this code since it is pretty much the same as the android patcher

packages/shorebird_cli/lib/src/extensions/arg_results.dart
```diff
@@ -61,6 +61,9 @@ extension CodeSign on ArgResults {
         ? base64Encode(publicKeyFile.readAsBytesSync())
         : null;
   }
+
+  /// Gets a File reference to private from the args if any.
+  File? get privateKeyFile => file('private-key-path');
```
bryanoltman: It might be better to have these arg names as constants. It would probably make this method unnecessary.

bryanoltman: Doing this would also make it impossible to mistype the arg name, which I could see happening (we have it hardcoded in a lot of places)

bryanoltman:
It works!

packages/shorebird_cli/lib/src/extensions/arg_results.dart
erickzanardo: Did that! I've also thought in a way to encapsulate the signing logic for use, lmk what you think

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
erickzanardo: Created an extension on (String, File?), so it is easier to handle null keys

packages/shorebird_cli/lib/src/common_arguments.dart
```diff
@@ -0,0 +1,6 @@
+/// A class that houses the name of arguments that are shared between different
+/// commands and layers.
+class CommonArguments {
+  static const String publicKeyArgName = 'public-key-path';
+  static const String privateKeyArgName = 'private-key-path';
+}
```
bryanoltman, line 6: I like! I wonder how much further we'll be able to build this out (for example, can we also include descriptions? Accept an arg parser and add standard options to it?)

Not for this PR, just brainstorming.

packages/shorebird_cli/lib/src/extensions/string.dart
```diff
@@ -1,4 +1,21 @@
...
+    final privateKeyFile = this.$2;
+
+    return privateKeyFile != null
+        ? codeSigner.sign(
+            message: this.$1,
+            privateKeyPemFile: privateKeyFile,
+          )
+        : null;
+  }
+}
```
bryanoltman: I don't love this as an extension on a record with two unnamed fields. This isn't saving us much code, so it might be better to inline this.

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -210,12 +213,17 @@ class IosPatcher extends Patcher {
     }
 
     final patchFileSize = patchFile.statSync().size;
+    final privateKeyFile = argResults.file(CommonArguments.privateKeyArgName);
+    final hash = sha256.convert(patchBuildFile.readAsBytesSync()).toString();
+    final hashSignature = (hash, privateKeyFile).signature;
```
felangel:
```suggestion
    final hashSignature = sign(hash, privateKeyFile);
```
minor but it would read a bit better if there was a `sign` method instead of `signature` extension on on a record type.

packages/shorebird_cli/lib/src/common_arguments.dart
felangel: minor but might be nice to have a `static const Options options` on the `CommandRunner` 🤷

packages/shorebird_cli/lib/src/extensions/string.dart
felangel: I agree. Also it's weird that signature returns a nullable string and doesn't do anything if a null file is passed. I think it'd be better to conditionally call `sign` instead and remove this extension.

packages/shorebird_cli/lib/src/common_arguments.dart
erickzanardo: Yeah, I like the idea of making this more robust. I don't it would be that hard, would just take a bit of time to refactor the arguments

felangel (approved):

PR #2149, opened 2024-05-24
feat: adding public key to the patch command

erickzanardo:
## Description

Makes the patch command receive both the private and public key when signing a patch.

DRAFT note: Still need some local test to do
 - Android seems to be working!

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -61,6 +61,11 @@ class IosPatcher extends Patcher {
   ArchiveDiffer get archiveDiffer => IosArchiveDiffer();
 
   @override
+  Future<void> assertArgsAreValid() async {
+    argResults.assertAbsentOrValidKeyPair();
+  }
+
+  @override
   Future<void> assertPreconditions() async {
     try {
       await shorebirdValidator.validatePreconditions(
@@ -100,6 +105,7 @@ class IosPatcher extends Patcher {
           flavor: flavor,
           target: target,
           args: argResults.forwardedArgs,
+          base64PublicKey: argResults.encodedPublicKey,
         );
       } on ProcessException catch (error) {
         buildProgress.fail('Failed to build: ${error.message}');
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -61,11 +61,6 @@ class IosPatcher extends Patcher {
   ArchiveDiffer get archiveDiffer => IosArchiveDiffer();
 
   @override
-  Future<void> assertArgsAreValid() async {
-    argResults.assertAbsentOrValidKeyPair();
-  }
-
-  @override
   Future<void> assertPreconditions() async {
     try {
       await shorebirdValidator.validatePreconditions(
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -219,7 +219,7 @@ class IosPatcher extends Patcher {
     }
 
     final patchFileSize = patchFile.statSync().size;
-    final privateKeyFile = argResults.file(CommonArguments.privateKeyArgName);
+    final privateKeyFile = argResults.file(CommonArguments.privateKeyArg.name);
     final hash = sha256.convert(patchBuildFile.readAsBytesSync()).toString();
     final hashSignature = privateKeyFile != null
         ? codeSigner.sign(
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -95,11 +95,14 @@ of the iOS app that is using this module.''',
         help: 'Validate but do not upload the patch.',
       )
       ..addOption(
-        CommonArguments.privateKeyArgName,
+        CommonArguments.privateKeyArg.name,
         hide: true,
-        help: '''
-The path for a private key file that will be used to sign the patch artifact.
-''',
+        help: CommonArguments.privateKeyArg.description,
+      )
+      ..addOption(
+        CommonArguments.publicKeyArg.name,
+        hide: true,
+        help: CommonArguments.publicKeyArg.description,
       );
   }
 
--- a/packages/shorebird_cli/lib/src/commands/patch/patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patcher.dart
@@ -5,6 +5,7 @@ import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/archive_analysis/archive_differ.dart';
 import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
+import 'package:shorebird_cli/src/extensions/arg_results.dart';
 import 'package:shorebird_cli/src/patch_diff_checker.dart';
 import 'package:shorebird_cli/src/platform/platform.dart';
 import 'package:shorebird_cli/src/release_type.dart';
@@ -67,7 +68,9 @@ https://docs.shorebird.dev/status#link-percentage-ios
   Future<void> assertPreconditions();
 
   /// Asserts that the combination arguments passed to the command are valid.
-  Future<void> assertArgsAreValid() async {}
+  Future<void> assertArgsAreValid() async {
+    argResults.assertAbsentOrValidKeyPair();
+  }
 
   /// Builds the release artifacts for the given platform. Returns the "primary"
   /// artifact for the platform (e.g. the AAB for Android, the IPA for iOS).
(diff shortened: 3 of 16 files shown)


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -60,6 +60,11 @@ class IosPatcher extends Patcher {
   @override
   ArchiveDiffer get archiveDiffer => IosArchiveDiffer();
 
+  @override
+  Future<void> assertArgsAreValid() async {
+    argResults.assertAbsentOrValidKeyPair();
+  }
```
bryanoltman: I think we want to do this in all of our patchers now. This should probably live in `PatchCommand`

packages/shorebird_cli/lib/src/common_arguments.dart
```diff
@@ -1,6 +1,27 @@
+class ArgumentDescriber {
```
bryanoltman, line 4: This could use a brief doc string

packages/shorebird_cli/lib/src/extensions/arg_results.dart
```diff
@@ -49,12 +51,29 @@ extension CodeSign on ArgResults {
...
   void assertAbsentOrValidPublicKey() {
-    file(CommonArguments.publicKeyArgName)?.assertExists();
+    file(CommonArguments.publicKeyArg.name)?.assertExists();
+  }
+
+  void assertAbsentOrValidPrivateKey() {
+    file(CommonArguments.privateKeyArg.name)?.assertExists();
+  }
+
+  void assertAbsentOrValidKeyPair() {
```
bryanoltman: Also could use docs

packages/shorebird_cli/lib/src/extensions/arg_results.dart
```diff
@@ -49,12 +51,29 @@ extension CodeSign on ArgResults {
...
+    final public = wasParsed(CommonArguments.publicKeyArg.name);
+    final private = wasParsed(CommonArguments.privateKeyArg.name);
+
+    if (public == private) {
```
bryanoltman:
```suggestion
    final publicKeyWasParsed = wasParsed(CommonArguments.publicKeyArg.name);
    final publicKeyWasParsed = wasParsed(CommonArguments.privateKeyArg.name);

    if (publicKeyWasParsed == publicKeyWasParsed) {
```

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -113,6 +113,15 @@ void main() {
...
     tearDownAll(restoreExitFunction);
 
+    File createFakeKey(String name) {
+      return File(
+        p.join(
+          Directory.systemTemp.createTempSync().path,
+          name,
+        ),
+      )..createSync();
+    }
```
bryanoltman: This doesn't seem key-specific. Maybe `createTempFile`?

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -95,11 +95,14 @@ of the iOS app that is using this module.''',
...
       ..addOption(
-        CommonArguments.privateKeyArgName,
+        CommonArguments.privateKeyArg.name,
         hide: true,
-        help: '''
-The path for a private key file that will be used to sign the patch artifact.
-''',
+        help: CommonArguments.privateKeyArg.description,
+      )
+      ..addOption(
```
eseidel, line 102: I think you're very quickly going to end up wanting to add CommonArguments flavors to addOption, wasParsed, etc.  I also suspect you'll end up just calling it "Argument" and Arguments.privateKey. But that can all be later patches.

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -159,9 +173,86 @@ void main() {
...
+        test('is valid', () {
+          expect(
+            runWithOverrides(patcher.assertArgsAreValid),
+            completes,
+          );
+        });
       });
+
+      group(
+        'when the private key is provided, it exists, so does the public',
```
bryanoltman:
```suggestion
        'when given existing private and public key files',
```

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -159,9 +173,86 @@ void main() {
...
+            expect(
+              runWithOverrides(patcher.assertArgsAreValid),
+              completes,
+            );
+          });
+        },
+      );
+
+      group(
+        'when the private key is provided, it exists, but not the public',
```
bryanoltman:
```suggestion
        'when given an existing private key and nonexistent public key',
```

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -159,9 +173,86 @@ void main() {
...
+        () {
+          test('fails and logs the err', () async {
```
bryanoltman:
```suggestion
          test('logs error and exits with usage code', () async {
```

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -159,9 +173,86 @@ void main() {
...
+              () => logger.err(
+                'Both public and private keys must be provided or absent.',
+              ),
+            ).called(1);
+          });
+        },
+      );
+
+      group(
+        'when the public key is provided, it exists, but not the private',
```
bryanoltman:
```suggestion
        'when given an existing public key and nonexistent private key',
```

packages/shorebird_cli/test/src/commands/patch/android_patcher_test.dart
```diff
@@ -319,6 +411,38 @@ Looked in:
           });
         });
 
+        group('when the key pair is provided', () {
+          setUp(() {
+            when(() => codeSigner.base64PublicKey(any()))
+                .thenReturn('public_key_encoded');
+          });
+
+          test('calls the buildIpa passing the key', () async {
```
bryanoltman:
```suggestion
          test('calls buildIpa with the provided key', () async {
```

packages/shorebird_cli/test/src/commands/patch/ios_patcher_test.dart
```diff
@@ -100,6 +100,15 @@ void main() {
...
       tearDownAll(restoreExitFunction);
 
+      File createFakeKey(String name) {
+        return File(
+          p.join(
+            Directory.systemTemp.createTempSync().path,
+            name,
+          ),
+        )..createSync();
+      }
```
bryanoltman: Same comment here re: this not being specific to a key. All of this code will need to be written again (twice, once for aar and once for xcframework) if it continues to live in the patchers

bryanoltman (changes requested):

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
erickzanardo: Good call, moved to there!

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
erickzanardo: I am not completely sure I get what you mean about flavor.

But maybe renaming this to `Arguments` instead of `CommonArguments` would be a good call yes, right now it is a bit of a mouthful

packages/shorebird_cli/lib/src/commands/patch/patcher.dart
```diff
@@ -67,7 +68,9 @@ https://docs.shorebird.dev/status#link-percentage-ios
   Future<void> assertPreconditions();
 
   /// Asserts that the combination arguments passed to the command are valid.
-  Future<void> assertArgsAreValid() async {}
+  Future<void> assertArgsAreValid() async {
+    argResults.assertAbsentOrValidKeyPair();
+  }
```
bryanoltman: I might even move this up to the `PatchCommand` layer so Patchers don't need to remember to call `super` (just literally moving `argResults.assertAbsentOrValidKeyPair();` to PatchCommand's `createPatch` method)

packages/shorebird_cli/test/src/commands/patch/ios_patcher_test.dart
```diff
@@ -436,6 +432,41 @@ void main() {
...
+              when(() => codeSigner.base64PublicKey(any()))
+                  .thenReturn('public_key_encoded');
+            });
+
+            test('calls the buildIpa passing the key', () async {
+              when(
+                () => argResults.wasParsed(CommonArguments.publicKeyArg.name),
+              ).thenReturn(true);
+
+              final key = createTempFile('public.der')
```
bryanoltman:
```suggestion
              final key = createTempFile('public.pem')
```

bryanoltman (approved):

PR #2158, opened 2024-05-27
refactor: Improve where key pair are validated in the patch command

erickzanardo:
## Description

Fix a TODO in the code about moving the key pair validation logic from the `patcher` to the `patch_command`. That way patchers doesn't need call super to have the validation happen, which is less prone to errors.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/test/src/commands/patch/patch_command_test.dart
```diff
@@ -280,6 +286,109 @@ void main() {
       expect(command.description, isNotEmpty);
     });
 
+    group('createPatch', () {
+      test('publish the patch', () async {
```
erickzanardo:
```suggestion
      test('publishes the patch', () async {
```

eseidel (approved):
I don't have strong opinions here. Feel free to land things after posting, without review, while others are out. Lgtm

erickzanardo:
> I don't have strong opinions here. Feel free to land things after posting, without review, while others are out. Lgtm

This was an improvement that I discussed with Bryan on friday, but since we were getting to the end of the day, I would not be able to land it in time. In the end it wasn't that complicated of a refactor, I guess I just needed some time not looking at the code haha.

PR #2164, opened 2024-05-29
chore: use raw json for update release and create patch requests

bryanoltman:
## Description

Updates `CreatePatchRequest` and `UpdateReleaseRequest` to use unstructured json for more flexible metadata reporting.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -271,7 +271,7 @@ Please create a release using "shorebird release" and try again.
         releaseId: releaseId,
         platform: platform,
         status: status,
-        metadata: metadata,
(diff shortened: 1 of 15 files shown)


eseidel (approved):
I guess I'm not really sure what this does, if the only producer/consumer of this json is this class.  But I'm also not in any way opposed. :)

bryanoltman:
> if the only producer/consumer of this json is this class

Which class? There were several changed :) `shorebird_code_push_protocol` is a dependency of both our CLI and our backend, so there will also be a _shorebird PR following this.

This is changing "what is acceptable metadata" from a structured object to arbitrary JSON, which will allow the CLI to add/remove/change fields without needing to roll out a corresponding backend update.

PR #2165, opened 2024-05-29
fix(shorebird_cli): don't fail validation when flutter versions differ

### LINKED ISSUE #2136: fix: update flutter version warning in `shorebird doctor`
```
✓ Shorebird is up-to-date (3.1s)
✗ Flutter install is correct (2.1s)
  [!] The version of Flutter that Shorebird includes and the Flutter on your path are different.
        Shorebird Flutter: 3.22.0
        System Flutter:    3.19.5
```

Maybe it should say "the version of Flutter on your path is not the latest Flutter stable"

But probably we should really just fix `shorebird release` to default to whatever version is on your path rather than warning about it.

Maybe then also make `shorebird release` warn (lightly) when using older --flutter-versions that there are newer versions of Flutter that may have more fixes from shorebird.

bryanoltman:
## Description

Updates `shorebird doctor` output to only treat errors as validation failures (validators that return only warnings are now treated as successful).

Before
Screenshot 2024-05-29 at 2 47 25 PM

After
Screenshot 2024-05-29 at 2 47 01 PM

Fixes https://github.com/shorebirdtech/shorebird/issues/2136

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/doctor.dart
+++ b/packages/shorebird_cli/lib/src/doctor.dart
@@ -47,10 +47,10 @@ class Doctor {
       }
 
       final failedFixes = <ValidationIssue, dynamic>{};
-      final progress = logger.progress(validator.description);
+      final validatorProgress = logger.progress(validator.description);
       final issues = await validator.validate();
       if (issues.isEmpty) {
-        progress.complete();
+        validatorProgress.complete();
         continue;
       }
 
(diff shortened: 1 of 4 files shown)


erickzanardo (approved):

eseidel:
lgtm. I suspect the entire message could be deleted and replaced by a warning during `shorebird release`?

e.g.
`The version of `flutter` (3.16.4) on your path is different from the version (3.20.0) used by `shorebird release` by default.  You can specify a version with `--flutter-version` if you desire.`

Although honestly, we probably should just change `shorebird release` to use the path flutter version and similarly log if we think that's what customer intend.  🤷

PR #2169, opened 2024-05-29
feat: supporting gradle acronym prefixed flavors

### LINKED ISSUE #2031: fix: Android flavors starting with uppercase letters are incorrectly parsed
With a gradle file like:

```gradle
productFlavors {
        RJ {
            dimension "flavor-type"
            applicationId "br.com.brasizza.marcus.guia_rj"
            resValue "string", "app_name", "Guia do Rio de janeiro"
        }
        SP {
            dimension "flavor-type"
            applicationId "br.com.brasizza.marcus.guia_sp"
            resValue "string", "app_name", "Guia de São Paulo"
        }
    }
```

`shorebird init` recognized these flavors as

```
✓ 2 product flavors detected: (13.8s)
  - rJ
  - sP
```

See https://discord.com/channels/1030243211995791380/1237476178055397459

erickzanardo:
## Description

When using acronym flavors, like `SP` or `RJ` (Brazilian states, similar to `NY`, etc) on android, `shorebird init` would fail.

This PR "teaches" the cli how to detect them.

Fixes #2031

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/executables/gradlew.dart
```diff
@@ -113,3 +117,16 @@ class Gradlew {
     return productFlavors;
   }
 }
+
+extension on String {
+  bool get isAcronymPrefixed {
```
erickzanardo: This felt like too specific to this command and to gradle names, so I decided to keep them private.

No strong opinions though, happy to change if others think different.

packages/shorebird_cli/test/fixtures/gradle_app_tasks_upper_case_flavors.txt
bryanoltman, line 1: I'd consider creating a `gradle` folder under `fixtures` and putting this file there.

packages/shorebird_cli/lib/src/executables/gradlew.dart
bryanoltman: What is meant by "acronym" here? I worry this isn't as robust as we might want. For example, what behavior would I expect for the following flavors?

- `MyCoolFlavor`
- `NY`
- `NYC`
- `HAL9000`
- `HALNineThousand`

Some tests for this specifically would probably be good.

bryanoltman: Also: docs

packages/shorebird_cli/lib/src/executables/gradlew.dart
```diff
@@ -113,3 +117,16 @@ class Gradlew {
...
+    if (length >= 2) {
+      return this[0].isUpperCase() && this[1].isUpperCase();
+    }
+    return false;
+  }
+
+  bool isUpperCase() {
+    return this == toUpperCase();
+  }
```
bryanoltman: This seems general enough to go in `shorebird_cli/lib/src/extensions/string.dart`

packages/shorebird_cli/lib/src/executables/gradlew.dart
erickzanardo: acronym as in the literal meaning of the word, the combination of the first letters of different words.

I figured that if the first two letter are upper case, we can consider it be a flavor which starts with an acronym, since if we have at least two upper case letters in the beginning of the word, it starts with an acronym, hence the name acronym prefixed.

I will confirm how gradle handles some of the examples you sent there, I know that some of them should work from what I tested, but not all.

I am also open to a different name than acronym, was just the best that I could think of.

packages/shorebird_cli/test/fixtures/gradle_app_tasks_upper_case_flavors.txt
erickzanardo: yeah, I agree, I kept it here because we already have a fixture similar to this one that is used in the original tests, so I thought on keeping them together for consistency.

I am happy to move them both to a folder though

packages/shorebird_cli/lib/src/executables/gradlew.dart
bryanoltman: > acronym as in the literal meaning of the word, the combination of the first letters of different words.

That's what's confusing me. "mRNA" and "laser" are both acronyms, for example. I think something more to the point ("startsWithUpperCaseLetters" or something) would be clearer.

packages/shorebird_cli/test/fixtures/gradle_app_tasks_upper_case_flavors.txt
bryanoltman: I don't feel super strongly, just seemed tidier in a folder ¯\_(ツ)_/¯

packages/shorebird_cli/lib/src/executables/gradlew.dart
erickzanardo: Oh, TIL always thought acronyms were always uppercase

packages/shorebird_cli/lib/src/executables/gradlew.dart
```diff
@@ -113,3 +117,24 @@ class Gradlew {
...
+extension on String {
+  /// Returns true when the string starts with at least two upper case letters
+  ///
+  /// Gradle flavors that are not capital case will not be transformed
+  /// to camel case and should be used as is. So this method helps to identify
+  /// those cases.
+  ///
+  /// Example:
+  /// ```dart
+  /// 'Test'.startsWithUpperCaseLetters; // false
```
bryanoltman, line 144: This one is still a bit surprising to me. Maybe we call this `areFirstTwoLetterUppercase`? I realize it's startsWithUpperCase**Letters** and not startsWithUpperCase**Letter**, but the subtlety is kind of misleading.

packages/shorebird_cli/lib/src/executables/gradlew.dart
```diff
@@ -93,7 +93,11 @@ class Gradlew {
       if (match != null) {
         final variant = match.group(1)!;
         if (!variant.toLowerCase().endsWith('test')) {
-          variants.add(variant[0].toLowerCase() + variant.substring(1));
+          if (variant.startsWithUpperCaseLetters) {
+            variants.add(variant);
+          } else {
+            variants.add(variant[0].toLowerCase() + variant.substring(1));
```
bryanoltman: A small comment explaining what we're doing and why here would be helpful

packages/shorebird_cli/test/src/executables/gradlew_test.dart
bryanoltman, line 1: Just to check my understanding, this is testing that the flavors are extracted exactly as they're written?

bryanoltman (approved):

packages/shorebird_cli/test/src/executables/gradlew_test.dart
erickzanardo: yeah, from all the testing that I did, gradle name transformation for flavors work like this:

`development` -> `development`
`developmentWithAnotherContext` -> `developmentWithAnotherContext`

`Development` -> `development`
`DevelopmentWithAnotherContext` -> `developmentWithAnotherContext`

`QA` -> `QA`
`QAInBrazil` -> `QAInBrazil`
`QAOver9000` -> `QAOver9000`

Which made identify that the pattern is:
 - Flavor starts with at least two capital latter, use as is
 - Otherwise, transform in camel case

So I added a bunch of tests to check on those different scenarios, many of them falls in the use as is rule, there are a couple of what seems to be duplicated tests, but since gradle flavor naming can be a bit confusing, I figured better to have many tests even if they seems redundant

bryanoltman: Yep, makes sense. The disconnect I think I had is that the flavors are defined in files and it wasn't immediately clear whether the expected values were exactly what was in those files or whether they were transformed in some way.

packages/shorebird_cli/lib/src/executables/gradlew.dart
erickzanardo: Yeah, makes sense, changed to the name you suggested!

packages/shorebird_cli/lib/src/executables/gradlew.dart
```diff
@@ -93,7 +93,26 @@ class Gradlew {
       if (match != null) {
         final variant = match.group(1)!;
         if (!variant.toLowerCase().endsWith('test')) {
-          variants.add(variant[0].toLowerCase() + variant.substring(1));
+          // Gradle flavor name transformation seems to work following these two
+          // conditions:
+          //  - Flavor starts with at least two capital latter, use as is
+          //  - Otherwise, transform in camel case
```
bryanoltman:
```suggestion
          // Gradle flavor name transformation seems to work with the following rules:
          //  - If the flavor starts with at least two capital letters, use as is
          //  - Otherwise, transform to camel case
```

PR #2178, opened 2024-05-30
chore: add license_check

eseidel:
This probably doesn't work yet.


eseidel:
Felix did this in another PR.

PR #2189, opened 2024-06-03
feat: adding checksum to artifact updater

### LINKED ISSUE #175: fix: bundletool.jar download is sometimes corrupt (need to verify artifacts)
QQ20240530-102417

erickzanardo:
## Description

Adds a checksum verification on the download of the `bundletool.jar` file.

Fixes https://github.com/shorebirdtech/updater/issues/175

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -5,6 +5,7 @@ import 'package:path/path.dart' as p;
 import 'package:platform/platform.dart';
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/artifact_manager.dart';
+import 'package:shorebird_cli/src/checksum_checker.dart';
 import 'package:shorebird_cli/src/http_client/http_client.dart';
 import 'package:shorebird_cli/src/logger.dart';
 import 'package:shorebird_cli/src/platform.dart';
@@ -133,6 +134,12 @@ abstract class CachedArtifact {
   /// If we fail to fetch it we will exit with an error.
   bool get required => true;
 
+  /// The SHA256 checksum of the artifact binary.
+  ///
+  /// When null, the checksum is not verified and the downloaded artifact
+  /// is assumed to be correct.
+  String? get checksum;
+
   Future<void> extractArtifact(http.ByteStream stream, String outputPath) {
     final file = File(p.join(outputPath, name))..createSync(recursive: true);
     return stream.pipe(file.openWrite());
@@ -140,7 +147,8 @@ abstract class CachedArtifact {
 
   Directory get location => cache.getArtifactDirectory(name);
 
-  Future<bool> isUpToDate() async => location.existsSync();
+  Future<bool> isUpToDate() async =>
+      File(p.join(location.path, name)).existsSync();
 
   Future<void> update() async {
     final request = http.Request('GET', Uri.parse(storageUrl));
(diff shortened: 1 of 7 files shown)


packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -171,6 +178,25 @@ allowed to access $storageUrl.''',
...
 
+    final expectedChecksum = checksum;
+    if (expectedChecksum != null) {
+      final artifactFile = File(p.join(location.path, name));
+
+      if (!checksumChecker.checkFile(artifactFile, expectedChecksum)) {
+        // Delete the location, so if the download is retried, it will be
+        // re-downloaded.
+        location.deleteSync(recursive: true);
+        // TODO(erickzanardo): Automatically retry the download.
```
erickzanardo, line 189: Thought on doing this in a follow up for easier to review PRs

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -296,4 +331,8 @@ class BundleToolArtifact extends CachedArtifact {
   String get storageUrl {
     return 'https://github.com/google/bundletool/releases/download/1.15.6/bundletool-all-1.15.6.jar';
   }
+
+  @override
+  String? get checksum =>
+      '''38ae8a10bcdacef07ecce8211188c5c92b376be96da38ff3ee1f2cf4895b2cb8''';
```
erickzanardo, line 343: I generated this manually, unfortunately Github Releases (where the bundle tool is hosted) does not offer an automatic checksum feature. Since we use hard coded version url, I can't think on any problem in this approach, but let me know if I am missing something.

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -205,6 +231,9 @@ class AotToolsDillArtifact extends CachedArtifact {
   @override
   String get storageUrl =>
       '${cache.storageBaseUrl}/${cache.storageBucket}/shorebird/${shorebirdEnv.shorebirdEngineRevision}/$name';
+
+  @override
+  String? get checksum => null;
```
erickzanardo, line 236: Should we add the checksum to the other artifacts in follow up?

eseidel: I'm not sure?  For artifacts that are tied to a Flutter version (rather than a Shorebird version) we would have to have them in the manifest.yaml on storage rather than checked into the code.

packages/shorebird_cli/lib/src/cache.dart
eseidel: Yeah, that's fine.  This is more useful as a comment in the source than a comment on the PR though. :)

packages/shorebird_cli/lib/src/cache.dart
eseidel: You probably want to do the retry at a higher level (the thing that call this), but yeah.

eseidel:
I think this is a good start.

We might end up pulling all the artifacts out into some sort of artifacts.yaml or something and have their checksums there, to have all the urls and checksums in one place (similar to a pubspec.lock), but at least giving us the capability is a good start.

eseidel (approved):

packages/shorebird_cli/lib/src/cache.dart
erickzanardo: But then, don't we run the risk of the manifest file getting corrupted as well?

packages/shorebird_cli/lib/src/cache.dart
erickzanardo: For sure, the comment was just to get a discussion going :). Gonna put a comment.

PR #2191, opened 2024-06-03
feat: adding a retry logic to the cache of the executables

erickzanardo:
## Description

Adds a retry logic to the executables update and caching in the CLI.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -64,7 +64,7 @@ class Cache {
         maxAttempts: 3,
         onRetry: (e) {
           logger
-            ..err('Failed to update ${artifact.name}, retrying...')
+            ..detail('Failed to update ${artifact.name}, retrying...')
             ..detail(e.toString());
         },
       );
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -3,6 +3,7 @@ import 'dart:io' hide Platform;
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as p;
 import 'package:platform/platform.dart';
+import 'package:retry/retry.dart';
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/artifact_manager.dart';
 import 'package:shorebird_cli/src/checksum_checker.dart';
@@ -52,30 +53,21 @@ class Cache {
 
   void registerArtifact(CachedArtifact artifact) => _artifacts.add(artifact);
 
-  static const _maxRetryAttempts = 3;
-
   Future<void> updateAll() async {
     for (final artifact in _artifacts) {
       if (await artifact.isUpToDate()) {
         continue;
       }
 
-      var retryAttempts = 0;
-      while (retryAttempts < _maxRetryAttempts) {
-        try {
-          await artifact.update();
-          break;
-        } catch (e) {
-          retryAttempts++;
-          if (retryAttempts < _maxRetryAttempts) {
-            logger
-              ..err('Failed to update ${artifact.name}, retrying...')
-              ..detail(e.toString());
-          } else {
-            rethrow;
-          }
-        }
-      }
+      await retry(
+        artifact.update,
+        maxAttempts: 3,
+        onRetry: (e) {
+          logger
+            ..err('Failed to update ${artifact.name}, retrying...')
+            ..detail(e.toString());
+        },
+      );
     }
   }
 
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -52,13 +52,30 @@ class Cache {
...
       }
 
-      await artifact.update();
+      var attempts = 0;
+      while (attempts < _maxAttempts) {
+        try {
+          await artifact.update();
+          break;
+        } catch (e) {
+          if (attempts + 1 < _maxAttempts) {
```
eseidel: Could just init to 1 instead of adding +1 here 🤷

eseidel (approved):

bryanoltman (changes requested):
Our http client already has retry logic https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/http_client/http_client.dart#L9-L14

bryanoltman (approved):
Approving as I'm now seeing that `update` can fail for not http-related reasons

packages/shorebird_cli/test/src/cache_test.dart
```diff
@@ -198,20 +198,36 @@ void main() {
...
+                ),
               ),
-            ),
-          );
+            );
+          });
+
+          test('retries and log', () async {
+            const exception = SocketException('test');
+            when(() => httpClient.send(any())).thenThrow(exception);
```
bryanoltman, line 219: fwiw, our httpClient already retries. https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/http_client/http_client.dart#L9-L14

This is probably more likely to address the case where the downloaded file's hash doesn't match what we expect

erickzanardo:
> Our http client already has retry logic https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/http_client/http_client.dart#L9-L14

Hmm, interesting. But this logic will also retry if we have a checksum check failure, which would not be taken into account by the http client.

I wonder how to "merge" both retry logic though. Any thoughts?

bryanoltman:
I'd be a little surprised if the httpClient didn't perform some validation of the bits it was downloading, but I'm honestly not sure. I suspect the file corruption is more likely to happen after a file download has finished (probably in between shorebird command runs) or because the shorebird command that was downloading the file was killed mid-download. If the latter, it might make sense to download the files to a temp dir, verify the checksum, and then move them into the expected location.

eseidel:
I really don't have a good sense of how bundletool.jar is failing.  Or if we're even corrupting the file?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -52,13 +52,30 @@ class Cache {
 
   void registerArtifact(CachedArtifact artifact) => _artifacts.add(artifact);
 
+  static const _maxAttempts = 3;
```
felangel:
```suggestion
  static const _maxRetryAttempts = 3;
```
nit

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -52,13 +52,30 @@ class Cache {
...
+
   Future<void> updateAll() async {
     for (final artifact in _artifacts) {
       if (await artifact.isUpToDate()) {
         continue;
       }
 
-      await artifact.update();
+      var attempts = 0;
```
felangel:
```suggestion
      var retryAttempts = 0;
```

felangel: Also I believe we already depend on `package:retry` so it might make sense to use that here as well. It would clean this up a bit imo.

```dart
await retry(artifact.update, maxAttempts: 3);
```

felangel (changes requested):
See comments about using `package:retry` here for consistency and to simplify the implementation.

packages/shorebird_cli/lib/src/cache.dart
erickzanardo: @felangel we don't depend on it, at least not on the CLI, but it seems interesting, should we add the dep or wouldn't it be too much to add a dep just for that?

erickzanardo: I went ahead and refactored to use the package, it cleared the code so much!

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -58,7 +59,15 @@ class Cache {
         continue;
       }
 
-      await artifact.update();
+      await retry(
+        artifact.update,
+        maxAttempts: 3,
+        onRetry: (e) {
+          logger
+            ..err('Failed to update ${artifact.name}, retrying...')
```
bryanoltman: We probably want this to remain detail, no? If subsequent attempts succeed, I worry this will only confuse users.

bryanoltman (approved):

felangel (approved):

PR #2196, opened 2024-06-03
chore: move codecov.yml into .github directory

eseidel:
It's not required to be at the root anymore:
https://docs.codecov.com/docs/codecov-yaml#can-i-name-the-file-codecovyml


felangel (approved):

felangel:
Should we do the same for the other one?

eseidel:
Sure.  🤷🏼‍♂️ I just saw it and remembered it could be inside GitHub and
moved it to clean up the top level a little.

PR #2199, opened 2024-06-04
feat: making sure debug data is created on ios patch even on an interrupt 

erickzanardo:
## Description

This PR adds logic to make sure that debug files are created on path ios command, even when the linking fails.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -304,17 +307,42 @@ class IosPatcher extends Patcher {
...
+    //
+    // So, to at least be able to get the logging data that were already
+    // created up until the point that the user interrupted the command, we
+    // hook this process to the interrupt signals and do nothing other than
+    // log a message.
+    //
+    // The interrupt signal will then propagate to the link command and fail,
+    // triggering the catch block which will make sure that the data is properly
+    // consolidated and warning in the console.
+    final interruptionSubscription = process.interrupts.listen((_) {
```
eseidel: This is listening to interrupts on the child process.  Are you sure that's what you want?

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -304,17 +307,42 @@ class IosPatcher extends Patcher {
...
-          : null;
+    final dumpDebugInfoDir = await aotTools.isLinkDebugInfoSupported()
+        ? Directory.systemTemp.createTempSync()
+        : null;
 
-      if (dumpDebugInfoDir != null) {
-        logger.detail(
-          'Dumping link debug info to ${dumpDebugInfoDir.path}',
-        );
+    // The linking process can take a while and there might be cases where it
```
felangel: Have you been able to reproduce the issue? We might want to hold off on merging anymore debugging changes until we can reproduce the issue. Also we just rolled Dart which may have resolved the issue for affected customers so we might want to double check and see if folks are still running into this with the latest version.

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
erickzanardo: I am not completely sure I understand the distinction here on `child process` or not. But from all the testing that I did, I would say yes. The last paragraph in the comment above tries to explain the behaviour that I identified, and that made it possible to make sure that the debug files are created even on an interrupt

eseidel:
I think we should just put this down for now and come back later with more information.

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
erickzanardo: > Have you been able to reproduce the issue?

I have not, all the things that I tested resulted in the link command failing, and the patch command would fail in the sequence, I didn't got to make it hang.

> We might want to hold off on merging anymore debugging changes until we can reproduce the issue.

We could hold off, but I see value in this PR nevertheless, before it, our debug files would be consolidated only on a linking success, this PR makes it happen when there is also a failing or an interrupt. So even if the issue is not reproducible, I still see value in the changes. Let me know if this makes sense

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -334,15 +343,10 @@ class IosPatcher extends Patcher {
...
       );
 
-      if (dumpDebugInfoDir != null) {
-        final debugInfoZip = await dumpDebugInfoDir.zipToTempFile();
-        debugInfoZip.copySync(p.join('build', debugInfoFile.path));
-        logger.detail(
-          'Link debug info saved to ${debugInfoFile.path}',
-        );
-      }
+      await consolidateDebugInfo();
```
felangel: can we move this to a `finally` block since we want it to happen regardless of success or failure?

erickzanardo:
@felangel @eseidel like we discussed on VC, I left only the consolidation on failure change and removed the interruption handling.

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -312,17 +314,24 @@ class IosPatcher extends Patcher {
...
-      if (dumpDebugInfoDir != null) {
-        logger.detail(
-          'Dumping link debug info to ${dumpDebugInfoDir.path}',
-        );
+    Future<void> consolidateDebugInfo({bool warn = false}) async {
+      if (dumpDebugInfoDir == null) return;
+
+      final debugInfoZip = await dumpDebugInfoDir.zipToTempFile();
+      debugInfoZip.copySync(p.join('build', debugInfoFile.path));
+      final msg = 'Link debug info saved to ${debugInfoFile.path}';
```
felangel: I think it's fine to just always log this as a detail since we always put the debug info in the same relative location.

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -312,17 +314,24 @@ class IosPatcher extends Patcher {
...
-          : null;
+    final dumpDebugInfoDir = await aotTools.isLinkDebugInfoSupported()
+        ? Directory.systemTemp.createTempSync()
+        : null;
 
-      if (dumpDebugInfoDir != null) {
-        logger.detail(
-          'Dumping link debug info to ${dumpDebugInfoDir.path}',
-        );
+    Future<void> consolidateDebugInfo({bool warn = false}) async {
```
felangel:
```suggestion
    Future<void> dumpDebugInfo() async {
```
nit

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
erickzanardo: Just wanted to make it more explicit to the user that the debug data exists, so it would probably save us support time explaining where those files are created.

But I can change it to detail, no strong opinion

eseidel (approved):

felangel (approved):

PR #2204, opened 2024-06-05
chore(shorebird_cli): rename build to build-internal

bryanoltman:
## Description

Renames the `build` command to `build-internal` because enough customers were finding it and using it instead of `shorebird release`.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/build/build_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/build/build_command.dart
@@ -15,7 +15,10 @@ class BuildCommand extends ShorebirdCommand {
   }
 
   @override
-  String get description => 'Build a new release of your application.';
+  String get description => '''
+Build a new release of your application.
+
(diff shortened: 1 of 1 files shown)


packages/shorebird_cli/lib/src/commands/build/build_command.dart
```diff
@@ -18,7 +18,7 @@ class BuildCommand extends ShorebirdCommand {
   String get description => 'Build a new release of your application.';
```
eseidel: Might want to explain what it's for, or that it does not save the release build for later patching, etc.

eseidel (approved):

packages/shorebird_cli/lib/src/commands/build/build_command.dart
bryanoltman: Done

PR #2224, opened 2024-06-12
feat: Adding override min Flutter version flag to patch and release

### LINKED ISSUE #2222: Add a --allow-broken-flutter or similar to disable flutter version check.
A customer has (understandably) asked to be able to choose if they wish to update their existing iOS apps using Shorebird, even though we now know all previous releases made with Shorebird have https://github.com/shorebirdtech/shorebird/issues/1986

erickzanardo:
## Description

Adds a new flag `override-min-flutter-version` to the patch and release command, that when informed, will allow users to release and or patch on iOS even on non supported flutter versions by shorebird.

This PR also adds some measures to guide the user to avoid a "foot gun" scenario.

 - Make it explicit that doing so is dangerous and might produce unwanted behaviours
 - Only allow patches with unsupported flutter versions to be published on staging, which hopefully will encourage users to test the patch before releasing it to their users.

Still in draft to gather feedback on the measures and messages, will add tests once we are settled in a path to move forward.

If merged,
fixes #2222 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -65,6 +65,25 @@ class IosPatcher extends Patcher {
   ArchiveDiffer get archiveDiffer => IosArchiveDiffer();
 
   @override
+  Future<void> assertArgsAreValid() async {
+    await super.assertArgsAreValid();
+
+    final isStaging = argResults['staging'] == true;
+    final overridingMinFlutterVersion =
+        argResults[CommonArguments.overrideMinFlutterVersionArg.name] == true;
+    if (overridingMinFlutterVersion && !isStaging) {
+      logger.err(
+        '''
+When overriding min Flutter Version, patches must be staged first.
+
+If you are really sure you want to proceed, add the --staging flag to the
+command and try again.''',
+      );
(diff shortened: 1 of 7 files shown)


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -97,12 +116,20 @@ class IosPatcher extends Patcher {
...
+        if (!overrideMinFlutterVersion) {
+          logger.err(
+            '''
 iOS patches are not supported with Flutter versions older than $minimumSupportedIosFlutterVersion.
-For more information see: $supportedVersionsLink''',
-        );
-        exit(ExitCode.software.code);
+For more information see: $supportedVersionsLink
+
+If you are certain about the compatibility of your Flutter version, you can bypass this check by passing the --override-min-flutter-version flag. Patches generated with the override flag must be sent to staging first, so make sure to also pass the --staging flag alongside.
```
felangel, line 128: I don’t think anyone can be certain given how these issues happen at the linker and compiler level 🙁

eseidel:
Yeah, I'm going to rescind my suggestion and apologize to customers and move on.

PR #2228, opened 2024-06-12
feat: making sure we print the error log file on a command failure

erickzanardo:
## Description

Some time ago, we implemented some handling in the CLI to forward all of the output to a log file print the following message if a CLI command failed and the logging level was non verbose:

```

/Users/erick/Library/Application Support/shorebird/logs/1718217148336_shorebird.log
```

But that noticed was only being printed when a command would return an `ExitCode`, or throw an error. In many commands, instead of return an error code, or throwing an error, `exit` would be directly called, effectively interrupting the command, and not printing the message.

This PR uses the same code infrastructure used in tests to override the exit function, so we can now be sure that any exit call made in a sub command or process will still print the message, which I believe will lead users to open issues with more information, making it easier for us to investigate eventual bugs.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
@@ -14,6 +14,7 @@ import 'package:shorebird_cli/src/shorebird_env.dart';
 import 'package:shorebird_cli/src/shorebird_flutter.dart';
 import 'package:shorebird_cli/src/shorebird_process.dart';
 import 'package:shorebird_cli/src/shorebird_version.dart';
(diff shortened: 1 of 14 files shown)


felangel:
> But that noticed was only being printed when a command would return an ExitCode, or throw an error. In many commands, instead of return an error code, or throwing an error, exit would be directly called, effectively interrupting the command, and not printing the message.

Where does this happen? Seems like that should be fixed instead or overriding exit

erickzanardo:
Closing in favor of #2230

PR #2230, opened 2024-06-12
feat(shorebird_cli): remove direct `exit` calls

erickzanardo:
## Description

Some time ago, we implemented some handling in the CLI to forward all of the output to a log file print the following message if a CLI command failed and the logging level was non verbose:

```

/Users/erick/Library/Application Support/shorebird/logs/1718217148336_shorebird.log
```

But that noticed was only being printed when a command would return an `ExitCode`, or throw an error. In many commands, instead of return an error code, or throwing an error, `exit` would be directly called, effectively interrupting the command, and not printing the message.

This PR replaces all the `exit` calls, to `throw ProcessExit`, which will be caught by the try/catch in the command runner and properly handle the error with the message.

Additional, the test helper that overrides the `exit` call now will throw a different exception, that will make sure that (or at least make it less likely) for `exit` calls be called inside our commands and methods.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -147,7 +147,7 @@ For more information see: $supportedVersionsLink''',
 
       buildProgress.complete();
     } catch (_) {
-      return exit(ExitCode.software.code);
+      return throw ProcessExit(ExitCode.software.code);
```
felangel: The return shouldn't be necessary if we're throwing right?

packages/shorebird_cli/lib/src/third_party/flutter_tools/lib/src/base/io.dart
```diff
@@ -71,6 +70,12 @@ ExitFunction get exit {
 bool _inUnitTest() {
   return Zone.current[#test.declarer] != null;
 }
+
+/// Thrown when [exit] is called unexpectedly in a test.
+class UnexpectExitCalled implements Exception {
+  /// Creates a new [UnexpectExitCalled] exception.
+  const UnexpectExitCalled();
+}
```
bryanoltman: I'm not sure this (or `setExitFunctionForTests`) is necessary now that we aren't using `exit`

bryanoltman:
I like the idea here, but I expect this will be a much more involved change. As is, I worry it won't work the way we expect—we're introducing new code paths for every failure state that previously expected an `exit`. Also, `setExitFunctionForTests` is intended to override `exit`, and if we aren't using exit, I think that disappears. We'll need to be sure we exit gracefully anywhere exit was called (which I believe is mostly `CodePushClientWrapper`, but also the Patchers and Releasers as well). I'd want to verify that throwing instead of exiting doesn't trigger any unexpected catch code in the commands/functions that currently expects its dependencies to exit on failure.

For historical context, the reason we started doing this was because we'd end up needing to handle the same errors in the same ways for every command that invoked shared functionality.

packages/shorebird_cli/lib/src/third_party/flutter_tools/lib/src/base/io.dart
```diff
@@ -79,7 +84,7 @@ bool _inUnitTest() {
 void setExitFunctionForTests([ExitFunction? exitFunction]) {
   _exitFunction = exitFunction ??
       (int exitCode) {
-        throw ProcessExit(exitCode, immediate: true);
+        throw const UnexpectExitCalled();
```
felangel: What is `UnexpectExitCalled()` used for? Seems like we should just be able to throw a `UnsupportedError` instead if the goal is just to ensure it never happens and fails the tests.

felangel:
```suggestion
        throw const UnsupportedError();
```

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
erickzanardo: Probably not! I did a full replace, didn't noticed this case. Gonna change

packages/shorebird_cli/lib/src/third_party/flutter_tools/lib/src/base/io.dart
erickzanardo: My goal with this, is that we add this call to the tests, this will make sure we (or other devs) don't unintentionally call `exit`.

I don't mind removing if you think I am being too overcautious here

packages/shorebird_cli/lib/src/third_party/flutter_tools/lib/src/base/io.dart
erickzanardo: Yeah, `UnsupportedError` fits nicely too, I forgot about that one

packages/shorebird_cli/lib/src/third_party/flutter_tools/lib/src/base/io.dart
bryanoltman: The tests just exit early (and fail) if exit is called, so I think it would probably be fine to remove.

erickzanardo: hum, yeah, you are actually right! Gonna remove this whole file.

erickzanardo:
> We'll need to be sure we exit gracefully anywhere exit was called (which I believe is mostly CodePushClientWrapper, but also the Patchers and Releasers as well). I'd want to verify that throwing instead of exiting doesn't trigger any unexpected catch code in the commands/functions that currently expects its dependencies to exit on failure.

@bryanoltman maybe this is a good time to add some more complex tests on the CLI? Most tests that we have right now are unit tests, which calls directly a method from the command class. Maybe we should add some "integration tests", that will instead run the command runner from its main entry point, and assert on the stdout/stderr/exit code?

That way we would be able to check on this concerns that you listed. Wdyt?

bryanoltman:
We have e2e tests that do this in `packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart`, unless you meant something different?

bryanoltman:
More tests is never a bad thing, but generally, you want to ensure coverage is present before making a code change. For this, I think just smoke testing a bit (probably through forcing a ProcessExit to be thrown when `exit` would have otherwise been called in a few places) might be sufficient.

erickzanardo:
> We have e2e tests that do this in `packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart`, unless you meant something different?

Oh that is great! This seems to be a bit more complete than what I was suggesting, it seems to even connect to backend. My idea was to mock the backend, so we would have more control over the responses, for example, we could return an error that code push client would catch and we would be able to assert that the cli exited gracefully.

But maybe I could just use this same folder/structure and add some more tests where I mock the result instead?

bryanoltman:
I think it'd probably be good to separate black-box tests (shorebird_cli_integration_test.dart) from [gray-box](https://en.wikipedia.org/wiki/Gray-box_testing) tests, but I'm supportive of adding that. Doesn't have to happen as part of this PR.

bryanoltman (approved):
LGTM as long as you've verified that the failures that were previously exits are handled gracefully.

erickzanardo:
Tested throwing the exception in different places, like on the releaser, the patcher, code push client wrapper and the cli handled the exit correctly! 

Merging 🚀

PR #2235, opened 2024-06-13
fix: cache clear command was breaking

### LINKED ISSUE #2233: fix:  shorebird cache clean  fails with log file not found
**Description**

When trying to run `shorebird cache clean` we get the following error:

```
✗ Failed to delete cache directory /Users/erick/projects/shorebird/shorebird/bin/cache: FileSystemException: Deletion failed, path = '/Users/erick/Library/Application Support/shorebird/logs' (OS Error: Directory not empty, errno = 66) (0.8s)
```

This is mostly likely happening because we delete the log file in this command, but tries to log to it to show the completed progress.

Also reported at https://github.com/shorebirdtech/shorebird/issues/2231

**Steps To Reproduce**

1. run `shorebird cache clean`
2. see erro

**Expected Behavior**

`shorebird cache clean` should no fail with error

erickzanardo:
## Description

The cache clean command would fail because it were deleting the log file, and trying to log right after.

Fixes #2233 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/integration_test/helpers.dart
+++ b/packages/shorebird_cli/integration_test/helpers.dart
@@ -1,0 +1,20 @@
+import 'dart:io';
+
+import 'package:mason_logger/mason_logger.dart';
+
+ProcessResult runCommand(
+  String command, {
+  required String workingDirectory,
+  Logger? logger,
+}) {
+  final parts = command.split(' ');
+  final executable = parts.first;
+  final arguments = parts.skip(1).toList();
+  (logger ?? Logger()).info('running $command in $workingDirectory');
+  return Process.runSync(
+    executable,
+    arguments,
+    runInShell: true,
+    workingDirectory: workingDirectory,
+  );
+}
--- a/packages/shorebird_cli/integration_test/helpers.dart
+++ b/packages/shorebird_cli/integration_test/helpers.dart
@@ -2,6 +2,10 @@ import 'dart:io';
 
 import 'package:mason_logger/mason_logger.dart';
 
+/// Helper function to run a command in the shell, meant to be used in tests.
+///
+/// It will take a command string, like `shorebird --version`, run it in the
+/// shell, and return the result.
 ProcessResult runCommand(
   String command, {
   required String workingDirectory,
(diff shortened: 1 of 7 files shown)


packages/shorebird_cli/lib/src/logger.dart
```diff
@@ -42,6 +42,11 @@ void _writeToLogFile(Object? message, {required File logFile}) {
     return;
   }
 
+  // Making sure the log file exists before writing to it.
+  // This is necessary because the log file may be deleted by a cache
+  // clear command.
+  if (!logFile.existsSync()) return;
```
erickzanardo: My reasoning on this: The file should always be there, since it is created upon initialization. So if it is not there, that means that we shouldn't log, i.g. on the cache clear commando I don't think we should recreate the file that we just created.

Lmk if you all think that this makes sense or if you can think in a better approach.

packages/shorebird_cli/integration_test/helpers.dart
```diff
@@ -0,0 +1,20 @@
+import 'dart:io';
+
+import 'package:mason_logger/mason_logger.dart';
+
+ProcessResult runCommand(
```
bryanoltman, line 9: Docs for this function (usage, what it does beyond invoking `Process.run`) would be helpful

packages/shorebird_cli/lib/src/logger.dart
bryanoltman: I don't know that I'm convinced that `cache clear` should delete logs. Maybe that's a cleaner solution? I worry this might put us in a silent failure scenario.

erickzanardo: humm, yeah, that is a good argument! logs are not cache! I am gonna instead remove the log clearing call from the command then!

erickzanardo: Updated, lmk wdyt!

bryanoltman (approved):

Issue #2240, opened 2024-06-14
`--release-version` doesn't work for `shorebird release android` or `shorebird release ios`


bryanoltman:
The `--release-version` option exists for flutter modules (hybrid/add-to-app targets), which are not intrinsically associated with a specific version number at build time (they are embedded in an app after the release command completes). 

`--release-version` does nothing for full-Flutter-app releases. In the short term, we should error if a user tries to use this command with `release ios` or `release android`, but in the longer term, it would be good to decouple release versions from the version reported by the app, which would help to solve https://github.com/shorebirdtech/shorebird/issues/793.

bryanoltman:
Closing as addressed by https://github.com/shorebirdtech/shorebird/pull/2241, even though I think the better long-term fix would be to allow users to specify a release version that is different than what the app reports (and probably falling back to the current behavior if none is provided).

Issue #793, opened 2023-07-09
feat: Add a single AAR to multiple android apps (library_version)


julioamadortejeda:
**Description**

I'm working on an AAR module that can be used in multiple Android applications (all different).
My idea is to create a release of my AAR that is embedded within the android apps but, if I patch my AAR, that patch will work in all the android apps where my AAR is.
 My problem is that every android app has a
VersionName and VersionCode are different from each other, so when I want to install the patch I get the error "updater::c_api: Error downloading update: Invalid state: Hash mismatch. This is most often caused by using the same version number with a different app binary ".
I understand the error but I need the patch to be able to be applied regardless of whether the binary or the hashes are different since I am not the owner of the android apps and I cannot create a release for each app.

**Requirements**

I would like to know if there is a way to create an AAR release without it being tied to a versionname + versioncode?

**Additional Context**

![discord](https://github.com/shorebirdtech/shorebird/assets/36451076/52f3745c-4938-4432-8863-eeb9635915e6)

felangel:
Hi @julioamadortejeda 👋 
Thanks for opening an issue!

I don't think we want to disable the hash check because it would likely lead to patches that are incompatible and result in corrupt subsequent boot-ups.

> My problem is that every android app has a
VersionName and VersionCode are different from each other, so when I want to install the patch I get the error "updater::c_api: Error downloading update: Invalid state: Hash mismatch. This is most often caused by using the same version number with a different app binary ".

Would it be possible for you to generate a patch for each release that you'd like to patch? This will ensure that the changes in the patch apply cleanly to the underlying app release.

bryanoltman:
Would it be possible to use flavors to target the different apps? I would need to know more about how you're distributing the AAR to the folks who develop/release the Android app that uses the AAR, but flavors should allow Shorebird to recognize each separate Android app the AAR is embedded in as a separate app.

bryanoltman:
To address this question specifically:

> I would like to know if there is a way to create an AAR release without it being tied to a versionname + versioncode?

Not currently. Version name+code is how Shorebird identifies the version of the app.

julioamadortejeda:
> Hi @julioamadortejeda 👋 Thanks for opening an issue!
> 
> I don't think we want to disable the hash check because it would likely lead to patches that are incompatible and result in corrupt subsequent boot-ups.
> 
> > My problem is that every android app has a
[...]

PR #2241, opened 2024-06-14
fix(shorebird_cli): don't allow --release-version for ios and android releases

bryanoltman:
## Description

A quick fix for https://github.com/shorebirdtech/shorebird/issues/2240. When `--release-version` is provided to `shorebird release ios` or `shorebird release android`, print a usage error and exit. 

This flag is only useful for `aar` and `ios-framework` release targets and does nothing for `ios` or `android` releases. Because it's a valid option to the `release` command, it is parsed and the user has no idea that it isn't working.

A better long-term fix would be addressing https://github.com/shorebirdtech/shorebird/issues/793, which would decouple release versions from the actual app version.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/ios_releaser.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/ios_releaser.dart
@@ -42,6 +42,17 @@ class IosReleaser extends Releaser {
   @override
   Future<void> assertArgsAreValid() async {
     argResults.assertAbsentOrValidPublicKey();
+
+    if (argResults.wasParsed('release-version')) {
+      logger.err(
+        '''
+The "--release-version" flag is only supported for aar and ios-framework releases.
+        
+To change the version of this release, change your app's version in your pubspec.yaml.''',
+      );
+      throw ProcessExit(ExitCode.usage.code);
+    }
+
     if (argResults.rest.contains('--obfuscate')) {
       // Obfuscated releases break patching, so we don't support them.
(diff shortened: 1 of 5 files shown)


packages/shorebird_cli/lib/src/commands/release/ios_releaser.dart
```diff
@@ -42,6 +42,17 @@ class IosReleaser extends Releaser {
   @override
   Future<void> assertArgsAreValid() async {
     argResults.assertAbsentOrValidPublicKey();
+
+    if (argResults.wasParsed('release-version')) {
+      logger.err(
+        '''
```
erickzanardo, line 48: Should this text go in the CommandArguments for the sake of dry?

Maybe adding a new field to the CommonArgument to store it, like `wrongUsageHelp` (or similar).

Then on both places you could just `logger.err(CommonArguments.releaseVersion.wrongUsageHelp)`.

erickzanardo (approved):
Left one suggestion, but LGTM either way!

packages/shorebird_cli/lib/src/commands/release/ios_releaser.dart
bryanoltman: I think the instinct is good, but both the help message and error messages are command-specific.

PR #2247, opened 2024-06-14
fix: gracefully handling patching on a app with no releases

### LINKED ISSUE #2242: `shorebird patch` fails awkwardly if there are no releases
**Description**

We should detect that there are no releases before asking which release they would like to patch.

**Steps To Reproduce**

1. Create a new shorebird app
2. Run `shorebird patch ios`
3. See the following

```
⑆ shorebird patch ios
✓ Fetching apps (0.2s)
✓ Fetching releases (0.1s)
Which release would you like to patch?
Which release would you like to patch? RangeError (index): Invalid value: Valid value range is empty: 0

You can also file an issue if you think this is a bug. Please include the following log file in your report:
/Users/bryanoltman/Library/Application Support/shorebird/logs/1718375682768_shorebird.log
```

erickzanardo:
## Description

`shorebird patch` would fail with a weird message if executed in an app with no releases.

This PR tries to gracefully handle that by displaying a warning and exiting the command.

Fixes #2242

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -313,7 +313,7 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
       logger.warn(
         '''No releases found for app $appId. You need to make first a release before you can create a patch.''',
       );
-      throw ProcessExit(ExitCode.success.code);
+      throw ProcessExit(ExitCode.usage.code);
     }
 
     return logger.chooseOne<Release>(
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -308,6 +308,14 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
     final releases = await codePushClientWrapper.getReleases(
       appId: appId,
     );
+
+    if (releases.isEmpty) {
+      logger.warn(
+        '''No releases found for app $appId. You need to make first a release before you can create a patch.''',
+      );
+      throw ProcessExit(ExitCode.success.code);
+    }
+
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -308,6 +308,14 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
     final releases = await codePushClientWrapper.getReleases(
       appId: appId,
     );
+
+    if (releases.isEmpty) {
+      logger.warn(
+        '''No releases found for app $appId. You need to make first a release before you can create a patch.''',
+      );
+      throw ProcessExit(ExitCode.success.code);
```
erickzanardo: @bryanoltman I was on the fence when choosing the code here, I ended up returning success because this is an expected error, but I wonder if `usage` would be more fitting, since in a way, the command was used wrongly?

bryanoltman: `usage` sounds like a better option.

packages/shorebird_cli/test/src/commands/patch/patch_command_test.dart
```diff
@@ -678,6 +678,26 @@ void main() {
         );
       });
 
+      group('when prompting for releases, but there is none', () {
+        setUp(() {
+          when(
+            () => codePushClientWrapper.getReleases(appId: any(named: 'appId')),
+          ).thenAnswer((_) async => []);
+        });
+        test('warns and exits', () async {
```
bryanoltman:
```suggestion
          ).thenAnswer((_) async => []);
        }); 

        test('warns and exits', () async {
```

bryanoltman (approved):
LGTM with comments

PR #2248, opened 2024-06-14
fix: shorebird release deletes older preview cache

### LINKED ISSUE #1746: Build does not update if you delete a release and then upload new one with same version number
App ID: 130984c0-a89c-44a0-972b-f29f9725b216

**Description**

I have a shorebird release for android with version number 1.1.55+73 which is good for use. So far, so good. 

However, i am having some platform specific difficulty with my IOS version of that release. Because i want my Android and IOS build numbers to be consistent, my working flow whilst fixing/investigating my IOS issues is to 

1. delete the IOS build from that release (which i want to keep because the android version works)
2. create a fixed IOS build with the same version/release number

However, when i do this.. the edits to the IOS build in subsequent submissions are NOT reflected when i run previews.

For example, i added some print logs to help me debug, but when i run preview after i released those changes, i was still seeing the OLD print logs when i run preview. 

So it's as if the old original build i thought i had deleted is actually retained and never overwritten

**Steps To Reproduce**

1. Create a release on both Android and IOS
2. From the shorebird console, delete the IOS version of that release
3. Change some code and create a NEW IOS release with the same version number 

**Expected Behavior**

I would expect the code changes to be reflected when i run preview

erickzanardo:
## Description

When running a preview, the CLI will always use the cache artifact if one if found. This might make the command run an outdated artifact when a release is delete by some reason.

This command implements a new logic in the `release` command that will check if a preview artifact is cached for that app release, and delete it if there is.

Fixes #1746 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/models/release_platform.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/models/release_platform.dart
@@ -1,12 +1,15 @@
 /// A platform to which a Shorebird release can be deployed.
 enum ReleasePlatform {
   // ignore: public_member_api_docs
-  android('Android'),
+  android('Android', 'aab'),
   // ignore: public_member_api_docs
-  ios('iOS');
+  ios('iOS', 'app');
 
-  const ReleasePlatform(this.displayName);
+  const ReleasePlatform(this.displayName, this.extension);
 
   /// The display name of the platform.
   final String displayName;
+
+  /// The file extension for the platform.
+  final String extension;
 }
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -89,6 +89,19 @@ class Cache {
     );
   }
 
+  String getPreviewArtifactPath({
+    required String appId,
+    required String releaseVersion,
+    required String platformName,
+    required String extension,
+  }) {
+    final previewDirectory = cache.getPreviewDirectory(appId);
+    return p.join(
+      previewDirectory.path,
+      '${platformName}_$releaseVersion.$extension',
+    );
+  }
+
   /// The Shorebird cache directory.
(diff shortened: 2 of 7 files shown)


packages/shorebird_code_push_protocol/lib/src/models/release_platform.dart
```diff
@@ -1,12 +1,15 @@
 /// A platform to which a Shorebird release can be deployed.
 enum ReleasePlatform {
   // ignore: public_member_api_docs
-  android('Android'),
+  android('Android', 'aab'),
```
erickzanardo, line 4: I wonder if I need to extract this change to its own PR? It slipped my attention that this file is in a different package...

bryanoltman:
I'm not sure this fully addresses the issue. This will only solve the problem for the developer who runs the release command. If the release is created on CI, for example, this problem still exists. I would instead maybe update the caching logic to include the release ID as part of the directory name, so two `1.0.0+1` releases would still end up in different directories and no one would have this problem.

erickzanardo:
> I'm not sure this fully addresses the issue. This will only solve the problem for the developer who runs the release command. If the release is created on CI, for example, this problem still exists. I would instead maybe update the caching logic to include the release ID as part of the directory name, so two `1.0.0+1` releases would still end up in different directories and no one would have this problem.

Oh, that is true, I didn't thought on that case, adding the release id is so much more efficient 🤦

bryanoltman:
You could also change the directory name to `${platform.name}_${release.id}.$extension` from `${platform.name}_${release.version}.$extension` and that would solve the problem too without including redundant info

erickzanardo:
Closing this since I totally overcomplicated this.

PR #2254, opened 2024-06-17
fix(shorebird_cli): `shorebird preview` to not use outdated cached preview artifacts

### LINKED ISSUE #1746: Build does not update if you delete a release and then upload new one with same version number

erickzanardo:
## Description

When running a preview, the CLI will always use the cache artifact if one if found. This might make the command run an outdated artifact when a release is delete by some reason.

This PR makes the artifact id to be part of the path of the preview artifact, that way, if a release is deleted and recreated, the id will be a fresh one and the cli will not accidentally use the wrong cached artifact.

Fixes https://github.com/shorebirdtech/shorebird/issues/1746

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/preview_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/preview_command.dart
@@ -213,41 +213,54 @@ class PreviewCommand extends ShorebirdCommand {
     String? deviceId,
   }) async {
     const platform = ReleasePlatform.android;
-    final aabFile = File(
-      getArtifactPath(
-        appId: appId,
-        release: release,
-        platform: platform,
-        extension: 'aab',
-      ),
-    );
 
-    if (!aabFile.existsSync()) {
-      aabFile.createSync(recursive: true);
-      final downloadArtifactProgress = logger.progress('Downloading release');
-      try {
-        final releaseAabArtifact =
-            await codePushClientWrapper.getReleaseArtifact(
+    final downloadArtifactProgress = logger.progress('Downloading release');
+    late File aabFile;
+    late ReleaseArtifact artifact;
+
+    try {
+      artifact = await codePushClientWrapper.getReleaseArtifact(
+        appId: appId,
+        releaseId: release.id,
+        arch: 'aab',
+        platform: platform,
+      );
+    } catch (e, s) {
+      logger
+        ..err('Error getting release artifact: $e')
+        ..detail('Stack trace: $s');
+      return ExitCode.software.code;
+    }
+
+    try {
+      aabFile = File(
+        getArtifactPath(
           appId: appId,
-          releaseId: release.id,
-          arch: 'aab',
+          release: release,
+          artifact: artifact,
           platform: platform,
-        );
+          extension: 'aab',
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -213,41 +213,54 @@ class PreviewCommand extends ShorebirdCommand {
     String? deviceId,
   }) async {
     const platform = ReleasePlatform.android;
-    final aabFile = File(
-      getArtifactPath(
+
+    final downloadArtifactProgress = logger.progress('Downloading release');
+    late File aabFile;
+    late ReleaseArtifact artifact;
```
bryanoltman:
```suggestion
    late ReleaseArtifact releaseAabArtifact;
```

packages/shorebird_cli/test/src/commands/preview_command_test.dart
```diff
@@ -915,6 +917,28 @@ channel: ${track.channel}
...
+          when(
+            () => codePushClientWrapper.getReleaseArtifact(
+              appId: any(named: 'appId'),
+              releaseId: any(named: 'releaseId'),
+              arch: any(named: 'arch'),
+              platform: any(named: 'platform'),
+            ),
+          ).thenThrow(Exception('oops'));
+        });
+        test('returns error and logs', () async {
```
felangel:
```suggestion
        });

        test('returns error and logs', () async {
```
nit newline

packages/shorebird_cli/test/src/commands/preview_command_test.dart
```diff
@@ -1202,6 +1226,34 @@ channel: ${DeploymentTrack.staging.channel}
...
```
felangel:
```suggestion
        });

        test('returns error and logs', () async {
```
nit newline

bryanoltman (approved):

felangel (approved):
LGTM ✅

PR #2259, opened 2024-06-18
fix: e2e tests

erickzanardo:
## Description

A while ago we added a [shorebird cache clear test](https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/integration_test/shorebird_cli_cache_test.dart), which like the name suggests, tests the `cache clear` command.

That test interfered with the other e2e tests though, since they expected that everything were pre-cached.

This PR splits the cli integration tests into two, running the cache related ones in an isolated step, in order for it to not break the main tests expectation.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart
+++ b/packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart
@@ -32,11 +32,15 @@ R runWithOverrides<R>(R Function() body) {
 }
 
 void main() {
+  final shorebirdHostedURL = Platform.environment['SHOREBIRD_HOSTED_URL'];
+  if (shorebirdHostedURL == null || shorebirdHostedURL.isEmpty) {
+    throw Exception('SHOREBIRD_HOSTED_URL environment variable is not set.');
+  }
   final logger = Logger();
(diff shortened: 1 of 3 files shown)


bryanoltman (commented):
I think it might be better to have the cache test restore shorebird to the state it was in prior to the test than to split each test into its own step.

packages/shorebird_cli/integration_test/shorebird_cli_integration_test.dart
```diff
@@ -32,11 +32,15 @@ R runWithOverrides<R>(R Function() body) {
 }
 
 void main() {
+  final shorebirdHostedURL = Platform.environment['SHOREBIRD_HOSTED_URL'];
```
erickzanardo, line 35: Not really related to fixing the tests, but I got confused when I tried to run the tests and it failed because of null check. Made this small change to make it clear why it is failing when the variable is not set.

bryanoltman (approved):

PR #2264, opened 2024-06-18
fix: better error handling for release with no platforms provided

### LINKED ISSUE #2261: fix: `shorebird release` prints unhelpful error message when no platform is provided.
**Description**

When running `shorebird release` without any platforms, it fails with a message that doesn't tell much: `Bad state: No element`

**Steps To Reproduce**

1. Go to a shorebird proejct
2. Run `shorebird release` intentionally not passing any platforms
3. See `Bad state: No element` error

**Expected Behavior**

It should show a more useful message like `no messages` informed or similar.

erickzanardo:
## Description

Adds a better error handling for the release command when no platforms were provided.

Fixes #2261 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/release_type.dart
+++ b/packages/shorebird_cli/lib/src/release_type.dart
@@ -49,16 +49,22 @@ enum ReleaseType {
 
 extension ReleaseTypeArgs on ArgResults {
   Iterable<ReleaseType> get releaseTypes {
-    final List<String> releaseTypeCliNames;
+    List<String>? releaseTypeCliNames;
     if (wasParsed('platforms')) {
       releaseTypeCliNames = this['platforms'] as List<String>;
     } else {
-      final platformCliName = arguments.first;
-      if (ReleaseType.values
-          .none((target) => target.cliName == platformCliName)) {
-        throw ArgumentError('Invalid platform: $platformCliName');
+      if (arguments.isNotEmpty) {
+        final platformCliName = arguments.first;
+        if (ReleaseType.values
+            .none((target) => target.cliName == platformCliName)) {
+          throw ArgumentError('Invalid platform: $platformCliName');
+        }
+        releaseTypeCliNames = [platformCliName];
       }
-      releaseTypeCliNames = [platformCliName];
+    }
+
+    if (releaseTypeCliNames == null) {
+      return const [];
     }
 
     return releaseTypeCliNames.map(
--- a/packages/shorebird_cli/lib/src/release_type.dart
+++ b/packages/shorebird_cli/lib/src/release_type.dart
@@ -63,13 +63,11 @@ extension ReleaseTypeArgs on ArgResults {
       }
     }
 
-    if (releaseTypeCliNames == null) {
-      return const [];
-    }
-
-    return releaseTypeCliNames.map(
-      (cliName) =>
-          ReleaseType.values.firstWhere((target) => target.cliName == cliName),
-    );
+    return releaseTypeCliNames?.map(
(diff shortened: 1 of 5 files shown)


packages/shorebird_cli/test/src/commands/release/release_command_test.dart
```diff
@@ -521,5 +521,25 @@ $exception''',
...
+      });
+
+      test('fails and log the correct message', () async {
+        final exitCode = await runWithOverrides(command.run);
+
+        expect(exitCode, equals(ExitCode.usage.code));
+
+        verify(
+          () => logger.err(
+            'At least one platform must be specified.',
```
eseidel: Is this all the user sees?  Will they know what to do with that or how to specify a platform?

packages/shorebird_cli/test/src/commands/release/release_command_test.dart
```diff
@@ -521,5 +521,25 @@ $exception''',
         });
       });
     });
+
+    group('when no platform argument is provided', () {
+      // https://github.com/shorebirdtech/shorebird/issues/2261
```
bryanoltman: I'm not sure this is necessary

packages/shorebird_cli/test/src/commands/release/release_command_test.dart
erickzanardo: yeah, that is the only error message that I implemented, open to suggestions on a better message :)

bryanoltman (approved):

packages/shorebird_cli/test/src/commands/release/release_command_test.dart
erickzanardo: @eseidel I think they will known what to do as this is probably a often used command and they just did forgot by mistake to add the platforms.

But we could eventually make this message something like:

```
No platforms were provided, use the --platforms argument to provide one or more platforms
```

Wdyt?

bryanoltman: I agree with @eseidel here—we should endeavor to tell the user how to fix their problems when we can.

bryanoltman: > `No platforms were provided, use the --platforms argument to provide one or more platforms`

Let's use this.

erickzanardo: That is fair! I've changed, can you take another look please?

packages/shorebird_cli/lib/src/release_type.dart
```diff
@@ -49,16 +49,22 @@ enum ReleaseType {
...
+          throw ArgumentError('Invalid platform: $platformCliName');
+        }
+        releaseTypeCliNames = [platformCliName];
       }
-      releaseTypeCliNames = [platformCliName];
+    }
+
+    if (releaseTypeCliNames == null) {
+      return const [];
     }
```
bryanoltman: I don't think we need this early return. `map` on an empty list will return an empty list

bryanoltman (approved):

packages/shorebird_cli/lib/src/release_type.dart
erickzanardo: oh, but it is not checking on an empty list, it is checking on nullable map, we need an early return.

Either way, your comment made me think and I changed the early return for an elvis operator and a coalesce to empty list directly on the return and I guess it looks better!

Issue #2257, opened 2024-06-17
Compare hashes of Podfile.lock as a proxy for native changes


bryanoltman:
**Description**

Context: https://github.com/shorebirdtech/shorebird/issues/2210

Because Xcode builds are sometimes nondeterministic, our native code change detection yields a frustrating number of false positives. We are currently just comparing the hashes of unsigned MachO binaries.

A reasonable proxy for this might be to compare the Podfile.lock as it was at release vs. how it is at patch (we would store the hash of these so as not to leak any PII). This would not catch all potential bad native changes, and may yield false negatives (i.e., native code changed without the podfile.lock updating), but should catch what we believe to be the biggest cause of potential native change issues, which is plugins updating between release and patch.

This work will probably look like:

- [x] Add `podfile_lock_hash` column to the release table
- [x] Update CLI/backend API to communicate this
- [x] Add check on CLI

PR #2267, opened 2024-06-19
fix(shorebird_cli): use Podfile.lock hash to detect native changes

bryanoltman:
## Description

Updates iOS patch diff checking to no longer warn if binary files differ, and instead compares the hash of the current Podfile.lock to what it was when the release was created, warning if the two are different. 

Part of https://github.com/shorebirdtech/shorebird/issues/2257

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -82,6 +82,26 @@ class IosPatcher extends Patcher {
   }
 
   @override
+  Future<void> checkPatchability({
+    required ReleaseArtifact releaseArtifact,
+  }) async {
+    final podfileLockHash = await ios.podfileLockHash();
+    if (podfileLockHash != releaseArtifact.podfileLockHash) {
+      logger
+        ..warn(
+          '''Your ios/Podfile.lock is different from the one used to build the release.''',
+        )
+        ..info(
+          yellow.wrap(
+            '''This may indicate that the patch contains native changes, which cannot be applied with a patch.''',
+          ),
+        );
+
+      // TODO
+    }
+  }
+
+  @override
   Future<File> buildPatchArtifact() async {
     final File exportOptionsPlist;
     try {
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -98,8 +98,6 @@ class IosPatcher extends Patcher {
       confirmNativeChanges: false,
     );
 
-    print('has native changes: ${diffStatus.hasNativeChanges}');
-
     if (!diffStatus.hasNativeChanges) {
       return diffStatus;
     }
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -109,7 +109,8 @@ class IosPatcher extends Patcher {
         .convert(shorebirdEnv.podfileLockFile.readAsBytesSync())
         .toString();
 
-    if (podfileLockHash != releaseArtifact.podfileLockHash) {
+    if (releaseArtifact.podfileLockHash != null &&
+        podfileLockHash != releaseArtifact.podfileLockHash) {
       logger
         ..warn(
           '''Your ios/Podfile.lock is different from the one used to build the release.''',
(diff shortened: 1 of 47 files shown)


packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -354,18 +358,24 @@ Please re-run the release command for this version or create a new release.''');
...
-        localArtifact: patchArtifact,
+      return patcher.assertUnpatchableDiffs(
         releaseArtifact: releaseArtifact,
-        archiveDiffer: archiveDiffer,
-        allowAssetChanges: allowAssetDiffs,
-        allowNativeChanges: allowNativeDiffs,
+        releaseArchive: releaseArchive,
+        patchArchive: patchArchive,
       );
+      //   localArchive: patchArtifact,
```
erickzanardo: Probably forgotten?

bryanoltman: Good catch, removed

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -81,6 +78,65 @@ class IosPatcher extends Patcher {
...
+      return diffStatus;
+    }
+
+    final String? podfileLockHash;
+    if (shorebirdEnv.podfileLockFile.existsSync()) {
+      podfileLockHash = sha256
+          .convert(shorebirdEnv.podfileLockFile.readAsBytesSync())
+          .toString();
+    } else {
+      podfileLockHash = null;
```
felangel, line 111: can't we return early here?
```suggestion
      return diffStatus;
```

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -81,6 +78,65 @@ class IosPatcher extends Patcher {
...
+    }
+
+    if (releaseArtifact.podfileLockHash != null &&
+        podfileLockHash != releaseArtifact.podfileLockHash) {
+      logger
+        ..warn(
+          '''Your ios/Podfile.lock is different from the one used to build the release.''',
+        )
+        ..info(
```
felangel: any reason why we didn't add the content of the second `info` into the original `warn` as a multi-line string?

erickzanardo (approved):
LGTM!

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
bryanoltman: No—if a release has a podfile lock hash but we don't currently have a podfile lock, we need to warn about that.

packages/shorebird_cli/lib/src/patch_diff_checker.dart
```diff
@@ -12,7 +12,7 @@ import 'package:shorebird_cli/src/shorebird_env.dart';
 /// {@endtemplate}
 class DiffStatus {
   /// {@macro diff_status}
-  DiffStatus({
+  const DiffStatus({
```
felangel, line 15: nice 💯

packages/shorebird_cli/test/src/code_push_client_wrapper_test.dart
```diff
@@ -1592,6 +1611,7 @@ You can manage this release in the ${link(uri: uri, message: 'Shorebird Console'
               xcarchivePath: p.join(projectRoot.path, xcarchivePath),
               runnerPath: p.join(projectRoot.path, runnerPath),
               isCodesigned: true,
+              podfileLockHash: 'TODO',
```
felangel: is this TODO intentional?

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
bryanoltman: Just that this was carried over from the PatchDiffChecker, which does things this way. Updated.

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
felangel: yeah nvm makes sense 👍

felangel (approved):

packages/shorebird_cli/test/src/code_push_client_wrapper_test.dart
bryanoltman: Nope 😄  fixed

Issue #2154, opened 2024-05-26
If release exists but is not previewable on one or more platforms, explain in `shorebird preview` output


bryanoltman:
**Description**

If we don't have any sideloadable release artifacts for a platform, we don't display it as an option in `shorebird preview`. This can be confusing if, for example, a release is available on Android and iOS, only iOS devices are available, but the iOS release is not previewable (due to not being signed).

This would make situations like https://github.com/shorebirdtech/shorebird/issues/2153 more understandable.

erickzanardo:
Solved by #2278

PR #2269, opened 2024-06-19
feat(shorebird_code_push_protocol): Adding can sideload to artifacts

erickzanardo:
## Description

Part of #2154 

Adding `canSideload` property to `ReleaseArtifact`.

This property will then be "consumed" on the CLI to check if the artifact can be pre-viewable or not

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/test/src/code_push_client_wrapper_test.dart
+++ b/packages/shorebird_cli/test/src/code_push_client_wrapper_test.dart
@@ -140,6 +140,7 @@ void main() {
       size: 4,
       url: 'url',
       podfileLockHash: 'podfile-lock-hash',
+      canSideload: true,
     );
 
     late CodePushClient codePushClient;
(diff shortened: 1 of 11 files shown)


bryanoltman:
This seems fine, but it's not clear to me how this moves us towards solving https://github.com/shorebirdtech/shorebird/issues/2154. Maybe explain a bit more here or in a doc?

erickzanardo:
> This seems fine, but it's not clear to me how this moves us towards solving #2154. Maybe explain a bit more here or in a doc?

#2154 says

> If we don't have any sideloadable release artifacts for a platform, we don't display it as an option in shorebird preview. This can be confusing if, for example, a release is available on Android and iOS, only iOS devices are available, but the iOS release is not previewable (due to not being signed).

Then I noticed that release artifacts already have a can_sideload field in the database. But the field is not exposed in our APIs, so I figured that the first step would be to add this field in the `shorebird_code_push_protocol`, which we will then afterwards be able to consume it on the CLI to show or not the platform as a preview option.

Does it makes sense?

bryanoltman:
Makes sense!

bryanoltman (approved):

PR #2275, opened 2024-06-19
feat: assert artifact is sideloadable on preview

### LINKED ISSUE #2154: If release exists but is not previewable on one or more platforms, explain in `shorebird preview` output

erickzanardo:
## Description

Fixes #2154 

When previewing a release that cannot be side loaded, the user will face a hard to understand message.

This PR checks before previewing if the artifact can be side loaded, and will show a more meaningful message.

Maybe a better solution to this, would be to not even show non sideloadable releases for the user, but that will introduce a lot of complexness since it we will need to refactor the flow of the preview command, and we will need to handle it differently when the user provide the release version, and when they prompt.

So for the sake of simplicity and since this issue hasn't caused that much confusion since it is in a way, an edge case, I opted for a simple solution.

WIP: Still need to write docs and test it more.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/preview_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/preview_command.dart
@@ -19,8 +19,10 @@ import 'package:shorebird_cli/src/executables/executables.dart';
 import 'package:shorebird_cli/src/logger.dart';
 import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/shorebird_command.dart';
+import 'package:shorebird_cli/src/shorebird_documentation.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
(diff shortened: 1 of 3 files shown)


erickzanardo:
As I was testing this, I realized that we don't allow artifacts that can't side load in the preview command. This filter render these changes not needed. Closing this.

PR #2293, opened 2024-06-26
feat(shorebird_cli): add note about sharability of app_id to `shorebird init`

### LINKED ISSUE #969: We should add a note in `shorebird init` output that these are not secrets.  People still get confused by that.
We should add a note in `shorebird init` output that these are not secrets.  People still get confused by that.

_Originally posted by @eseidel in https://github.com/shorebirdtech/docs/pull/106#discussion_r1281069570_

bryanoltman:
## Description

Adds a note in the `shorebird init` output telling users that the content of `shorebird.yaml` is shareable and should be checked in to version control.

Fixes https://github.com/shorebirdtech/shorebird/issues/969

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -216,6 +216,9 @@ ${lightGreen.wrap('🐦 Shorebird initialized successfully!')}
 ✅ A "shorebird.yaml" has been created.
 ✅ The "pubspec.yaml" has been updated to include "shorebird.yaml" as an asset.
 
+shorebird.yaml does not contain any sensitive information and should be checked into version control.
+You can freely share your app_id with others.
+
 Reference the following commands to get started:
 
 📦 To create a new release use: "${lightCyan.wrap('shorebird release')}".
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -216,9 +216,6 @@ ${lightGreen.wrap('🐦 Shorebird initialized successfully!')}
 ✅ A "shorebird.yaml" has been created.
 ✅ The "pubspec.yaml" has been updated to include "shorebird.yaml" as an asset.
 
-shorebird.yaml does not contain any sensitive information and should be checked into version control.
-You can freely share your app_id with others.
-
 Reference the following commands to get started:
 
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -216,6 +216,9 @@ ${lightGreen.wrap('🐦 Shorebird initialized successfully!')}
 ✅ A "shorebird.yaml" has been created.
 ✅ The "pubspec.yaml" has been updated to include "shorebird.yaml" as an asset.
 
+shorebird.yaml does not contain any sensitive information and should be checked into version control.
+You can freely share your app_id with others.
```
eseidel: I think this is good, but it woudl be better to just put these as comments in shorebird.yaml.

Some might also view "share your app_id" as sharing app_id between apps?

eseidel (approved):

packages/shorebird_cli/test/src/commands/init_command_test.dart
```diff
@@ -883,6 +883,31 @@ flutter:
...
+      verify(
+        () => logger.info(
+          any(
+            that: stringContainsInOrder(
+              [
+                lightGreen.wrap('🐦 Shorebird initialized successfully!')!,
+                '✅ A shorebird app has been created.',
+                '✅ A "shorebird.yaml" has been created.',
+                '''✅ The "pubspec.yaml" has been updated to include "shorebird.yaml" as an asset.''',
+                '''shorebird.yaml does not contain any sensitive information and should be checked into version control.''',
```
erickzanardo: Should we emphasize "public version control"?

erickzanardo (approved):

eseidel (approved):

packages/shorebird_cli/lib/src/commands/init_command.dart
bryanoltman: It looks like we already say "this isn't secret" in the yaml comments. Updated those to be (I hope) clearer.

packages/shorebird_cli/test/src/commands/init_command_test.dart
bryanoltman: I don't know that users will always want to check their app into public version control. Most companies are closed-source by default.

PR #2294, opened 2024-06-26
feat: splitting the Building with Flutter log message into more steps

### LINKED ISSUE #2290: feat: Break progress "Building app with Flutter..." on more steps
**Description**

Right now, when running a command that builds the app (`release` and `patch`), we use the following message `Building app bundle with Flutter $flutterVersionString` with a spinner, without showing all the logs that `flutter build` is doing under the hood.

This gives the user the impression that a `flutter build` through shorebird takes longer than just running `flutter build`, this is specially impacting on CI, where builds are by nature slower due limited computational power, and that the flutter tool will need to do make caching operations since it is a first build in a "cold boot system".

To improve user's perception of what is happening, we could break that step in a couple of them, like:

`Building app with Flutter`
`Downloading gradlew and its dependencies`
`Preparing and installing Android SDKs`
`Compiling app with Gradle`

Or some steps similar to that.

Related to https://github.com/shorebirdtech/shorebird/issues/2211

erickzanardo:
## Description

This PR changes the logging progress when building an app bundle in a way that it will show more updates to the user of what is happening.

Now, instead of showing "Building with Flutter xxx" during the whole time while the `flutter build ...` command is in progress, we change to other status when those can be identified. For example, if the build command had to download gradlew because it is the first time that a build is executed in that computer, it will change to Downloading gradlew.

The status will change everytime a "special" step is identified, and will return to the "Building with Flutter xxx" when out of a special step.

For now, this is only implemented on android app bundle. Once we are agreed on the proposal and landed this PR, I plan on implementing a new PR applying the same changes to iOS.

Feature screen recording (video is at faster speed)
https://github.com/shorebirdtech/shorebird/assets/835641/38665914-4c49-4519-862d-9167b6cd7e76

Fixes #2290

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
```diff
@@ -93,22 +96,36 @@ class ArtifactBuilder {
...
         executable,
         arguments,
         runInShell: true,
         environment: base64PublicKey?.toPublicKeyEnv(),
       );
 
-      if (result.exitCode != ExitCode.success.code) {
+      final sub = spawnedProcess.stdout.map(utf8.decode).listen((event) {
+        progressUpdater?.onLog(event);
+      });
```
eseidel: This always subscribes even if there isn't a progressUpdater.  I wonder if it's better to only subscribe if there is one.  Then the listen call is just listen(progressUpdater.onLog). 🤷

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_log_updater.dart
```diff
@@ -0,0 +1,108 @@
...
+  /// Generic step called when the build is happening, but
+  /// no specifc, or outstanding, step has been identified.
+  building,
+
+  /// The project does not have a gradle wrapper binary and it is being
+  /// downloaded. (Android only)
+  downloadingGradleW,
+
+  /// Flutter is checking and preparing the Android SDK. (Android only)
+  preparingAndroidSDK,
```
eseidel: I wonder if this should be a super enum and store the message and parsing regexp as variables. 🤷 

Or I guess alternatively it could just be a list of objects which implement those two methods.  This design is fine too, mostly musing.

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
```diff
@@ -93,22 +96,36 @@ class ArtifactBuilder {
...
-      final result = await process.run(
+      final spawnedProcess = await process.start(
         executable,
         arguments,
         runInShell: true,
         environment: base64PublicKey?.toPublicKeyEnv(),
       );
 
-      if (result.exitCode != ExitCode.success.code) {
+      final sub = spawnedProcess.stdout.map(utf8.decode).listen((event) {
```
bryanoltman: `sub` could use a more descriptive name

eseidel (commented):
Felix or Bryan are better reviewers than I am here, but overall lgtm.

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
```diff
@@ -93,22 +96,36 @@ class ArtifactBuilder {
...
+      final stderr = StringBuffer();
+      final stderrSub =
+          spawnedProcess.stderr.map(utf8.decode).listen(stderr.write);
+
+      final exitCode = await spawnedProcess.exitCode;
+
+      await sub.cancel();
+      await stderrSub.cancel();
+
+      if (exitCode != ExitCode.success.code) {
```
bryanoltman: I'm wondering if this is the right level to make this change. To extend this to other build types, we'll need to add this exact same ~20 lines of code to every `build[ArtifactType]` method in this file. I wonder if a slightly modified version of this in https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/shorebird_process.dart#L109 would be easier to write/test/maintain?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -246,6 +246,14 @@ class AotToolsDillArtifact extends CachedArtifact {
...
 /// Should be removed sometime after June 2024.
+///
+/// The change to use a .dill file was made in
+/// https://github.com/shorebirdtech/_build_engine/commit/babbc37d93e7a2f36e62787e47eee5a3b5458901
+/// The Flutter versions that use this are:
+///  - 3.13.9 (a3d5f7c614aa1cc4d6cb1506e74fd1c81678e68e)
+///  - 3.16.3 (b9b23902966504a9778f4c07e3a3487fa84dcb2a)
+///  - 3.16.4 (7e92b034c5dddb727cf5e802c23cddd39b325a7f)
+///  - 3.16.5 (4e8a7c746ae6f10951f3e676f10b82b21d7300a5)
 class AotToolsExeArtifact extends CachedArtifact {
```
bryanoltman, line 257: Thanks for bringing this over!

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_log_updater.dart
```diff
@@ -0,0 +1,108 @@
...
+
+  /// The current step in the build process.
+  LogUpdaterStep get currentStep => _steps.last;
+
+  /// Callback that is called when a new step is identified in the build
+  final BuildStepCallback onBuildStep;
+
+  /// Should be called every time a new log is written to the process
+  /// stdout.
+  void onLog(String log) {
```
bryanoltman:
```suggestion
  void onLog(String logLine) {
```

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_log_updater.dart
bryanoltman, line 1: This looks like a state machine, and it might be good to make that more explicit. Some questions:

1. Can a step happen more than once per build?
2. What happens when we receive lines we don't recognize? Does the updater just stay on the previous step? I worry that, in trying to spell out every step the flutter build might take, we might give users bad info. For example, it doesn't look like a line like `2024-06-25T18:24:16.5672921Z [        ] Downloading https://dl.google.com/dl/android/maven2/com/android/tools/analytics-library/tracker/30.3.0/tracker-30.3.0.pom to /home/runner/.gradle/.tmp/gradle_download4852343579353170349bin` would be captured by any of these steps. If I'm reading the code correctly, it looks like it will stay on the existing step, which might mean we tell the user that the build is doing one thing when it's actually doing another. Is there a way would could make this more generic such that we don't need to write a new set of matchers for every platform and that we don't need to be aware of the specific steps Flutter takes to build? It's possible the answer is "no", but it would be good to make this robust to changes in flutter build steps between flutter revisions and between build types.

packages/shorebird_cli/test/src/commands/release/android_releaser_test.dart
```diff
@@ -411,6 +423,49 @@ To change the version of this release, change your app's version in your pubspec
...
+          ).thenAnswer((invocation) async {
+            final progressUpdater = invocation.namedArguments[#progressUpdater]
+                as FlutterBuildLogUpdater;
+
+            final lines = buildOutputFile.readAsLinesSync();
+            for (final line in lines) {
+              progressUpdater.onLog(line);
+            }
+
+            return aabFile;
```
bryanoltman: Nice!

bryanoltman (changes requested):
It's possible I'm doing something wrong, but it doesn't look like this works. When I check out this branch and run `shorebird release android`, I just see a single `Building release artifacts with Flutter 3.22.2` progress spinner with no other updates.

Also left a few other comments.

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
erickzanardo: Hum, that is an interesting idea! I will try to come up with something like that!

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
erickzanardo: yeah, I had that wondering when I wrote this, I ended with this because I felt like the conditions that would needed to be added would make the code harder to read, with little benefit. I will put more thought on that either way since this will change with the suggestion Bryan gave below.

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_log_updater.dart
erickzanardo: > Can a step happen more than once per build?
As far as I tested, none of the identified steps that I implemented happens more than once. 

> What happens when we receive lines we don't recognize?

We ignore, the status will be the previous one. I choose only events where I could clearly see where it started and where it ended, and the idea is that everytime that we are not in one of those blocks, we return to that generic building message.

> For example, it doesn't look like a line like 2024-06-25T18:24:16.5672921Z [        ] Downloading https://dl.google.com/dl/android/maven2/com/android/tools/analytics-library/tracker/30.3.0/tracker-30.3.0.pom to /home/runner/.gradle/.tmp/gradle_download4852343579353170349bin

You are correct, I intentionally decided to ignore these events because they seem to happen at random. Not really at random ofc, but they happen all in a defined block, there isn't a "Downloading gradle's dependencies".

But now that I am writing this message, I think I can improve that. I will make some changes and will be back to you.

> Is there a way would could make this more generic such that we don't need to write a new set of matchers for every platform and that we don't need to be aware of the specific steps Flutter takes to build? It's possible the answer is "no", but it would be good to make this robust to changes in flutter build steps between flutter revisions and between build types.

I spent some time in that same question. The way I implemented the flutter build log updater is meant to be both generic and specific. Some of its steps are generic, like the assemble one. But unfortunately all the steps that I managed to identify and that would fit the criteria that I explained above, are most platform specific. Gradle is a good example of it, it is one of the biggest offenders on the build time, but it is android only.

I hope that as we (if we?) decide to spend more time on this and make it smarter, we will identify additional generic steps, but at the moment I am unsure about any.

erickzanardo:
> It's possible I'm doing something wrong, but it doesn't look like this works. When I check out this branch and run `shorebird release android`, I just see a single `Building release artifacts with Flutter 3.22.2` progress spinner with no other updates.
> 
> Also left a few other comments.

The first time I tested I saw nothing and thought I had implemented all wrong, but then I realized that since we have everything cached in our computers because of our day to day, the log simply don't change because the steps doesn't exists (like the android sdk one, which will only mostly likely show in a fresh install computer) or they are too fast (the assemble steps for example, in my computer, in my test project that takes 0ms).

A quick way to see the log changing is to delete your `.gradle/cache` folder.

erickzanardo:
@bryanoltman @eseidel I attached a screen recording in the PR description to demo the feature.

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
erickzanardo: I refactored a lot the PR and it should now look like more to what you suggested here, lmk wdyt

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
erickzanardo: I refactored a bunch of stuff in the PR and this went away in the changes, lmk wdyt

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_process_tracker.dart
```diff
@@ -0,0 +1,136 @@
...
+      'Downloading https://services.gradle.org/distributions/gradle',
+    )) {
+      _emitStep(FlutterBuildProcessTrackerStep.downloadingGradleW);
+    } else if (logLine.contains('Downloading') && logLine.contains('.gradle')) {
+      _emitStep(FlutterBuildProcessTrackerStep.downloadingGradleDependency);
+    } else if (logLine.contains('Parsing') &&
+        logLine.contains('android${platform.pathSeparator}sdk') &&
+        currentStep != FlutterBuildProcessTrackerStep.preparingAndroidSDK) {
+      _emitStep(FlutterBuildProcessTrackerStep.preparingAndroidSDK);
+    } else if (logLine.contains('SDK initialized in') &&
```
eseidel, line 103: It seems like some of these care about step order and some of them don't?

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_process_tracker.dart
```diff
@@ -0,0 +1,136 @@
...
+      case FlutterBuildProcessTrackerStep.downloadingGradleDependency:
+        return 'Downloading Gradle Dependency';
+      case FlutterBuildProcessTrackerStep.preparingAndroidSDK:
+        return 'Preparing Android SDK';
+      case FlutterBuildProcessTrackerStep.flutterAssemble:
+        return 'Assembling Flutter Resources';
+      case FlutterBuildProcessTrackerStep.initial:
+      case FlutterBuildProcessTrackerStep.building:
+        return null;
+    }
```
eseidel, line 54: I think as we add more, you're going to want these to be closer to the detection logic and name.  I suspect you'll end up with a system like:

final steps = [
   step('Downloading Gradle Wrapper', (line) => line.contains('Downloading') && line.contains('.gradle))
...
];
Or something. 🤷

felangel:
This seems like a pretty complex solution that will introduce a significant amount of maintenance overhead (whenever upstream Flutter modifies their logs or build steps Shorebird will need to adjust). Is there any reason why we wouldn't want to just proxy the Flutter logs directly?

eseidel (commented):
If you've addressed bryan's feedback, than lgtm.

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_process_tracker.dart
erickzanardo: Yes. It is more about "starting and finishing an step" instead of the order itself.

For example the following:

```
 } else if (logLine.contains('Parsing') &&
        logLine.contains('android${platform.pathSeparator}sdk') &&
        currentStep != FlutterBuildProcessTrackerStep.preparingAndroidSDK) {
      _emitStep(FlutterBuildProcessTrackerStep.preparingAndroidSDK);
```

There isn't any obvious log like "Starting to install Android SDK" that we can look for, but I realized from testing that the preparing android SDK step starts when we start to see "Parsing ... android/sdk". But the log will have many lines like that in the sequence, so this logic will make sure that we trigger the `preparingAndroidSDK` step only when we see the first line of such message.

packages/shorebird_cli/lib/src/artifact_builder/flutter_build_process_tracker.dart
```diff
@@ -0,0 +1,136 @@
...
+      // then we know that the download is over.
+    } else if (currentStep ==
+        FlutterBuildProcessTrackerStep.downloadingGradleDependency) {
+      final lastNonGradleDownload = _steps.lastWhere(
+        (step) =>
+            step !=
+                FlutterBuildProcessTrackerStep.downloadingGradleDependency &&
+            step != FlutterBuildProcessTrackerStep.downloadingGradleW,
+        // We don't need a orElse here because there will always be a
+        // non gralde step.
```
felangel, line 122:
```suggestion
        // non gradle step.
```

eseidel:
We discussed this over Discord and even wrote up a design doc.

This current implementation does not leave a papertrail, so someone who came back to their computer would only see that "build" took 20mins rather than be able to say what parts of build.

I think we should pause this CL until we get more data from users like @mrgnhnt96 who have reported slow build times.

mrgnhnt96:
👋 
Looking through my previous CI runs, and it looks like the android build takes about an avg of ~17 minutes. I was able to download the log archive from Github, so you should get some pretty detailed logs. The logs will contain everything from the very start of the job to the end of it. Github does a pretty good job of filtering out all the secrets and sensitive info, but to be safe, I am going to send it to your email. I'll send you 2 different logs and totally happy to get you more if you need them!
image

felangel:
> 👋 Looking through my previous CI runs, and it looks like the android build takes about an avg of ~17 minutes. I was able to download the log archive from Github, so you should get some pretty detailed logs. The logs will contain everything from the very start of the job to the end of it. Github does a pretty good job of filtering out all the secrets and sensitive info, but to be safe, I am going to send it to your email. I'll send you 2 different logs and totally happy to get you more if you need them! image

Thanks! Do you happen to have the equivalent build times without using shorebird? It’d be super helpful to see what the discrepancy is on your specific CI setup.

mrgnhnt96:
> Thanks! Do you happen to have the equivalent build times without using shorebird? It’d be super helpful to see what the discrepancy is on your specific CI setup.

On Monday I can test how long a non-shorebird build would take. The latest CI build that didn't use shorebird took just under 6 minutes. But that was also a bit of time ago and we didn't have as many dependencies & code that we do now.

PR #2302, opened 2024-06-28
fix(shorebird_cli): always respect build-name and build-number args

### LINKED ISSUE #2301: `build-number` argument not respected for shorebird release
The ReleaseCommand as a `build-number` option that should be forwarded to `flutter build` but is not. https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/commands/release/release_command.dart#L51-L60

bryanoltman:
## Description

We were previously only using `build-number` for Flutter module releases. Because this argument is respected for all `flutter build` operations, we should support it for all of them if we support it at all.

Fixes https://github.com/shorebirdtech/shorebird/issues/2301

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patcher.dart
@@ -6,6 +6,8 @@ import 'package:args/args.dart';
 import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
+import 'package:shorebird_cli/src/common_arguments.dart';
+import 'package:shorebird_cli/src/extensions/iterable.dart';
 import 'package:shorebird_cli/src/patch_diff_checker.dart';
 import 'package:shorebird_cli/src/platform/platform.dart';
 import 'package:shorebird_cli/src/release_type.dart';
@@ -139,8 +141,17 @@ https://docs.shorebird.dev/status#link-percentage-ios
       return [];
     }
 
-    // If the user already provided --build-name or --build-number, we don't
-    // want to override them.
+    // If the user provided --build-name or --build-number before the --, we
+    // don't want to override them.
+    if (argResults.options.containsAnyOf([
+      CommonArguments.buildNameArg.name,
+      CommonArguments.buildNumberArg.name,
+    ])) {
+      return [];
+    }
+
+    // If the user provided --build-name or --build-number after the --, we
+    // don't want to override them.
     if (argResults.rest.any(
       (a) => a.startsWith('--build-name') || a.startsWith('--build-number'),
     )) {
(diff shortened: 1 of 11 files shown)


packages/shorebird_cli/lib/src/extensions/iterable.dart
```diff
@@ -0,0 +1,12 @@
+/// Provides the [containsAnyOf] method to all [Iterable]s.
+extension ContainsAnyOf<T> on Iterable<T> {
+  /// Returns `true` if any of the elements in [elements] are in this iterable.
+  bool containsAnyOf(Iterable<T> elements) {
+    for (final element in elements) {
+      if (contains(element)) {
+        return true;
+      }
+    }
+    return false;
```
bryanoltman: I was very surprised to not find something in `package:collection` for this. If you're aware of something similar that I missed, please lmk

eseidel: This is just `elements.any((e) => iterable.contains(e))`, right?  It's O(N*M) in worst case.

erickzanardo (approved):

packages/shorebird_cli/lib/src/commands/patch/patcher.dart
```diff
@@ -139,8 +141,17 @@ https://docs.shorebird.dev/status#link-percentage-ios
...
+    // don't want to override them.
+    if (argResults.options.containsAnyOf([
+      CommonArguments.buildNameArg.name,
+      CommonArguments.buildNumberArg.name,
+    ])) {
+      return [];
+    }
+
+    // If the user provided --build-name or --build-number after the --, we
+    // don't want to override them.
```
felangel, line 154: Should we refactor the next few lines to use the CommonArguments name instead of a string literal?

packages/shorebird_cli/lib/src/extensions/iterable.dart
```diff
@@ -0,0 +1,5 @@
+/// Provides the [containsAnyOf] method to all [Iterable]s.
+extension ContainsAnyOf<T> on Iterable<T> {
+  /// Returns `true` if any of the elements in [elements] are in this iterable.
+  bool containsAnyOf(Iterable<T> elements) => elements.any(contains);
```
felangel, line 4: What’s the benefit of this extension? Seems like it’s a one liner either way

felangel (approved):

packages/shorebird_cli/lib/src/commands/patch/patcher.dart
bryanoltman: Good call, done

packages/shorebird_cli/lib/src/extensions/iterable.dart
bryanoltman: It looks a bit nicer with list literals:

```dart
argResults.options.containsAnyOf([a,b]);
```

instead of 

```dart
[a,b].any((e) => argResults.options.contains(e));
```

Mostly just a minor readability improvement, happy to remove if you feel strongly.

PR #2344, opened 2024-07-11
feat: adding versions subcommand which

erickzanardo:
## Description

Adds a new subcommand under `shorebird flutter versions`, which given a revision, returns the flutter version for that revision.

This command helped me build the update script for the recent added `flutter_version` field and figured it could help us quick have access to the flutter version calculation currently in the CLI which might help us on support when troubleshooting older app releases.

Opening this as a draft to gather feedback, will be implementing the tests once we decide if this is worth or not.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_command.dart
@@ -9,6 +9,7 @@ class FlutterVersionsCommand extends ShorebirdCommand {
   /// {@macro flutter_versions_command}
   FlutterVersionsCommand() {
     addSubcommand(FlutterVersionsListCommand());
+    addSubcommand(FlutterVersionsWhichCommand());
   }
 
   @override
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_which_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_which_command.dart
@@ -1,0 +1,49 @@
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/logger.dart';
+import 'package:shorebird_cli/src/shorebird_command.dart';
+import 'package:shorebird_cli/src/shorebird_flutter.dart';
+
+/// {@template flutter_versions_which_command}
+/// `shorebird flutter versions which [rev]`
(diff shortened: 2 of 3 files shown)


bryanoltman (changes requested):
If we're solving the problem of "what is the flutter revision that's currently running in the shorebird cli", `shorebird doctor` already tells us that

felangel:
I’d prefer not to increase the api surface unless this will benefit the majority of our customers.

erickzanardo:
> If we're solving the problem of "what is the flutter revision that's currently running in the shorebird cli", `shorebird doctor` already tells us that

@bryanoltman that is not the problem that this PR tries to solve, it tries to solve "what is the Flutter Version for this given revision"

erickzanardo:
> I’d prefer not to increase the api surface unless this will benefit the majority of our customers.

@felangel I tried explaining in the description, but I guess we would be the ones having benefit from this command, AFAIK there isn't an available way for us to get the version string from a revision without using the CLI logic.

If there is one and I am not aware, I guess this would indeed be not needed

bryanoltman:
> access to the flutter version calculation currently in the CLI

I think I misunderstood this then. Why would a user want to know the Flutter version of an arbitrary revision?

felangel:
> > I’d prefer not to increase the api surface unless this will benefit the majority of our customers.
> 
> @felangel I tried explaining in the description, but I guess we would be the ones having benefit from this command, AFAIK there isn't an available way for us to get the version string from a revision without using the CLI logic.
> 
> If there is one and I am not aware, I guess this would indeed be not needed

We can easily script this in a few lines of bash if we need plus we have this information in the DB now so I don't really see the value of adding a CLI command.

erickzanardo:
> > > I’d prefer not to increase the api surface unless this will benefit the majority of our customers.
> > 
> > 
> > @felangel I tried explaining in the description, but I guess we would be the ones having benefit from this command, AFAIK there isn't an available way for us to get the version string from a revision without using the CLI logic.
> > If there is one and I am not aware, I guess this would indeed be not needed
> 
> We can easily script this in a few lines of bash if we need plus we have this information in the DB now so I don't really see the value of adding a CLI command.

Alright, closing in favor of doing a bash script when needed.

PR #2346, opened 2024-07-11
fix(shorebird_cli): add checksum support for gcp artifacts

### LINKED ISSUE #1771: fix: validate integrity of aot_tools.dill asset on disk
From discord:
https://discord.com/channels/1030243211995791380/1030243213002412164/1213550586134265886

```
⠋ Linking AOT files...[Process.run] /Users/james/.shorebird/bin/cache/flutter/5b9d29d67adb059103beefb65710ee3dabae2f85/bin/dart run /Users/james/.shorebird/bin/cache/artifacts/aot-tools/62fa4e8bb30f56eb658989d3e6a982cb787d79d6/aot-tools.dill link --base=/var/folders/tq/pd1qk95j1hs9gfr8yws2m47w0000gn/T/TOpybX/Products/Applications/Runner.app/Frameworks/App.framework/App --patch=/Users/james/development/project/sport_mobile/build/out.aot --analyze-snapshot=/Users/james/.shorebird/bin/cache/flutter/5b9d29d67adb059103beefb65710ee3dabae2f85/bin/cache/artifacts/engine/ios-release/analyze_snapshot_arm64 --output=/Users/james/development/project/sport_mobile/build/out.vmcode (in /Users/james/development/project/sport_mobile/build)
Exited with code 253

stderr:
Can't load Kernel binary: Invalid kernel binary: Indicated size is invalid.

[2K
✗ Failed to link AOT files: Exception: Failed to link: Can't load Kernel binary: Invalid kernel binary: Indicated size is invalid.
 (0.1s)
```

Why is dart pulling from 5b9d29d67adb059103beefb65710ee3dabae2f85 but aot-tools pulling from 62fa4e8bb30f56eb658989d3e6a982cb787d79d6?

https://github.com/shorebirdtech/engine/commit/62fa4e8bb30f56eb658989d3e6a982cb787d79d6

That appears to be Dart 3.3.0, which is the same one used by Flutter 3.19.2.

/Users/james/.shorebird/bin/cache/flutter/5b9d29d67adb059103beefb65710ee3dabae2f85/bin/dart --version
Dart SDK version: 3.3.0 (stable) (Tue Feb 13 10:25:19 2024 +0000) on "macos_arm64"

bryanoltman:
## Description

Checks the response header-provided CRC32C checksum of GCP artifacts on download.

This also updates download logic to always download and unzip artifacts in a temporary location before copying them to their final artifact path (previously we would download to a temp location, but unzip directly to the artifact path).

Fixes https://github.com/shorebirdtech/shorebird/issues/1771

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel (commented):
An ideal test would somehow insert an exception during the download process and thus allow us to test that network failures, or zip failures don't end up leaving turds in our artifacts directory that would confuse us 🤷

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -162,17 +157,64 @@ abstract class CachedArtifact {
...
+      } else {
+        tempFile.renameSync(file.path);
+      }
+
+      if (!platform.isWindows && isExecutable) {
+        final result = await process.start('chmod', ['+x', file.path]);
+        await result.exitCode;
+      }
+    } catch (e) {
+      // Delete the location, so if the download is retried, it will be
```
felangel:
```suggestion
      // Delete the directory, so if the download is retried, it will be
```

bryanoltman: Done

bryanoltman:
> An ideal test would somehow insert an exception during the download process and thus allow us to test that network failures, or zip failures don't end up leaving turds in our artifacts directory that would confuse us 🤷

Added a test as part of my coverage run that makes sure artifact directories are deleted if we fail to update them.

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -162,17 +162,64 @@ abstract class CachedArtifact {
...
+
+      final isZip =
+          response.headers[HttpHeaders.contentTypeHeader] == 'application/zip';
+
+      _verifyChecksum(file: tempFile, responseHeaders: response.headers);
+
+      // Create the directory containing the artifact if it does not already
+      // exist. Failing to do this will cause [renameSync] to throw an
+      // exception.
+      final parentDirectory = file.parent..createSync(recursive: true);
```
eseidel, line 193: Will this throw if file.parent already exists?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -162,17 +162,64 @@ abstract class CachedArtifact {
...
+      if (isZip) {
+        final unzipDirectory = Directory(
+          p.join(p.dirname(tempFile.path), fileName),
```
eseidel, line 196: Is there any risk of this colliding with the existing tempFile name?  e.g. p.basename(tempFile.path) == fileName?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -162,17 +162,64 @@ abstract class CachedArtifact {
...
+        );
+        await artifactManager.extractZip(
+          zipFile: tempFile,
+          outputDirectory: unzipDirectory,
+        );
+        await copyPath(unzipDirectory.path, parentDirectory.path);
```
eseidel, line 202: Is this the final install?  If this gets interrupted here will we end up with a corrupted cache?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -191,38 +238,64 @@ allowed to access $storageUrl.''',
...
+        response.headers[HttpHeaders.contentTypeHeader] == 'application/zip';
+    final tempDirectory = Directory.systemTemp.createTempSync();
+    final tempFile = File(
+      p.join(tempDirectory.path, '$fileName${isZip ? '.zip' : ''}'),
+    );
+    await response.stream.pipe(tempFile.openWrite());
+    return tempFile;
+  }
+
+  void _verifyChecksum({
```
eseidel, line 263: This is used for two different kind of verifications it seems?  Are you sure you want to make this function have this implicit sha256 behavior?  Maybe I'm just reacting to using the "expectedChecksum" locals, and what I'm implying is pulling most of this function out into a helper and making these two branches of the if just call the helper with an explicit expectedChecksum/algorithm set?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -311,12 +372,17 @@ class BundleToolArtifact extends CachedArtifact {
...
   @override
-  String? get checksum =>
+  String? get sha256Checksum =>
       // SHA-256 checksum of the bundletool.jar file.
       // When updating the bundletool version, be sure to update this checksum.
       // This can be done by running the following command:
       // ```shell
       // shasum --algorithm 256 /path/to/file
       // ```
+      // TODO(bryanoltman): github includes `content-md5` in the response
```
erickzanardo, line 382: That is insteresting is that new? Or something undocumented? I remember when I implemented this manually generated hash I had found an issue from GitHub that its release artifact didn't had any hash 🧐

Anyway totally agree that we should prefer the value on the hash is there is one!!!

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: From the `createSync` docs:

> If the directory already exists nothing is done

erickzanardo (approved):

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Good catch, fixing now

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Not sure if it's new or undocumented or what, but if you follow the redirect (which points to https://objects.githubusercontent.com/github-production-release-asset-2e65be/132104296/9b72e02c-edf3-4618-a84f-869ac3423e42?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=releaseassetproduction%2F20240712%2Fus-east-1%2Fs3%2Faws4_request&X-Amz-Date=20240712T001036Z&X-Amz-Expires=300&X-Amz-Signature=aa5aadcee7d2a327e9643b1d17350416575f1ead1110275ed5452bd15187d188&X-Amz-SignedHeaders=host&actor_id=0&key_id=0&repo_id=132104296&response-content-disposition=attachment%3B%20filename%3Dbundletool-all-1.15.6.jar&response-content-type=application%2Foctet-stream) and make a HEAD request to that, you'll see in the headers:

```
content-md5: 2RIwqYbUqZDnu2k1G56UZA==
```

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: I'm not totally sure I follow. 

> This is used for two different kind of verifications it seems?

We can currently get checksums from two different places (although I added a TODO later on to support always getting the checksum from the header), but otherwise they're used in exactly the same manner (i.e., get the checksum from the file, make sure it matches what we expect it to).

> Are you sure you want to make this function have this implicit sha256 behavior?

By "implicit sha256 behavior", do you mean that we use the hardcoded sha256 hash and ignore the response headers if it's present? 

Maybe this is a naming thing? Would something like `_findAndVerifyChecksum` be more appropriate?

packages/shorebird_cli/lib/src/cache.dart
erickzanardo: Cool! I found the issue I mentioned: https://github.com/orgs/community/discussions/23512

Anyhow, was just curious, glad there is a better solution than manually generating

eseidel:
My overall thought is that the bug we're seeing doesn't have to do with checksumming.  It's nice that we're adding it!  But the bugs we're seeing are due to the user canceling the program during zip/download and leaving a corrupted file in the artifact cache.  So I think most of what matters (at least in terms of fixing those bugs) is making sure that we've carefully downloaded to a temp and unzipped to a temp and atomically moved, or if we can't atomically move create some record (e.g. stamp file) after the move so we can know if that record doesn't exist the move failed.

Again, not at all against the checksumming, just better explaining my earlier comment about a test for failure during the unzip/install.

packages/shorebird_cli/lib/src/cache.dart
eseidel: I think it's fine.

I think where I got wrapped up is that we have logically two separate methods.  One is "decide what type of verification to do" and the other is "do verification X" and it's currently spread across 2 methods (this one and the ChecksumVerifier below).  I probably would have sharded it slightly differently (made this one cleanly about deciding even if that meant making a 3rd method in between to do the logging + calling checksumverifier -- my sense is checksumverifier was already there and you might have factored it slightly differently if you wrote it from scratch), but I think I'm mostly tired and this factoring detail wont matter. Happy to leave the doer to decide. :)

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Yes to both. I believe our discussion on the call was that `renameSync` was not atomic. This seemed easier to read, although it's probably slower and thus more prone to being interrupted, so maybe `renameSync` is the way to go, even if it doesn't 100% guarantee that it can't be cancelled in flight. wdyt?

packages/shorebird_cli/lib/src/cache.dart
eseidel: Looks like content-md5 is even a standardized header! (note the lack of x- prefix): https://stackoverflow.com/questions/8300471/is-content-md5-field-in-the-http-response-universal

eseidel: Basically we're just sitting on top of the existing network infrastructure (which is already delivering us a correct binary) to then later verify again that the binary is still OK during usage, I think?

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Saw your comment below, disregard this

eseidel:
It occurs to me, if we're only duing this check during download, that's redundant (since the network layer already gave us a known-good body).  Only if we're caching the crc32 or md5 and using it later (during usage) could provide something new.  But I'm not actually sure that would provide new... since that cache of the md5/checksum could equally be corrupted.

I should look at this when I'm less tired 🤣 .

bryanoltman:
Closing because this adds a bunch of complexity and doesn't exactly solve the problem. Moving forward with https://github.com/shorebirdtech/shorebird/pull/2348 instead.

PR #2348, opened 2024-07-12
fix(shorebirdl_cli): add stamp file to cache artifacts to mark successful download

### LINKED ISSUE #1771: fix: validate integrity of aot_tools.dill asset on disk

bryanoltman:
## Description

Updates our cache management to write a zero-byte stamp file alongside the cache artifact when an artifact has successfully been downloaded and (optionally) extracted. We check for the existence of this file when determining that an artifact is valid. If this file does not exist, the artifact will be redownloaded.

Fixes https://github.com/shorebirdtech/shorebird/issues/1771

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -178,9 +178,7 @@ abstract class CachedArtifact {
 
   Future<void> update() async {
     // Clear any existing artifact files.
-    if (file.parent.existsSync()) {
-      await file.parent.delete(recursive: true);
-    }
+    await _delete();
 
     final request = http.Request('GET', Uri.parse(storageUrl));
     final http.StreamedResponse response;
@@ -239,6 +237,16 @@ allowed to access $storageUrl.''',
   // installed.
   void _writeStampFile() {
     stampFile.createSync(recursive: true);
+  }
+
+  Future<void> _delete() async {
+    if (file.existsSync()) {
+      await file.delete();
+    }
+
+    if (stampFile.existsSync()) {
+      await stampFile.delete();
+    }
   }
 }
 
--- a/packages/shorebird_cli/lib/src/cache.dart
+++ b/packages/shorebird_cli/lib/src/cache.dart
@@ -154,11 +154,15 @@ abstract class CachedArtifact {
     return stream.pipe(file.openWrite());
   }
 
+  /// The artifact file on disk.
   File get file =>
       File(p.join(cache.getArtifactDirectory(fileName).path, fileName));
 
+  /// Used to validate that the artifact was fully downloaded and extracted.
+  File get stampFile => File('${file.path}.stamp');
+
   Future<bool> isValid() async {
-    if (!file.existsSync()) {
+    if (!file.existsSync() || !stampFile.existsSync()) {
       return false;
     }
 
@@ -173,6 +177,11 @@ abstract class CachedArtifact {
   }
 
   Future<void> update() async {
+    // Clear any existing artifact files.
+    if (file.parent.existsSync()) {
+      await file.parent.delete(recursive: true);
+    }
+
     final request = http.Request('GET', Uri.parse(storageUrl));
     final http.StreamedResponse response;
     try {
@@ -222,6 +231,14 @@ allowed to access $storageUrl.''',
       final result = await process.start('chmod', ['+x', file.path]);
       await result.exitCode;
     }
+
+    _writeStampFile();
+  }
+
+  // Writes a 0-byte file to indicate that the artifact was successfully
+  // installed.
+  void _writeStampFile() {
+    stampFile.createSync(recursive: true);
   }
 }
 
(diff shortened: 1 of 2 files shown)


erickzanardo (approved):

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -154,11 +154,15 @@ abstract class CachedArtifact {
     return stream.pipe(file.openWrite());
   }
 
+  /// The artifact file on disk.
   File get file =>
       File(p.join(cache.getArtifactDirectory(fileName).path, fileName));
 
+  /// Used to validate that the artifact was fully downloaded and extracted.
+  File get stampFile => File('${file.path}.stamp');
```
felangel, line 162: Minor but I was under the impression that a stamp file was conventionally used to maintain a timestamp. Maybe there’s a better name here?

bryanoltman: Yeah, I was uncertain about this. I could write a timestamp to this file, but it seems a bit silly to do so if we never use it for anything.

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -173,6 +177,11 @@ abstract class CachedArtifact {
   }
 
   Future<void> update() async {
+    // Clear any existing artifact files.
+    if (file.parent.existsSync()) {
+      await file.parent.delete(recursive: true);
```
felangel: Any reason why we're using `delete` (async) instead of `deleteSync`?

packages/shorebird_cli/lib/src/cache.dart
eseidel: Looks like ninja stamp files are empty too.  It's just a "we finished task X" is all it's marking.  Then when you next wonder if you should run task X you can check to see last time it was finished and compare if it's inputs (in this case, presumably just the version of flutter or version of shorebird has changed since that time?

eseidel (approved):
Confirming, this is going to cause all customers to re-download all artifacts, correct?

packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -173,6 +177,11 @@ abstract class CachedArtifact {
   }
 
   Future<void> update() async {
+    // Clear any existing artifact files.
+    if (file.parent.existsSync()) {
```
felangel: Do we want to delete the entire parent directory? This is assuming that each artifact will be cached in its own nested directory (which I'm not sure is a valid assumption) 🤷 ?

eseidel:
Do we need to have any extra caution around deleting directories? (e.g. making sure they're under the cache?)

packages/shorebird_cli/lib/src/cache.dart
felangel: Eric looked and said ninja does something similar so seems reasonable to me 👍

felangel (approved):

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: If this takes any amount of time, the async option makes sure any progress indicators or other updates don't lock up

packages/shorebird_cli/lib/src/cache.dart
bryanoltman: Updated to only delete the artifact and stamp file.

PR #2266, opened 2024-06-18
feat(shorebird_cli): improve output for gradle "unsupported file major version" errors

erickzanardo:
## Description

This PR adds a common error handling system in the gradlew executable class and adds a handler for the `Unsupported class file major version`, which will link the user to our trouble shooting section.

This PR introduces some new concepts to the CLI so it is mostly a proposal, none of the technical choices here are strong opinionated and I am happy to change if there are better approaches!

Before this PR, this were the output of the CLI when it broken:

```
⠼ Detecting product flavors... (3.5s)Exited with code 1

stdout:
> Task :gradle:compileJava NO-SOURCE
> Task :gradle:compileGroovy FAILED
1 actionable task: 1 executed


stderr:

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':gradle:compileGroovy'.
> BUG! exception in phase 'semantic analysis' in source unit '/home/kingwill101/flutter/packages/flutter_tools/gradle/src/main/groovy/app_plugin_loader.groovy' Unsupported class file major version 65

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3s

✗ Detecting product flavors (3.6s)
Unable to extract product flavors.
Exception: > Task :gradle:compileJava NO-SOURCE
> Task :gradle:compileGroovy FAILED
1 actionable task: 1 executed


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':gradle:compileGroovy'.
> BUG! exception in phase 'semantic analysis' in source unit '/home/kingwill101/flutter/packages/flutter_tools/gradle/src/main/groovy/app_plugin_loader.groovy' Unsupported class file major version 65

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3s
```

After the PR, this is the output
Screenshot 2024-06-18 at 17 47 22

Fixes #1821 

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
[...]

PR #2352, opened 2024-07-12
feat: enhanced error logging for patch errors on windows

### LINKED ISSUE #2329: fix: give instructions when encountering exit code -1073741515
App ID: (masrpay-UserApp)

when make shorebird patch android  give me Failed to create diff (exit code -1073741515).
  stdout: 
  stderr:  (1.4s)

2024-07-09T16:34:43.386162 

C:\Users\AshrafFathalla\AppData\Roaming\shorebird\logs\1720531892399_shorebird.log

erickzanardo:
## Description

This PR enhances (in a similar approach used in the past for gradle errors https://github.com/shorebirdtech/shorebird/pull/2266) the error message when we get an error when running the patch command on windows platforms when we negative exit codes.

I weren't able to reproduce the error locally due lack of windows computer, so I couldn't test the approach other than on unit testing, but I guess the way this breaks, unit tests are good enough, but if someone could test this on a windows computer, I would appreciate!

Fixes #2329

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/executables/patch_executable.dart
+++ b/packages/shorebird_cli/lib/src/executables/patch_executable.dart
@@ -51,7 +51,8 @@ class PatchExecutable {
 
     var messageDetails = '';
 
-    // A Windows-specific error code indicating that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
+    // A Windows-specific error code indicating that the Microsoft C++ runtime
+    // (VCRUNTIME140.dll) could not be found.
     // More info: https://github.com/shorebirdtech/shorebird/issues/2329
     const vcRuntimeNotFoundExitCode = -1073741515;
     if (result.exitCode == vcRuntimeNotFoundExitCode && platform.isWindows) {
--- a/packages/shorebird_cli/lib/src/executables/patch_executable.dart
+++ b/packages/shorebird_cli/lib/src/executables/patch_executable.dart
@@ -2,6 +2,7 @@ import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/cache.dart';
+import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/shorebird_process.dart';
 
 /// A reference to a [PatchExecutable] instance.
@@ -49,9 +50,20 @@ class PatchExecutable {
     final result = await process.run(diffExecutable, diffArguments);
 
     if (result.exitCode != ExitCode.success.code) {
+      var messageDetails = '';
+
+      if (result.exitCode < 0 && platform.isWindows) {
+        messageDetails = '''
+This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
+
+The C++ Runtime can be installed from microsoft at:
+${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}
+''';
+      }
+
       throw PatchFailedException(
         '''
-Failed to create diff (exit code ${result.exitCode}).
+Failed to create diff (exit code ${result.exitCode}).$messageDetails
   stdout: ${result.stdout}
   stderr: ${result.stderr}''',
       );
--- a/packages/shorebird_cli/lib/src/executables/patch_executable.dart
+++ b/packages/shorebird_cli/lib/src/executables/patch_executable.dart
@@ -52,7 +52,7 @@ class PatchExecutable {
     if (result.exitCode != ExitCode.success.code) {
       var messageDetails = '';
 
-      if (result.exitCode < 0 && platform.isWindows) {
+      if (result.exitCode == -1073741515 && platform.isWindows) {
         messageDetails = '''
 This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
 
--- a/packages/shorebird_cli/lib/src/executables/patch_executable.dart
+++ b/packages/shorebird_cli/lib/src/executables/patch_executable.dart
@@ -51,23 +51,24 @@ class PatchExecutable {
 
     var messageDetails = '';
 
-    // This is a specific error exit code returned when the VC runtime is not
-    // found.
+    // A Windows-specific error code indicating that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
     // More info: https://github.com/shorebirdtech/shorebird/issues/2329
     const vcRuntimeNotFoundExitCode = -1073741515;
     if (result.exitCode == vcRuntimeNotFoundExitCode && platform.isWindows) {
       messageDetails = '''
-This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
+This error code indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
 
 The C++ Runtime can be installed from Microsoft at:
 ${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}
+
+Please try again once you have installed this software.
 ''';
     }
 
     if (result.exitCode != ExitCode.success.code) {
       throw PatchFailedException(
         '''
-Failed to create diff (exit code ${result.exitCode}).$messageDetails
+Failed to create diff (exit code ${result.exitCode}). $messageDetails
   stdout: ${result.stdout}
   stderr: ${result.stderr}''',
       );
--- a/packages/shorebird_cli/lib/src/executables/patch_executable.dart
+++ b/packages/shorebird_cli/lib/src/executables/patch_executable.dart
@@ -56,7 +56,7 @@ class PatchExecutable {
         messageDetails = '''
 This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
 
-The C++ Runtime can be installed from microsoft at:
+The C++ Runtime can be installed from Microsoft at:
 ${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}
 ''';
       }
--- a/packages/shorebird_cli/lib/src/executables/patch_executable.dart
+++ b/packages/shorebird_cli/lib/src/executables/patch_executable.dart
@@ -49,18 +49,22 @@ class PatchExecutable {
 
     final result = await process.run(diffExecutable, diffArguments);
 
-    if (result.exitCode != ExitCode.success.code) {
-      var messageDetails = '';
+    var messageDetails = '';
 
-      if (result.exitCode == -1073741515 && platform.isWindows) {
-        messageDetails = '''
+    // This is a specific error exit code returned when the VC runtime is not
+    // found.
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -49,9 +50,20 @@ class PatchExecutable {
     final result = await process.run(diffExecutable, diffArguments);
 
     if (result.exitCode != ExitCode.success.code) {
+      var messageDetails = '';
+
+      if (result.exitCode < 0 && platform.isWindows) {
```
eseidel: There is a specific error code which means this is missing. I'm not sure it true in the general negative case that installing that package would help.

erickzanardo: For some reason I thought the exit code on the two issues were different, so I tried to make a generic handling here.

But I now I see that they are the same code 🙃 

Updated to a specific checking

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -49,9 +50,20 @@ class PatchExecutable {
...
+      var messageDetails = '';
+
+      if (result.exitCode == -1073741515 && platform.isWindows) {
+        messageDetails = '''
+This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
+
+The C++ Runtime can be installed from microsoft at:
+${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}
+''';
+      }
```
felangel: Minor but might be nice to pull this check out above the generic result.exitCode != ExitCode.success.code to reduce the nesting and make it easier to reason about.

```dart
const vcRuntimeNotFound = -1073741515;
// Explanation of this special case with a link to the relevant GitHub issue.
if (result.exitCode == vcRuntimeNotFound && platform.isWindows) {
  // Specific exception
  throw PatchFailedException(...);
}

// Generic error handling
if (result.exitCode != ExitCode.success.code) {
  // ...
}
```

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
```diff
@@ -125,5 +131,45 @@ void main() {
         ),
       );
     });
+
+    group('when is on windows and we got a -1073741515 exit code', () {
```
felangel:
```suggestion
    group('when running on windows and process exits with code -1073741515', () {
```

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
```diff
@@ -125,5 +131,45 @@ void main() {
...
+          () => shorebirdProcess.run(
+            any(that: endsWith('patch')),
+            any(),
+            runInShell: any(named: 'runInShell'),
+          ),
+        ).thenAnswer((_) async => patchProcessResult);
+
+        when(() => platform.isWindows).thenReturn(true);
+      });
+      test('enhances the message with additional details', () async {
```
felangel:
```suggestion
      });

      test('throws a missing C++ runtime exception', () async {
```

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
felangel: I would split this into two groups:

1. When process exits with code ...
2. When running on windows

This way we can test that we handle the missing C++ runtime on windows and have a separate test where the exit code is set but the platform is not windows to ensure we only have this special error handling case for windows.

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -49,9 +50,20 @@ class PatchExecutable {
     final result = await process.run(diffExecutable, diffArguments);
 
     if (result.exitCode != ExitCode.success.code) {
+      var messageDetails = '';
+
+      if (result.exitCode == -1073741515 && platform.isWindows) {
+        messageDetails = '''
+This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
+
+The C++ Runtime can be installed from microsoft at:
```
felangel:
```suggestion
The C++ Runtime can be installed from Microsoft at:
```

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
bryanoltman: Generally, any time you write "and" in a group definition, that's a sign you probably want two groups.

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -49,9 +50,20 @@ class PatchExecutable {
     final result = await process.run(diffExecutable, diffArguments);
 
     if (result.exitCode != ExitCode.success.code) {
+      var messageDetails = '';
+
+      if (result.exitCode == -1073741515 && platform.isWindows) {
```
bryanoltman: I would pull this out into a const with a comment and a link explaining the significance of this value

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
```diff
@@ -125,5 +131,46 @@ void main() {
...
+        when(
+          () => shorebirdProcess.run(
+            any(that: endsWith('patch')),
+            any(),
+            runInShell: any(named: 'runInShell'),
+          ),
+        ).thenAnswer((_) async => patchProcessResult);
+
+        when(() => platform.isWindows).thenReturn(true);
+      });
```
bryanoltman: We might prefer using `testOn` for the group instead of overriding `platform.isWindows`

erickzanardo: I have mixed feelings about that. Cause that would make our dev experience more difficult, because we would need to have a windows machine to develop this test and feature. And I just have a mac right now.

I feel that for small things like these, mocking the platform works better and testOn would makes more sense when we are making a more complex or when a feature "needs more features" from the platform?

Happy to change if you disagree though.

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
```diff
@@ -132,44 +132,47 @@ void main() {
...
       });
-
-      test('throws a missing C++ runtime exception', () async {
-        await expectLater(
-          () => runWithOverrides(
-            () => patchExecutable.run(
-              releaseArtifactPath: 'release',
-              patchArtifactPath: 'patch',
-              diffPath: 'diff',
+      group('when the process exits with code -1073741515', () {
```
felangel: These should be inverted imo. The group for the process exit code should be first and the windows group should be nested so you can test both the windows and non-windows behavior for the exit code.

erickzanardo: Hum, but does it makes sense? This error is specific to windows...What am I missing? 😅

bryanoltman: I think Felix is suggesting that we should ensure that this path doesn't trigger for non-Windows users and that the error is otherwise handled gracefully

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
bryanoltman: Makes sense, it's ok to leave as-is

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
erickzanardo: Got it! Updated, please lmk if I understood it correctly

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -48,10 +49,25 @@ class PatchExecutable {
 
     final result = await process.run(diffExecutable, diffArguments);
 
+    var messageDetails = '';
+
+    // This is a specific error exit code returned when the VC runtime is not
+    // found.
```
bryanoltman:
```suggestion
    // A Windows-specific error code indicating that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
```

bryanoltman: Future you will always be happy for more context

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -48,10 +49,25 @@ class PatchExecutable {
...
+The C++ Runtime can be installed from Microsoft at:
+${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}
+''';
+    }
+
     if (result.exitCode != ExitCode.success.code) {
       throw PatchFailedException(
         '''
-Failed to create diff (exit code ${result.exitCode}).
+Failed to create diff (exit code ${result.exitCode}).$messageDetails
```
bryanoltman:
```suggestion
Failed to create diff (exit code ${result.exitCode}). $messageDetails
```

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -48,10 +49,25 @@ class PatchExecutable {
...
 
+    var messageDetails = '';
+
+    // This is a specific error exit code returned when the VC runtime is not
+    // found.
+    // More info: https://github.com/shorebirdtech/shorebird/issues/2329
+    const vcRuntimeNotFoundExitCode = -1073741515;
+    if (result.exitCode == vcRuntimeNotFoundExitCode && platform.isWindows) {
+      messageDetails = '''
+This indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
```
bryanoltman:
```suggestion
This error code indicates that the Microsoft C++ runtime (VCRUNTIME140.dll) could not be found.
```

packages/shorebird_cli/lib/src/executables/patch_executable.dart
```diff
@@ -48,10 +49,25 @@ class PatchExecutable {
...
+
+The C++ Runtime can be installed from Microsoft at:
+${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}
+''';
```
bryanoltman:
```suggestion
${link(uri: Uri.parse('https://www.microsoft.com/en-us/download/details.aspx?id=52685'))}

Please try again once you have installed this software.
''';
```

packages/shorebird_cli/test/src/executables/patch_executable_test.dart
```diff
@@ -125,5 +131,79 @@ void main() {
...
+          );
+        });
+      });
+
+      group('when not on windows', () {
+        setUp(() {
+          when(() => platform.isWindows).thenReturn(false);
+        });
+
+        test('does not adds the windows specific message', () async {
```
bryanoltman:
```suggestion
        test('does not add the Windows specific message', () async {
```

bryanoltman (approved):
Some minor suggestions, but LGTM!

PR #2353, opened 2024-07-12
feat: improving versions on different release versions error

### LINKED ISSUE #2349: feat: add flutter version string to different version releases message
**Description**

When adding a second release to an existing release, and accidentally using a different flutter version than the one originally used, we give a message explaining the error and reference the Flutter version just by its revision, we should also add the flutter string version (e.g. 3.22.1) to it, for improved readability.

**Additional Context**

Today:

```
....
  Existing release built with: 2f366bd7c9cdb59a1231fdb13c985be48f47149b
  Current release built with: 72996fba5d330551d694c6e09189e215c09b1c5b
```

Suggestion:

```
....
  Existing release built with: 3.22.1 (2f366bd7c9)
  Current release built with: 3.22.2 (72996fba5d)
```

erickzanardo:
## Description

When doing a release for a platform whereas it's release version is the same as an already published release for a different platform with a different flutter version, we give an error messages saying that the flutter version must be the same.

In that message we use the flutter revision to identify the flutter versions, which is not the most user friendly value.

This PR changes so it will show the flutter version along side the shortened revision. Example:

```
A release with version 4.1.4+5 already exists but was built using a different Flutter revision.


  Existing release built with: 3.22.1 (2f366bd7c9)
  Current release built with: 3.22.2 (853d13d954)

All platforms for a given release must be built using the same Flutter revision.

To resolve this issue, you can:
  * Re-run the release command with "--flutter-version=3.22.1".
  * Delete the existing release and re-run the release command with the desired Flutter version.
  * Bump the release version and re-run the release command with the desired Flutter version.

You can also file an issue if you think this is a bug. Please include the following log file in your report:
/Users/erick/Library/Application Support/shorebird/logs/1720805902724_shorebird.log
```

Fixes #2349

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_command.dart
@@ -358,15 +358,14 @@ Use `shorebird flutter versions list` to list available versions.
       // All artifacts associated with a given release must be built
       // with the same Flutter revision.
       if (existingRelease.flutterRevision != flutterRevision) {
-        final existingReleaseStringVersion =
-            await shorebirdFlutter.humanReadableVersion(
+        final existingReleaseStringVersionString =
+            await shorebirdFlutter.getVersionAndRevision(
           flutterRevision: existingRelease.flutterRevision,
-          flutterVersion: existingRelease.flutterVersion,
         );
 
-        final stringVersion = await shorebirdFlutter.humanReadableVersion(
+        final releaseVersionString =
+            await shorebirdFlutter.getVersionAndRevision(
           flutterRevision: flutterRevision,
-          flutterVersion: flutterVersionArg,
         );
 
         logger
@@ -375,8 +374,8 @@ ${styleBold.wrap(lightRed.wrap('A release with version $version already exists b
 ''')
           ..info('''
 
-  Existing release built with: ${lightCyan.wrap(existingReleaseStringVersion)}
-  Current release built with: ${lightCyan.wrap(stringVersion)}
+  Existing release built with: ${lightCyan.wrap(existingReleaseStringVersionString)}
+  Current release built with: ${lightCyan.wrap(releaseVersionString)}
 
 ${styleBold.wrap(lightRed.wrap('All platforms for a given release must be built using the same Flutter revision.'))}
 
--- a/packages/shorebird_cli/lib/src/commands/release/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_command.dart
@@ -358,19 +358,30 @@ Use `shorebird flutter versions list` to list available versions.
       // All artifacts associated with a given release must be built
       // with the same Flutter revision.
       if (existingRelease.flutterRevision != flutterRevision) {
+        final existingReleaseStringVersion =
+            await shorebirdFlutter.humanReadableVersion(
+          flutterRevision: existingRelease.flutterRevision,
+          flutterVersion: existingRelease.flutterVersion,
+        );
+
+        final stringVersion = await shorebirdFlutter.humanReadableVersion(
+          flutterRevision: flutterRevision,
+          flutterVersion: flutterVersionArg,
+        );
+
         logger
           ..err('''
 ${styleBold.wrap(lightRed.wrap('A release with version $version already exists but was built using a different Flutter revision.'))}
 ''')
           ..info('''
 
-  Existing release built with: ${lightCyan.wrap(existingRelease.flutterRevision)}
-  Current release built with: ${lightCyan.wrap(flutterRevision)}
+  Existing release built with: ${lightCyan.wrap(existingReleaseStringVersion)}
+  Current release built with: ${lightCyan.wrap(stringVersion)}
 
 ${styleBold.wrap(lightRed.wrap('All platforms for a given release must be built using the same Flutter revision.'))}
 
 To resolve this issue, you can:
-  * Re-run the release command with "${lightCyan.wrap('--flutter-version=${existingRelease.flutterRevision}')}".
+  * Re-run the release command with "${lightCyan.wrap('--flutter-version=${existingRelease.flutterVersion ?? existingRelease.flutterRevision}')}".
   * Delete the existing release and re-run the release command with the desired Flutter version.
   * Bump the release version and re-run the release command with the desired Flutter version.''');
         throw ProcessExit(ExitCode.software.code);
--- a/packages/shorebird_cli/lib/src/shorebird_flutter.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_flutter.dart
@@ -224,7 +224,7 @@ class ShorebirdFlutter {
   ///
   /// If [flutterRevision] is null, the current revision will be used.
   /// If [flutterVersion] is null, it will be calculated from the current
-  /// revision using [getVersionString].
+  ///   revision using [getVersionString].
   Future<String> humanReadableVersion({
     String? flutterRevision,
     String? flutterVersion,
--- a/packages/shorebird_cli/lib/src/shorebird_flutter.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_flutter.dart
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
     shorebirdEnv.flutterRevision = revision;
     useFlutterProgress.complete();
   }
+
+  /// Returns an human readable string from the received [flutterRevision] and
+  /// [flutterVersion].
+  ///
+  /// If [flutterRevision] is null, the current revision will be used.
+  /// If [flutterVersion] is null, it will be calculated from the current
+  /// revision using [getVersionString].
+  Future<String> humanReadableVersion({
+    String? flutterRevision,
+    String? flutterVersion,
+  }) async {
+    final revision = flutterRevision ?? shorebirdEnv.flutterRevision;
+    var version = 'unknown';
+
+    if (flutterVersion != null) {
+      version = flutterVersion;
+    } else {
+      try {
+        final calculatedValue = await getVersionString(revision: revision);
+        if (calculatedValue != null) {
+          version = calculatedValue;
+        }
+      } catch (_) {}
+    }
+
+    return '$version (${shortRevisionString(revision)})';
+  }
 }
--- a/packages/shorebird_cli/lib/src/shorebird_flutter.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_flutter.dart
@@ -135,13 +135,14 @@ class ShorebirdFlutter {
 
   /// Returns the current Shorebird Flutter version and revision.
   /// Returns unknown if the version check fails.
-  Future<String> getVersionAndRevision() async {
+  Future<String> getVersionAndRevision({String? flutterRevision}) async {
+    final revision = flutterRevision ?? shorebirdEnv.flutterRevision;
     String? version = 'unknown';
     try {
-      version = await getVersionString();
+      version = await getVersionString(revision: revision);
     } catch (_) {}
 
-    return '$version (${shortRevisionString(shorebirdEnv.flutterRevision)})';
+    return '$version (${shortRevisionString(revision)})';
   }
 
   /// Returns the current Shorebird Flutter version.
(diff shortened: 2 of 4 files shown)


packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
...
     useFlutterProgress.complete();
   }
+
+  /// Returns an human readable string from the received [flutterRevision] and
+  /// [flutterVersion].
+  ///
+  /// If [flutterRevision] is null, the current revision will be used.
+  /// If [flutterVersion] is null, it will be calculated from the current
+  /// revision using [getVersionString].
+  Future<String> humanReadableVersion({
```
bryanoltman:
```suggestion
  /// If [flutterRevision] is null, the current revision will be used.
  /// If [flutterVersion] is null, it will be calculated from the current
  ///   revision using [getVersionString].
  Future<String> humanReadableVersion({
```

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
     shorebirdEnv.flutterRevision = revision;
     useFlutterProgress.complete();
   }
+
+  /// Returns an human readable string from the received [flutterRevision] and
+  /// [flutterVersion].
+  ///
+  /// If [flutterRevision] is null, the current revision will be used.
+  /// If [flutterVersion] is null, it will be calculated from the current
+  /// revision using [getVersionString].
```
bryanoltman: What happens if both are null?

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
     shorebirdEnv.flutterRevision = revision;
     useFlutterProgress.complete();
   }
+
+  /// Returns an human readable string from the received [flutterRevision] and
+  /// [flutterVersion].
```
bryanoltman: An example of the returned string would be helpful here.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
...
+  ///
+  /// If [flutterRevision] is null, the current revision will be used.
+  /// If [flutterVersion] is null, it will be calculated from the current
+  /// revision using [getVersionString].
+  Future<String> humanReadableVersion({
+    String? flutterRevision,
+    String? flutterVersion,
+  }) async {
+    final revision = flutterRevision ?? shorebirdEnv.flutterRevision;
```
bryanoltman: Defaulting to the current revision is unexpected to me based on this function's API (e.g., if I passed in `flutterRevision: null, flutterVersion: '1.0.0`, it would tell me that the revision for `1.0.0` is the current revision, which is wrong). I would consider making flutterRevision required. Is there a case I'm not thinking of where we wouldn't have a value for `flutterRevision`?

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
...
+    if (flutterVersion != null) {
+      version = flutterVersion;
+    } else {
+      try {
+        final calculatedValue = await getVersionString(revision: revision);
+        if (calculatedValue != null) {
+          version = calculatedValue;
+        }
+      } catch (_) {}
+    }
```
bryanoltman:
```suggestion
    final String resolvedFlutterVersion;
    try {
      resolvedFlutterVersion = flutterVersion ?? await getVersionString(revision: revision);
    } catch (_) {
      resolvedFlutterVersion = 'unknown';
    }
```

bryanoltman (changes requested):

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: This will not work, `getVersionString` can `null`. 

The current choice in the source base was because I didn't wanted to repeat the string `unknown`

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: I don't think so, the choice I made were to keep this API consistent to the other methods in this class, where when you don't receive a revision, we fallback to the one in the env.

I agree that this can be a bit confusing though, so I guess better to simplify and make it always required?

packages/shorebird_cli/lib/src/shorebird_flutter.dart
bryanoltman: New suggestion:

```suggestion
    final String resolvedFlutterVersion;
    try {
      resolvedFlutterVersion = flutterVersion ?? await getVersionString(revision: revision) ?? 'unknown';
    } catch (_) {
      resolvedFlutterVersion = 'unknown';
    }
```

Even if we have to repeat unknown, simplifying the control flow makes this a ton easier to read.

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -358,19 +358,30 @@ Use `shorebird flutter versions list` to list available versions.
       // All artifacts associated with a given release must be built
       // with the same Flutter revision.
       if (existingRelease.flutterRevision != flutterRevision) {
+        final existingReleaseStringVersion =
```
felangel:
```suggestion
        final existingReleaseVersionString =
```

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -358,19 +358,30 @@ Use `shorebird flutter versions list` to list available versions.
...
+            await shorebirdFlutter.humanReadableVersion(
+          flutterRevision: existingRelease.flutterRevision,
+          flutterVersion: existingRelease.flutterVersion,
+        );
+
+        final stringVersion = await shorebirdFlutter.humanReadableVersion(
```
felangel:
```suggestion
        final releaseVersionString = await shorebirdFlutter.humanReadableVersion(
```

packages/shorebird_cli/lib/src/shorebird_flutter.dart
felangel: When would the revision ever be null?

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +218,31 @@ class ShorebirdFlutter {
...
     useFlutterProgress.complete();
   }
+
+  /// Returns an human readable string from the received [flutterRevision] and
+  /// [flutterVersion].
+  ///
+  /// If [flutterRevision] is null, the current revision will be used.
+  /// If [flutterVersion] is null, it will be calculated from the current
+  ///   revision using [getVersionString].
+  Future<String> humanReadableVersion({
```
felangel: Why does this accept a flutterRevision and flutterVersion instead of using the environment?

packages/shorebird_cli/lib/src/shorebird_flutter.dart
felangel: We can also pull `unknown` into a constant 🤷

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: as discussed on vc `flutterRevision` is now required.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: Added one! Lmk what you think

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: as discussed on vc `flutterRevision` is now required.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: I didn't changed to the suggested code, but I did change to something which I believe is better, let me know with you agree.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
erickzanardo: I've changed for it to require the `flutterRevision`, so it will never take from the env.

The flutter version is optional because the user of this function might already have that value available (like when dealing with an `AppRelease`, so it can be passed and when that is done, this method will not try to "calculate" the version from the revision.

erickzanardo:
@bryanoltman I merged those both methods and refactored in the places where the old method were used to use the new name.

I believe there are code improvements that we can do here, like for example `getVersionString` could also require the flutter revision instead of getting it from the env, and probably be renamed to `getVersionFromRevision`, so it would match the `getRevisionFromVersion`.

But I would argue that we should do that refactor in a follow up to keep the size of this PR in a manageable one.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
bryanoltman: The branching code looks to be the same as it was when I originally reviewed. I still prefer what I suggested.

bryanoltman: To elaborate a bit, the reasons I prefer the non-branching option:

1. There is less cognitive load on the reader due to `resolvedFlutterVersion` being final and only to assigned once.
2. It's shorter
3. It makes the fallback that's happening from flutterVersionArg -> getVersionString -> 'unknown' much more explicit.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +207,36 @@ class ShorebirdFlutter {
...
+  ///
+  /// e.g. 3.16.3 (b9b2390296)
+  ///
+  /// If [flutterVersion] is null, it will be calculated from the current
+  ///   revision using [getVersionString].
+  Future<String> humanReadableVersion({
+    required String flutterRevision,
+    String? flutterVersion,
+  }) async {
+    var version = 'unknown';
```
felangel:
```suggestion
    var version = flutterVersion ?? 'unknown';
```

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +207,36 @@ class ShorebirdFlutter {
...
+  }
+
+  /// Returns an human readable string from the received [flutterRevision] and
+  /// [flutterVersion].
+  ///
+  /// e.g. 3.16.3 (b9b2390296)
+  ///
+  /// If [flutterVersion] is null, it will be calculated from the current
+  ///   revision using [getVersionString].
+  Future<String> humanReadableVersion({
```
felangel: Minor but looks like we renamed `getVersionAndRevision` to `humanReadableVersion`. I personally think the previous name was more descriptive of what this method does 🤷

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -218,4 +207,36 @@ class ShorebirdFlutter {
...
     useFlutterProgress.complete();
   }
+
+  /// Helper method to return the human readable version from the
+  /// current flutter revision in [ShorebirdEnv.flutterRevision].
+  Future<String> humanReadableVersionFromEnv() async {
+    return humanReadableVersion(
+      flutterRevision: shorebirdEnv.flutterRevision,
+    );
+  }
```
felangel: What's the value of having an extra level of indirection here? I'd delete this helper method and just pass `shorebirdEnv.flutterRevision` when needed.

bryanoltman:
Disabling automerge to make sure everyone's comments are addressed

packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
```diff
@@ -76,7 +76,8 @@ class AarReleaser extends Releaser {
 
   @override
   Future<FileSystemEntity> buildReleaseArtifacts() async {
-    final flutterVersionString = await shorebirdFlutter.getVersionAndRevision();
+    final flutterVersionString =
+        await shorebirdFlutter.humanReadableVersionFromEnv();
```
felangel: I'd advocate for reverting the naming change since the previous name was more descriptive and it would dramatically reduce the diff 🤷

packages/shorebird_cli/lib/src/shorebird_flutter.dart
felangel: The API is pretty confusing imo. If we already have the version and the revision we shouldn't need to call this method at all right?

felangel: I think this is all a sign that we probably have the wrong abstraction. The different methods we need are:

1. Get the Flutter Version
2. Get the Flutter Revision
3. Format the Flutter Version and Flutter Revision into a human-readable string.

Right now this one method is attempting to do too much imo.

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -358,19 +358,29 @@ Use `shorebird flutter versions list` to list available versions.
...
       if (existingRelease.flutterRevision != flutterRevision) {
+        final existingReleaseStringVersionString =
+            await shorebirdFlutter.getVersionAndRevision(
+          flutterRevision: existingRelease.flutterRevision,
+        );
+
+        final releaseVersionString =
+            await shorebirdFlutter.getVersionAndRevision(
+          flutterRevision: flutterRevision,
+        );
```
felangel, line 369: Can't we revert `getVersionAndRevision` to not accept an optional `flutterRevision` and instead in this case get the Flutter version for the revision and output the formatted string?

erickzanardo:
Closing in favor of #2354

PR #2370, opened 2024-07-19
chore: remove was_forced from CreatePatchRequest

bryanoltman:
## Description

We haven't supported the `--force` flag in our patch command for quite a while, so this is no longer a useful field.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [x] 🗑️ Chore
- [ ] 🧪 Tests


eseidel (approved):

felangel (changes requested):
Won’t we need to bump the code push client version since this is a breaking change in the protocol?

bryanoltman:
> Won’t we need to bump the code push client version since this is a breaking change in the protocol?

We can, but it's not necessary. With https://github.com/shorebirdtech/_shorebird/pull/737, we no longer read this field.

felangel (approved):

PR #2394, opened 2024-07-29
feat(shorebird_cli): require Flutter 3.22.3 for new iOS releases

felangel:
## Description

- feat(shorebird_cli): require Flutter 3.22.3 for new iOS releases
  - this change results in users still being allowed to patch releases made with older Flutter versions on iOS

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -25,7 +25,6 @@ import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/platform/platform.dart';
 import 'package:shorebird_cli/src/release_type.dart';
 import 'package:shorebird_cli/src/shorebird_artifacts.dart';
-import 'package:shorebird_cli/src/shorebird_documentation.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 import 'package:shorebird_cli/src/shorebird_flutter.dart';
 import 'package:shorebird_cli/src/shorebird_validator.dart';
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -146,20 +145,8 @@ This may indicate that the patch contains native changes, which cannot be applie
...
-
-      if ((flutterVersion ?? minimumSupportedIosFlutterVersion) <
-          minimumSupportedIosFlutterVersion) {
-        logger.err(
-          '''
-iOS patches are not supported with Flutter versions older than $minimumSupportedIosFlutterVersion.
-For more information see: ${supportedFlutterVersionsUrl.toLink()}''',
-        );
-        throw ProcessExit(ExitCode.software.code);
-      }
```
bryanoltman, line 162: Did we decide to open up patching for all iOS versions? This seems like it could be dangerous

felangel: Seems super restrictive to only allow patching and releasing using the latest stable version on iOS (when we introduced this customers were pretty frustrated) but happy to revert this 🤷

felangel: I personally think it's ok for folks to patch older releases especially since they can stage and preview the patches 🤷 
It causes more harm than good to be so restrictive imo since we disincentivize engagement/usage.

eseidel:
After discussion I think we should just live with the support cost for now.

Yes, some users will release with some versions of 3.22.2 and encounter bugs.  But fewer users will end up having to take action to move themselves to 3.22.3.

We could do a more surgical fix of making an exclusion-list for the known-bad revisions of 3.22.2, but for now I think it's easier to just punt.

PR #2430, opened 2024-08-19
fix(shorebird_code_push_protocol): make current period start/end nonnull

bryanoltman:
## Description

Makes currentPeriodStart and currentPeriodEnd nonnull. These will be always populated by the backend (free plans will use the start of this month and next month for start and end, respectively). 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel (approved):
I guess the DB was already this way so it doesn't need a migration?  If so, lgtm.

bryanoltman:
> I guess the DB was already this way so it doesn't need a migration? If so, lgtm.

We don't store plans in the DB—these are 100% derived from Stripe subscriptions.

PR #2499, opened 2024-09-27
feat: created pared-down version of User object

bryanoltman:
## Description

Renames our current User object to "FullUser" and creates a new User object with only id, email, and display name.

This is to ensure we don't accidentally leak any sensitive information when returning user data to a different user (for example, if user A receives a response that references users B, C, etc.)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_code_push_protocol/lib/src/models/full_user.dart
```diff
@@ -0,0 +1,76 @@
...
+
+part 'full_user.g.dart';
+
+/// {@template user}
+/// A fully-detailed user object, possibly including sensitive information.
+/// This should only be used when querying the user's own information. For other
+/// users, use [User].
+/// {@endtemplate}
+@JsonSerializable()
+class FullUser {
```
felangel: For consistency would it make sense to use UserMetata here similar to AppMetadata?

bryanoltman: Happy to adjust the naming, although what we're communicating here isn't really metadata, just a slimmed-down version of a full object.

felangel (commented):
What’s the goal of this change? Where is User used? Also this is a breaking change to the protocol so we should probably bump the version etc if we want to proceed.

packages/shorebird_code_push_protocol/lib/src/models/full_user.dart
felangel: In that case it seems like it might be better to leave User as is and introduce a new class specifically for the new use case (I don’t think I have enough context here to suggest a name)

bryanoltman:
@felangel this is addressing the concern expressed in https://github.com/shorebirdtech/_shorebird/pull/848#discussion_r1777771115. Updated the PR description to explain better.

packages/shorebird_code_push_protocol/lib/src/models/full_user.dart
bryanoltman: The desire is to avoid accidentally leaking sensitive user info by including a User in a response object. Given a choice between PublicUser/User and User/PrivateUser, I feel like the former (i.e., the default "User" object containing no sensitive information) is less of a footgun.

felangel: Thanks for the context! I’m fine with it just feel like the naming could be improved 🤷 

PrivateUser vs PublicUser seems better and more explicit based on the goals of this PR imo

bryanoltman: Done

PR #2514, opened 2024-10-07
feat: allow user to choose organization in shorebird init

bryanoltman:
## Description

When running `shorebird init`, ask the user to specify an organization for the app if the user is a member of multiple organizations.

Screenshot 2024-10-07 at 5 49 58 PM

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -105,6 +105,18 @@ class CodePushClientWrapper {
     );
   }
 
+  /// Returns the currently logged in user, or null if no user is logged in.
+  Future<PrivateUser?> getCurrentUser() async {
+    final progress = logger.progress('Fetching user');
+    try {
+      final user = await codePushClient.getCurrentUser();
+      progress.complete();
+      return user;
+    } catch (error) {
+      _handleErrorAndExit(error, progress: progress);
+    }
+  }
+
   Future<List<OrganizationMembership>> getOrganizationMemberships() async {
     final progress = logger.progress('Fetching organizations');
     final List<OrganizationMembership> memberships;
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -81,8 +81,8 @@ class CodePushClientWrapper {
   final CodePushClient codePushClient;
 
   Future<App> createApp({
+    required int organizationId,
     String? appName,
-    int? organizationId,
   }) async {
     late final String displayName;
     if (appName == null) {
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -80,7 +80,10 @@ class CodePushClientWrapper {
 
   final CodePushClient codePushClient;
 
-  Future<App> createApp({String? appName}) async {
+  Future<App> createApp({
+    String? appName,
+    int? organizationId,
+  }) async {
     late final String displayName;
     if (appName == null) {
       String? defaultAppName;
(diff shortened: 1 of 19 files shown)


packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -65,6 +66,26 @@ Please make sure you are running "shorebird init" from within your Flutter proje
...
     }
 
+    final orgs = await codePushClientWrapper.getOrganizationMemberships();
+    if (orgs.isEmpty) {
+      logger.err('''
+You do not have any organizations.
+Please create an organization on the Shorebird website before continuing.
+''');
+      return ExitCode.software.code;
+    }
```
eseidel: Is this even possible?  Or does the "personal" organization not show up in the "organization memberships" list?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -80,7 +80,10 @@ class CodePushClientWrapper {
 
   final CodePushClient codePushClient;
 
-  Future<App> createApp({String? appName}) async {
+  Future<App> createApp({
+    String? appName,
+    int? organizationId,
```
eseidel: If the personal organization is exposed, I'd recommend making this required, to make sure we have updated all callers to pass the organization around. 🤷

packages/shorebird_cli/lib/src/commands/init_command.dart
bryanoltman: It shouldn't be, but I wanted to be sure we give people clear instructions in case it somehow happens. 

This is a draft, btw, so I wouldn't spend too much time on it. Will let you know when it's ready.

bryanoltman:
I could see wanting to rephrase "Personal" as "(None)" or similar, depending on if we expect users to think of "apps in personal organizations" as "apps without an organization"

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -65,6 +66,25 @@ Please make sure you are running "shorebird init" from within your Flutter proje
       return ExitCode.software.code;
     }
 
+    final orgs = await codePushClientWrapper.getOrganizationMemberships();
+    if (orgs.isEmpty) {
+      logger.err(
+        '''You do not have any organizations. This should never happen. Please contact us on Discord or send us an email at contact@shorebird.dev.''',
```
eseidel, line 72: That method on the backend could also do this log. 🤷

felangel:
> I could see wanting to rephrase "Personal" as "(None)" or similar, depending on if we expect users to think of "apps in personal organizations" as "apps without an organization"

I would probably list their display name (email fallback) to stay consistent with the console

eseidel (approved):

packages/shorebird_cli/lib/src/extensions/organization.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:shorebird_code_push_protocol/shorebird_code_push_protocol.dart';
+
+/// {@template organization_display}
+/// Returns the user-facing display name of an [Organization].
+/// {@endtemplate}
+extension OrganizationDisplay on Organization {
+  /// {@macro organization_display}
+  String get displayName => switch (organizationType) {
+        OrganizationType.team => name,
+        OrganizationType.personal => 'Personal',
```
felangel: I’d use the user display name and fallback to email instead of “Personal” here to be consistent with the console.

packages/shorebird_cli/lib/src/commands/init_command.dart
bryanoltman: It probably makes sense to do both, yeah

packages/shorebird_cli/lib/src/extensions/organization.dart
bryanoltman: Screenshot 2024-10-08 at 2 54 48 PM

So something like this, you think?

eseidel:
Prompt lgtm

eseidel (approved):

packages/shorebird_cli/lib/src/extensions/organization.dart
felangel: Yup 👍

felangel (approved):

PR #2515, opened 2024-10-08
feat: add OrganizationRole.none

bryanoltman:
## Description

Adds the "none" role to OrganizationMembership for cases where a user has visibility into an organization but is not a member.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/models/organization.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/models/organization.dart
@@ -15,7 +15,11 @@ enum OrganizationRole {
   admin,
 
   /// Users who are part of the organization but have limited permissions.
-  member;
+  member,
+
+  /// Users who are not part of the organization but have visibility into it via
+  /// app collaborator permissions.
+  none;
 }
 
 /// {@template organization_type}
--- a/packages/shorebird_code_push_protocol/lib/src/models/organization_membership.g.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/models/organization_membership.g.dart
@@ -35,4 +35,5 @@ const _$OrganizationRoleEnumMap = {
   OrganizationRole.owner: 'owner',
   OrganizationRole.admin: 'admin',
   OrganizationRole.member: 'member',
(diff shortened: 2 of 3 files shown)


packages/shorebird_code_push_protocol/lib/src/models/organization.dart
```diff
@@ -15,7 +15,11 @@ enum OrganizationRole {
   admin,
 
   /// Users who are part of the organization but have limited permissions.
-  member;
+  member,
+
+  /// Users who are not part of the organization but have visibility into it via
+  /// app collaborator permissions.
+  none;
```
felangel, line 22: Thoughts on `limited` instead? I think none is a bit confusing since the user would have the org in their org membership but the role would be none.

bryanoltman: I prefer none—they don't have a role in the org. Limited raises a bunch of questions (limited how? how is limited different than member as it relates to admin)

felangel: They technically have a role which allows them to see the org name

felangel: Maybe external? Or guest are better names. Either way I’ll leave it to you

felangel (approved):

packages/shorebird_code_push_protocol/lib/src/models/organization.dart
bryanoltman: > They technically have a role which allows them to see the org name

They don't have an organization role, they have a role as a collaborator, which grants them permission to see parts of the organization.

felangel: Fine by me. Still feels weird to have no organization role yet still have the organization returned via the GET organization membership endpoint but we can revisit the naming later 🤷

Issue #2532, opened 2024-10-15
fix: patch artifact download takes much longer than it should on Windows


bryanoltman:
**Description**

Users have reported extremely long (thousands of seconds) download times of release artifacts when creating a patch on Windows.

Example: https://discord.com/channels/1030243211995791380/1293125397394751581

This seems to be mostly affecting Windows users but I'm not confident it is a Windows-only issue.

bryanoltman:
We've released an update to the `shorebird` tool with better diagnostics (v1.4.1). If anyone experiencing this can run `shorebird doctor -v` with this new version and share the output, that would be very helpful.

bryanoltman:
I'm currently able to reproduce this on my windows machine. I:

1. Downgraded to Flutter 3.19.6 as my non-Shorebird flutter
2. Created a new app
3. Ran `shorebird release android` (no issues)
4. Ran `shorebird patch android` and am seeing very very slow (but progressing) downloads.

I'm going to try re-upgrading my flutter to see if that changes anything.

Sample output:
```
PS C:\Users\bryan\Documents\sandbox\windows_old_flutter_test
$ shorebird patch android
✓ Fetching apps (0.2s)
✓ Fetching releases (0.1s)
Which release would you like to patch?
Which release would you like to patch? 1.0.0+1
✓ Fetching aab artifact (0.1s)
✓ Downloading aab (100%) (1.8s)
✓ Building patch with Flutter 3.24.3 (5a6dc546e8) (8.7s)
✓ Verifying patch can be applied to release (35ms)
✓ Fetching release artifacts (4.7s)
⠦ Downloading release artifact 1/3 (28%)... (135.4s)
```

I killed the process shortly after this. The progress percentage had been slowly creeping up.

bryanoltman:
The issue above was using powershell. Can confirm that git bash does not have this issue (patching works as expected).

bryanoltman:
Frustratingly, after trying this in git bash and seeing that it worked fine there, I'm not able to reproduce it in Powershell

bryanoltman:
Correction: if I don't re-upgrade, the issue is reproducible in both powershell and git bash

bryanoltman:
AAB url, from which I have no trouble downloading a file:

[...]

PR #2563, opened 2024-10-21
feat: show download progress when downloading patch artifacts

bryanoltman:
## Description

Updates `ArtifactManager.downloadFile` to accept an `onProgress` callback, which is used by the patch command and the android patcher to display download progress.

I hope for this to be helpful in diagnosing https://github.com/shorebirdtech/shorebird/issues/2532

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel (approved):
I think we tried this style of on-progress before and found it didn't work as expected (since the stream will get read in one big chunk when reading from disk iirc?), but printing here is still better than nothing and hopefully will help us diagnose.

bryanoltman:
After testing this on a slow connection, the frequent progress updates cause an unpleasant flickering in the terminal. Going to address that in a followup PR.

Issue #2549, opened 2024-10-16
fix: `shorebird preview --app-id {flavor_id}` downloads and runs wrong flavor on iOS


AhmedLSayed9:
App ID: (2c13f029-59fa-4e67-a044-4e9ecf3e33d9)

**Prerequisites**

I've 3 flavors (Prod - Dev - Staging)

**Steps To Reproduce:**

1. Release then run `shorebird preview --app-id {flavor_id}` with flavor_id of Dev or Staging
2. It runs the correct flavor Dev/Staging.
3. Release then run `shorebird preview --app-id {flavor_id}` with flavor_id of Prod
4. It runs the correct flavor Prod.
5. Run `shorebird preview --app-id {flavor_id}` with flavor_id of Dev or Staging again
6. it somehow downloads and runs `Prod` flavor (even if it's deleted from Shorebird Console) and I can't preview Dev/Staging again.

It only occurs on iOS, while on Android it previews the correct flavor.

I've tested this on both simulator and real device and tested it on 2 different apps.

eseidel:
Thanks for the report, will look.

felangel:
`shorebird preview --app-id` takes the actual app id not the flavor name (e.g Dev or Staging). Let me know if that helps!

AhmedLSayed9:
> `shorebird preview --app-id` takes the actual app id not the flavor name (e.g Dev or Staging). Let me know if that helps!

Yes I'm using the app id for the flavor (from shorebird.yaml), not the name itself.

bryanoltman:
Do you have this problem if you run `shorebird preview` without providing the `--app-id`? e.g., if you select dev or staging when prompted by the preview command, does it preview the flavor you select?

AhmedLSayed9:
> Do you have this problem if you run `shorebird preview` without providing the `--app-id`? e.g., if you select dev or staging when prompted by the preview command, does it preview the flavor you select?

I didn't know that I'd be prompted If I didn't provide `--app-id` in advance. I'll try it now.

bryanoltman:
Interesting! Do you remember why you thought you needed to provide the `--app-id` argument? There might be a place in our docs that needs to be clarified

AhmedLSayed9:
> Interesting! Do you remember why you thought you needed to provide the `--app-id` argument? There might be a place in our docs that needs to be clarified

https://docs.shorebird.dev/guides/flavors/android/#preview-the-release
[...]

PR #2577, opened 2024-10-24
fix: choose newest xcarchive in ios release

### LINKED ISSUE #2575: fix: shorebird taking build info from the wrong xcarchive for iOS release
App ID: bbbc22d3-7147-4c9e-85e9-ae138aed281d

The Shorebird CLI is pulling build information from the wrong .xcarchive file when `${PRODUCT_NAME}` is set, and different `.xcarchive` files are generated based on the build scheme in Xcode. This causes commands like `shorebird release ios --flavor staging --build-name=1.1.3 --build-number=2` to display the incorrect build name and build number when Shorebird attempts to upload the data to the Shorebird Console. The issue occurs because the Shorebird CLI uses the `production.xcarchive` build information instead of the `staging.xcarchive` build that was just built with the command.

**Steps To Reproduce**

1. Set the ${PRODUCT_NAME} in the Xcode settings for a `staging` scheme.
2. Run shorebird release with the production flavor and specify `--build-name=1.0.0 --build-number=1`.
3. Run shorebird release with the staging flavor and specify `--build-name=1.0.1 --build-number=1`.
4. Shorebird will display a confirmation of the build name and build number before proceeding with the upload to the Shorebird Console.
5. The displayed build name and build number do not match what was specified for the staging build. Instead, the production.xcarchive is used, resulting in 1.0.0+1.

**Expected Behavior**

The build name and build number should match the values specified in the arguments for the `staging` build.

**Screenshots**
The screenshots here doesn't use the versions mentioned in the steps.

1. The build was generated with the specified `build-name` and `build-number`, but the wrong release version was shown and uploaded.
image
<br/><br/>
2. The `production.xcarchive` here contains the release version used by the Shorebird CLI, but this was not built using the command shown above.<br/><br/>
image
<br/><br/>
3. The `staging.xcarchive` here contains the correct release version. The `--verbose` flag confirms that the Flutter build command is executed correctly, but when Shorebird uploads the artifact, it uses the `production.xcarchive` version.<br/><br/>
image

**Additional Context**

None

bryanoltman:
## Description

Updates the ArtifactManager's `getXcarchive` function to return the most recently edited xcarchive rather than the first that appears in `Directory.listSync`

Fixes https://github.com/shorebirdtech/shorebird/issues/2575 and https://github.com/shorebirdtech/shorebird/issues/2549

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -282,6 +282,11 @@ class ArtifactManager {
     return archiveDirectory
         .listSync()
         .whereType<Directory>()
+        // Get the most recently modified xcarchive to handle cases where an app
+        // may produce multiple xcarchives with different names.
```
eseidel, line 286:
```suggestion
        // may produce multiple xcarchives with different names.
        // This still could grab the wrong ipa, if multiple `flutter` commands
        // are running in parallel or the clock is/was broken on the machine.
        // If either of those occurs in the wild we can check the contents of the
        // xcarchives, but this should be good enough for now.
```

eseidel (approved):

felangel (approved):

packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -282,6 +282,15 @@ class ArtifactManager {
...
+        // This still could grab the wrong ipa, if multiple `flutter` commands
+        // are running in parallel or the clock is/was broken on the machine.
+        // If either of those occurs in the wild we can check the contents of
+        // the xcarchives, but this should be good enough for now.
```
AhmedLSayed9, line 290: @bryanoltman 
Isn't this crucial?
As the bug in the `shorebird release` itself, this will break patching too not just previewing. Developer will get a confirmation of the patch successfully published but it'll never be installed for app users.

bryanoltman: It is indeed crucial. We believe that it is a "good enough" fix, in that it will solve the problem in 99.9% of cases, except those outlined in the comment above.

AhmedLSayed9: Understood. Thanks!

PR #2607, opened 2024-11-05
feat: adds an extremely simply AppCapabilities model to support percentage-based rollouts

bryanoltman:
## Description

This is just to support percentage-based rollouts. We will likely want to revamp this system to a more general actor+subject+capability model down the road, but I don't think we know enough about our needs for such a system to correctly design it now.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/models/app_capability.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/models/app_capability.dart
@@ -1,0 +1,7 @@
+/// Operations that a user may or may not be able to perform to or in relation
+/// to an app.
+enum AppCapability {
+  /// Allows the user to create patches for an app at a non-100 rollout
+  /// percentage.
+  phasedPatchRollout,
+}
--- a/packages/shorebird_code_push_protocol/lib/src/messages/get_app_capabilities/get_app_capabilities.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/get_app_capabilities/get_app_capabilities.dart
@@ -1,0 +1,1 @@
+export 'get_app_capabilities_response.dart';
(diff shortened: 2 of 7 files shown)


packages/shorebird_code_push_protocol/lib/src/models/app_capability.dart
```diff
@@ -0,0 +1,7 @@
+/// Operations that a user may or may not be able to perform to or in relation
+/// to an app.
+enum AppCapability {
+  /// Allows the user to create patches for an app at a non-100 rollout
+  /// percentage.
+  phasedPatchRollout,
```
eseidel, line 6: This will have an unstable serialization, since it will serialize as the enum name 🤷

eseidel: That's probably fine for now.

felangel:
What problem is this trying to solve? Is this just to answer the question of can I specify a rollout percentage for a given patch? Shouldn't that be possible to derive from the plan associated with the app's organization?

This seems tangentially related to the design doc I'm working on for a role based access control system.

bryanoltman:
> What problem is this trying to solve? Is this just to answer the question of can I specify a rollout percentage for a given patch?

Yes

> Shouldn't that be possible to derive from the plan associated with the app's organization?

Yep, that's how it will be implemented on the backend. We don't have an endpoint to get a plan for an organization (we only have an endpoint for the current user's plan), and this seemed simpler/faster/more secure than exposing the full plan, given that the user might not be the plan owner.

> This seems tangentially related to the design doc I'm working on for a role based access control system.

Great! I imagine we'll rip this out before too long, this was just the fastest, least-invasive route from where I am to where I want to be

bryanoltman:
Rolling back as per today's decision to rework this as part of `package:shorebird_code_push`

PR #2608, opened 2024-11-07
feat: replace --staging with --channel in patch command

bryanoltman:
## Description

Updates the patch command to accept the `--channel` argument (which defaults to `stable`). Currently accepted values are `staging` and `stable`. This replaces `--staging`, which now prints a deprecation error.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -86,10 +86,18 @@ of the iOS app that is using this module.''',
         help: allowAssetDiffsHelpText,
         negatable: false,
       )
+      ..addOption(
+        'channel',
+        allowed: DeploymentTrack.values.map((v) => v.channel),
+        help: 'The channel to publish the patch to.',
+        defaultsTo: DeploymentTrack.production.channel,
+      )
       ..addFlag(
         'staging',
         negatable: false,
-        help: 'Whether to publish the patch to the staging environment.',
+        help: '''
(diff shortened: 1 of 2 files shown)


felangel:
Rather than making this a breaking change we could add the beta flag and enforce that only one is provided. Also we should align on naming (track vs channel) and be consistent throughout the cli and code push package

bryanoltman:
> Rather than making this a breaking change we could add the beta flag and enforce that only one is provided. Also we should align on naming (track vs channel) and be consistent throughout the cli and code push package

We could, but we'd effectively be using bool flags for what is actually an enum, which gets nasty if/when we add more and we'd have to decide which takes precedence when more than one are passed.

felangel:
> > Rather than making this a breaking change we could add the beta flag and enforce that only one is provided. Also we should align on naming (track vs channel) and be consistent throughout the cli and code push package
> 
> We could, but we'd effectively be using bool flags for what is actually an enum, which gets nasty if/when we add more and we'd have to decide which takes precedence when more than one are passed.

Yeah that’s fair I’m fine with the change just want to make sure we are consistent in the naming (channel vs track). I personally think of the pre-defined channels as tracks but don’t have a strong opinion as long as we’re consistent🤷

bryanoltman:
sg, renamed "channel" to "track"

felangel (approved):

PR #2644, opened 2024-11-26
feat: add support for releasing and patching macOS apps

### LINKED ISSUE #2654: feat: make macOS releases previewable
**Description**

It's not immediately clear what the best way to do this is, but macOS prevents running downloaded apps without a really annoying visit to the System Preferences security pane. To see

1. Download the app.zip from [here](https://console.cloud.google.com/storage/browser/shorebird_release_artifacts/b98ac807-cd22-40e5-87fc-bd22e5cb113c/macos/app/136688?project=code-push-prod&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22)))
2. Unzip
3. Attempt to open
4. See this
Screenshot 2024-12-04 at 5 16 17 PM
5. Open System Settings.app and open the Privacy & Security pane
6. Scroll to the bottom
7. Select "open anyway"

This is way more cumbersome than it should be. The proper solution might be to sign macOS apps before uploading them.

bryanoltman:
## Description

Adds `shorebird release macos` and `shorebird patch macos`

## Remaining work

Note: These should probably be moved to issues, as they aren't prereqs for landing this PR, but for announcing beta mac support.

- [x] Add beta warning to mac release and patch
- [x] Add validator for network access in mac app entitlements
- [ ] Preview needs to stream logs
- [x] Preview should (maybe) open the directory where the app is downloaded for relaunching (tbd if this is the right solution, just trying to find a way to make it easy to view patches in a previewed app)
- [x] Patch diffing (native and asset change detection)

Fixes https://github.com/shorebirdtech/shorebird/issues/2654

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,370 @@
...
+      p.join(
+        appDirectory.path,
+        'Contents',
+        'Frameworks',
+        'App.framework',
+        'App',
+      ),
+    );
+
+    final useLinker = AotTools.usesLinker(shorebirdEnv.flutterRevision);
```
felangel: need to unzip and copy the App.ct.link and App.class_table.json from supplementArtifact here just like in ios_patcher.dart

bryanoltman: That did it! Thanks

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -256,6 +257,98 @@ class ArtifactBuilder {
...
+    String? base64PublicKey,
+    DetailProgress? buildProgress,
+  }) async {
+    final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
+    // Delete the .dart_tool directory to ensure that the app is rebuilt.
+    // Without this, the build command will not print the app.dill path
+    final dartToolDir = Directory(p.join(projectRoot.path, '.dart_tool'));
+    if (dartToolDir.existsSync()) {
+      dartToolDir.deleteSync(recursive: true);
+    }
```
eseidel, line 293: Amusing.  This also "needlessly" slows down our build since it causes the source->kernel generation to have to run again.

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -256,6 +257,98 @@ class ArtifactBuilder {
...
+    });
+
+    if (appDillPath == null) {
+      throw ArtifactBuildException('''
+Unable to find app.dill file.
+Please file a bug at https://github.com/shorebirdtech/shorebird/issues/new with the logs for this command.
+''');
+    }
+
+    return IpaBuildResult(kernelFile: File(appDillPath!));
```
felangel: Shouldn't we make a separate result type rather than sharing the IpaBuildResult for iOS?

bryanoltman: Yeah, I think we'll probably want to unify a bunch of the linking behavior that exists in all of the apple targets. Not sure what we call it.

packages/shorebird_cli/lib/src/artifact_builder.dart
bryanoltman: This could definitely be improved. A couple alternatives:

1. We could be more surgical in what we remove. I tried just deleting the app.dill file and the build failed with an error (don't have it handy)
2. We could return to our hackier way of detecting app.dill files (i.e., combing through the .dart_tool directory for the newest app.dill)

eseidel: I suspect the real fix is we'll just add this to the pile of things we fix when we move more of `shorebird` down into `flutter`

packages/shorebird_cli/lib/src/artifact_builder.dart
bryanoltman: Called it MacosBuildResult. We'll fix it later.

packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -264,6 +264,38 @@ class ArtifactManager {
...
+        // Get the most recently modified app to handle cases where an app
+        // may produce multiple apps with different names.
+        // This still could grab the wrong app, if multiple `flutter` commands
+        // are running in parallel or the clock is/was broken on the machine.
+        // If either of those occurs in the wild we can check the contents of
+        // the apps, but this should be good enough for now.
+        .sorted(
+          (a, b) => b.statSync().modified.compareTo(a.statSync().modified),
+        )
+        .firstWhereOrNull((directory) => directory.path.endsWith('.app'));
```
eseidel, line 296: I thought in the iOS case we'd move to reading the path from the logs?  But maybe that was a different kind of path?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -616,6 +617,66 @@ aar artifact already exists, continuing...''',
...
+    await ditto.archive(source: appPath, destination: zippedApp.path);
+
+    try {
+      await codePushClient.createReleaseArtifact(
+        appId: appId,
+        releaseId: releaseId,
+        artifactPath: zippedApp.path,
+        arch: 'app',
+        platform: ReleasePlatform.macos,
+        hash: sha256.convert(await zippedApp.readAsBytes()).toString(),
```
eseidel, line 641: I probably would have had a hash function somewhere rather than encoding the use of sha256 in multiple places. 🤷

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,446 @@
...
+  /// {@macro macos_patcher}
+  MacosPatcher({
+    required super.argParser,
+    required super.argResults,
+    required super.flavor,
+    required super.target,
+  });
+
+  String get _patchClassTableLinkInfoPath =>
+      p.join(buildDirectory.path, 'macos', 'shorebird', 'App.ct.link');
```
eseidel, line 48: I probably would have written myself a little _buildPath('macos', 'shorebird','App.ct.link') helper. 🤷

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,446 @@
...
+        ? [
+            '--dwarf-stack-traces',
+            '--resolve-dwarf-paths',
+            '''--save-debugging-info=${saveDebuggingInfoPath(splitDebugInfoPath)}''',
+          ]
+        : <String>[];
+  }
+
+  /// The path to save the split debug info file.
+  static String saveDebuggingInfoPath(String directory) {
```
eseidel: Leading with a verb makes me think this method has side-effects, rather than being a getter.  🤷

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,446 @@
...
+    File? releaseClassTableLinkInfoFile;
+    File? releaseClassTableLinkDebugInfoFile;
+    if (supplementArtifact != null) {
+      final tempDir = Directory.systemTemp.createTempSync();
+      await artifactManager.extractZip(
+        zipFile: supplementArtifact,
+        outputDirectory: tempDir,
+      );
+      releaseClassTableLinkInfoFile = File(p.join(tempDir.path, 'App.ct.link'));
+      if (!releaseClassTableLinkInfoFile.existsSync()) {
```
eseidel: I probably would have had an _existsOrExit(file, 'class table link debug info') if I'm doing this more often. 🤷

packages/shorebird_cli/lib/src/commands/release/macos_releaser.dart
```diff
@@ -0,0 +1,212 @@
...
+
+  @override
+  ReleaseType get releaseType => ReleaseType.macos;
+
+  @override
+  Future<void> assertArgsAreValid() async {
+    if (argResults.wasParsed('release-version')) {
+      logger.err(
+        '''
+The "--release-version" flag is only supported for aar and ios-framework releases.
```
eseidel, line 49: macos-framework?

packages/shorebird_cli/lib/src/shorebird_artifacts.dart
```diff
@@ -10,14 +10,20 @@ import 'package:shorebird_cli/src/shorebird_env.dart';
 
 /// All Shorebird artifacts used explicitly by Shorebird.
 enum ShorebirdArtifact {
-  /// The analyze_snapshot executable.
-  analyzeSnapshot,
+  /// The iOS analyze_snapshot executable.
+  analyzeSnapshotIos,
+
+  /// The macOS analyze_snapshot executable.
+  analyzeSnapshotMacOS,
```
eseidel, line 17: I guess android doesn't use analyze_snapshot or gen_snapshot thus this is the first time these have had a paltform component?

packages/shorebird_cli/lib/src/shorebird_artifacts.dart
```diff
@@ -71,6 +81,20 @@ class ShorebirdCachedArtifacts implements ShorebirdArtifacts {
...
 
+  File get _analyzeSnapshotMacosFile {
+    return File(
+      p.join(
+        shorebirdEnv.flutterDirectory.path,
+        'bin',
+        'cache',
+        'artifacts',
+        'engine',
+        'darwin-x64-release',
```
eseidel, line 92: We're using the enum to differentiate based on Target, but we may also need to differentiate based on Host

eseidel (commented):
I left a bunch of nits, none of which need to be addressed.

I do think there is a ton of copy/paste code that we may wish to try and unify more before soon.

packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -377,6 +409,26 @@ class ArtifactManager {
...
+  ///
+  /// Returns null if there is no supplement directory
+  /// (e.g. when using older Flutter revisions).
+  Directory? getMacosReleaseSupplementDirectory() {
+    final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
+    final releaseSupplementDir = Directory(
+      p.join(projectRoot.path, 'build', 'macos', 'shorebird'),
+    );
+
+    if (!releaseSupplementDir.existsSync()) {
```
felangel, line 422: This should technically never happen so we might want to have a more fatal error here since MacOS was only introduced after the supplement artifact was introduced.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -616,6 +617,66 @@ aar artifact already exists, continuing...''',
...
+      );
+    } catch (error) {
+      _handleErrorAndExit(
+        error,
+        progress: createArtifactProgress,
+        message: 'Error uploading app: $error',
+      );
+    }
+
+    if (supplementPath != null) {
```
felangel: This should never be null so we should probably make this a non-nullable param.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -652,6 +714,7 @@ aar artifact already exists, continuing...''',
 
     final zippedRunner = await Directory(runnerPath).zipToTempFile();
     try {
+      logger.detail('[archive] zipped runner.app to ${zippedRunner.path}');
```
felangel, line 715: Do we want this log? We don't have an equivalent for other platforms so just wondering if this was left in intentionally.

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
felangel: agreed can we rename to `splitDebuggingInfoPath`?

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,446 @@
...
+    final unzipProgress = logger.progress('Extracting release artifact');
+    final releaseAppDirectory = Directory.systemTemp.createTempSync();
+    await ditto.extract(
+      source: releaseArtifact.path,
+      destination: releaseAppDirectory.path,
+    );
+
+    File? releaseClassTableLinkInfoFile;
+    File? releaseClassTableLinkDebugInfoFile;
+    if (supplementArtifact != null) {
```
felangel: supplementArtifact should never be null for MacOS so we should refactor this to fail fast if it ever is null.

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
felangel: I'm planning to refactor all this and extract the shared logic in a subsequent PR.

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,446 @@
...
+        'App',
+      ),
+    );
+
+    final useLinker = AotTools.usesLinker(shorebirdEnv.flutterRevision);
+    if (useLinker) {
+      // If we're using a newer version of the linker, we need to also copy the
+      // necessary class table link information alongside the snapshots.
+      if (releaseClassTableLinkInfoFile != null &&
+          releaseClassTableLinkDebugInfoFile != null) {
```
felangel: These should never be null so we should refactor this to exit early if either one is null on MacOS

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -0,0 +1,446 @@
...
+    CreatePatchMetadata metadata,
+  ) async =>
+      metadata.copyWith(
+        linkPercentage: lastBuildLinkPercentage,
+        environment: metadata.environment.copyWith(
+          xcodeVersion: await xcodeBuild.version(),
+        ),
+      );
+
+  Future<_LinkResult> _runLinker({
```
felangel, line 367: We should probably create a `Linker` class that contains this shared logic instead of duplicating on both iOS and MacOS.

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -280,6 +285,65 @@ class PreviewCommand extends ShorebirdCommand {
...
+          destination: appDirectory.path,
+        );
+        downloadArtifactProgress.complete();
+      } catch (error) {
+        downloadArtifactProgress.fail('$error');
+        return ExitCode.software.code;
+      }
+    }
+
+    // TODO(felangel): wrap `open` and stream logs.
```
felangel, line 342: Will work on this right after this PR is merged.

packages/shorebird_cli/lib/src/commands/release/macos_releaser.dart
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    // Delete the Shorebird supplement directory if it exists.
+    // This is to ensure that we don't accidentally upload stale artifacts
+    // when building with older versions of Flutter.
+    final shorebirdSupplementDir =
+        artifactManager.getMacosReleaseSupplementDirectory();
+    if (shorebirdSupplementDir?.existsSync() ?? false) {
+      shorebirdSupplementDir!.deleteSync(recursive: true);
+    }
```
felangel: This shouldn't be needed on MacOS because it was introduced after the supplement directory changes so there will always be a supplement artifact generated by the flutter_tool.

felangel (approved):

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
felangel: Will clean that up in a future PR

packages/shorebird_cli/lib/src/commands/release/macos_releaser.dart
bryanoltman: Will add when we support macos-framework. As written it's correct, as these are the two release platforms that support this argument.

packages/shorebird_cli/lib/src/shorebird_artifacts.dart
bryanoltman: Yep!

packages/shorebird_cli/lib/src/shorebird_artifacts.dart
bryanoltman: Yep, we'll probably want to refactor `ShorebirdArtifacts` to accept host and/or target in the future.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: I don't feel strongly. Users will only see it when running with the `-v` flag, and there's so much other output that I'm not sure it matters much one way or the other.

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
bryanoltman: Yep, I have a TODO earlier in this file on line 33 for the `_LinkResult` typedef.

packages/shorebird_cli/lib/src/artifact_manager.dart
bryanoltman: You might be thinking of the app.dill?

PR #2646, opened 2024-11-27
fix(shorebird_cli): "SYS_KEYS has no physical keys" during preview

elliots:
"** SYS_KEYS has no physical keys but with factor 2.0%."

## Description

When running preview, I got 

```
⠹ Starting app... (0.2s)Exited with code 251

stdout:
  bash arg: -p
  bash arg: com.my.app
  bash arg: 1


stderr:
args: [-p, com.my.app, 1]
 arg: "-p"
 arg: "com.my.app"
 arg: "1"
data="com.my.app"
** SYS_KEYS has no physical keys but with factor 2.0%.
```

Found https://stackoverflow.com/questions/44860475/how-to-use-the-monkey-command-with-an-android-system-that-doesnt-have-physical

Now it doesn't error.

But, I have no idea why I got it, or if this fix will cause other problems.

```
❯ adb --version
Android Debug Bridge version 1.0.41
Version 35.0.2-12147458
Installed as /Users/me/Library/Android/sdk/platform-tools/adb
Running on Darwin 23.5.0 (arm64)
```

## Type of Change

- [X] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/executables/adb.dart
```diff
@@ -51,6 +51,7 @@ class Adb {
       if (deviceId != null) ...['-s', deviceId],
       'shell',
       'monkey',
```
eseidel, line 53: I'm so confused why we're using monkey at all.

eseidel: Huh, I guess that's a known way.  I'm used to historically sending an intent: https://www.repeato.app/launching-android-applications-via-adb/

bryanoltman:
It looks like tests are failing and cspell is unhappy. These should both be easy to fix. For cspell, you can add "syskeys" to `shorebird/cspell.config.yaml`

felangel (approved):
LGTM thanks so much for the contribution!

felangel:
For context the reason you saw this error is most likely because you were trying to preview on an Android system which does not have hardware/physical system keys.

PR #2655, opened 2024-12-04
feat(shorebird_cli): iOS linking improvment (improved class table sort)

felangel:
## Description

- feat: iOS linking improvements (improved class table sort)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel:
What does updating this to support Mac look like (which Bryan is about to attempt tomorrow I suspect).

Should this be just class_table.zip instead of ios_supplement.zip or supplement.zip rather than ios_* like you originally had?  I don't have strong opinions on how we solve it, just we'll have a mac question immediately.

felangel:
> What does updating this to support Mac look like (which Bryan is about to attempt tomorrow I suspect).
> 
> Should this be just class_table.zip instead of ios_supplement.zip or supplement.zip rather than ios_* like you originally had? I don't have strong opinions on how we solve it, just we'll have a mac question immediately.

We'd add a `mac_supplement.zip`. The reason for the distinct names is because on the backend the release_id, platform, and arch combo must be unique. In the case of normal iOS and add-to-app iOS we'd have collisions if we used the same name.

I'd rather not name it class_table.zip since we'll eventually likely have other info and I don't think we want to upload and download N artifacts for the different linker phases (e.g. object pool, inliner, etc.). This way we could include them all in a single zip with a single download/upload.

eseidel (approved):

PR #2677, opened 2024-12-12
fix: remove __MACOSX directories from ditto archives

bryanoltman:
## Description

`ditto` archives include a lot of `__MACOSX` directories, which we don't need and confuse our archive differ. This change updates `ditto.archive` to remove those directories after zipping (there doesn't seem to be a flag we can pass to ditto that will keep them from being zipped in the first place).

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/executables/ditto.dart
+++ b/packages/shorebird_cli/lib/src/executables/ditto.dart
@@ -44,9 +44,20 @@ class Ditto {
       source,
       destination,
     ];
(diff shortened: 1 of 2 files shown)


bryanoltman:
Closing because this can be fixed by removing the `--sequesterRsrc` flag to the ditto call

PR #2681, opened 2024-12-13
chore: add windows to ReleasePlatform enum

bryanoltman:


packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -244,6 +244,7 @@ class PreviewCommand extends ShorebirdCommand {
           deviceId: deviceId,
           track: track,
         ),
+      ReleasePlatform.windows => throw UnimplementedError(),
```
felangel, line 247: can we add a unit test?

felangel (approved):

packages/shorebird_cli/lib/src/commands/preview_command.dart
bryanoltman: Done

PR #2710, opened 2024-12-18
fix(shorebird_cli): print shorebird checkout output on windows, verify longpaths

### LINKED ISSUE #2709: fix: shorebird flutter checkout can fail silently on Windows
**Description**

Shorebird commands on windows are executed through shorebird\bin\shorebird.ps1. This silently performs a checkout of our Flutter fork if needed, which can fail if the paths are too long. We currently redirect this output to null, so the failure is silent. Without redirection, errors like the following are printed:

```
warning: unable to access 'dev/integration_tests/ios_app_with_extensions/ios/watch Extension/Assets.xcassets/Complication.complicationset/Graphic Circular.imageset/.gitattributes': Filename too long
```

bryanoltman:
## Description

Updates shorebird.ps1 to print git output when we check out our fork of Flutter and to verify that git's core.longpaths config property is true. 

Fixes https://github.com/shorebirdtech/shorebird/issues/2709

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/bin/shorebird.ps1
+++ b/bin/shorebird.ps1
@@ -23,18 +23,25 @@ $dart = [IO.Path]::Combine($flutterPath, "bin", "cache", "dart-sdk", "bin", "dar
 # This is fixed in Powershell 7.1.
 # 
 # See https://github.com/PowerShell/PowerShell/issues/4002 for more info.
-function Invoke-SilentlyIfPossible($command) {
+function Invoke-SilentlyIfNeeded($command) {
     $psVersion = $PSVersionTable.PSVersion
-    $shouldRedirectStdErr = $psVersion.Major -ge 7 -and $psVersion.Minor -ge 1
-    if ($shouldRedirectStdErr) {
-        # Redirect everything to $null.
-        & $command *> $null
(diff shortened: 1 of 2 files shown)


eseidel:
kahn

bin/shorebird.ps1
```diff
@@ -38,6 +38,14 @@ function Invoke-SilentlyIfPossible($command) {
     }
 }
 
+function Test-GitConfigLongpaths {
+    $longpathsEnabled = git config --system core.longpaths
+    if ($longpathsEnabled -ne "true") {
```
eseidel, line 43: We have 1000s of customers on windows today.  Do most of them have this on? Should this just be a warning rather than fatal?

eseidel (approved):

bin/shorebird.ps1
bryanoltman: Updated to just print the warning and not exit

PR #2712, opened 2024-12-18
feat: support releasing and patching Windows apps

bryanoltman:
## Description

Adds ability to release, patch , and preview windows apps.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -316,6 +321,72 @@ This is only applicable when previewing Android releases.''',
...
+      getArtifactPath(
+        appId: appId,
+        release: release,
+        artifact: releaseExeArtifact,
+        platform: platform,
+        extension: 'exe',
+      ),
+    );
+
+    if (!appDirectory.existsSync()) {
```
eseidel, line 359: It's a bit odd that this "cache if needed" behavior is built into the command here, rather than some central caching class, but that can be a later refactor when needed. 🤷

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
```diff
@@ -0,0 +1,160 @@
...
+  /// {@macro windows_patcher}
+  WindowsPatcher({
+    required super.argParser,
+    required super.argResults,
+    required super.flavor,
+    required super.target,
+  });
+
+  @override
+  // TODO(bryanoltman): exe isn't technically correct - we upload a zip
```
eseidel: Will this be a pain to change later?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -520,6 +524,44 @@ aab artifact already exists, continuing...''',
...
+  }) async {
+    final createArtifactProgress = logger.progress('Uploading artifacts');
+
+    try {
+      // logger.detail('Uploading artifact for $aabPath');
+      await codePushClient.createReleaseArtifact(
+        appId: appId,
+        releaseId: releaseId,
+        artifactPath: releaseZipPath,
+        arch: 'exe',
```
eseidel: Should this be a shared constant somewhere?

packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -347,6 +347,21 @@ class ArtifactManager {
...
 
+  /// Returns the build/ subdirectory containing the compiled Windows exe.
+  Directory getWindowsReleaseDirectory() {
+    final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
+    return Directory(
+      p.join(
+        projectRoot.path,
+        'build',
+        'windows',
+        'x64',
```
eseidel, line 358: Eventually this method may need to support arch, since I think windows does support arm64. 🤷

packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -547,6 +548,53 @@ Either run `flutter pub get` manually, or follow the steps in ${cannotRunInVSCod
...
+        '--release',
+        ...args,
+      ];
+
+      final buildProcess = await process.start(
+        executable,
+        arguments,
+        runInShell: true,
+        // TODO(bryanoltman): support this
+        // environment: base64PublicKey?.toPublicKeyEnv(),
```
eseidel, line 573: Do we need a list of remaining windows bugs somewhere?

eseidel (approved):
overall lgtm.  Looks like a great start.

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
bryanoltman: We would probably want to change it before launching

packages/shorebird_cli/lib/src/artifact_manager.dart
bryanoltman: Yep. When we add support for arm64 :)

packages/shorebird_cli/lib/src/artifact_builder.dart
bryanoltman: Not a bad idea, started one at https://github.com/shorebirdtech/shorebird/issues/2745

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: Done

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
bryanoltman: Changed to `win_archive`

packages/shorebird_cli/lib/src/commands/preview_command.dart
felangel: This is specifically caching the artifacts for preview so it makes sense for it to be part of the preview command imo 🤷 

It's different from other cached artifacts in that we want to download them lazily unlike bundletool or aot_tools

felangel (approved):

PR #2746, opened 2025-01-09
fix: ensure windows releases include build number

bryanoltman:
## Description

If a flutter app's pubspec.yaml contains a version without a build number (e.g., `1.0.0`), our engine logic will parse that as `1.0.0+0`. This change updates the CLI's version parsing logic to append `+0` if no build number is present.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/executables/powershell.dart
+++ b/packages/shorebird_cli/lib/src/executables/powershell.dart
@@ -47,6 +47,10 @@ class Powershell {
       runInShell: true,
     );
 
-    return (result.stdout as String).trim();
+    var versionString = (result.stdout as String).trim();
+    if (!versionString.contains('+')) {
+      versionString += '+0';
+    }
+    return versionString;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/executables/powershell.dart
```diff
@@ -47,6 +47,10 @@ class Powershell {
       runInShell: true,
     );
 
-    return (result.stdout as String).trim();
+    var versionString = (result.stdout as String).trim();
+    if (!versionString.contains('+')) {
+      versionString += '+0';
```
eseidel, line 52: I wonder if this belongs here or somewhere deeper (to cover other platforms with the same bug)?

eseidel (approved):

packages/shorebird_cli/lib/src/executables/powershell.dart
bryanoltman: As of now, Windows is the only platform with this issue, primarily because we determine the release version differently in the CLI than we do from the executable. Leaving here for now.

PR #2772, opened 2025-01-15
fix: remove quotes from ditto files

### LINKED ISSUE #2764: Exception while releasing macos. console log ->  Failed to archive: ditto: Can't archive multiple sources
I keep getting exception when I try to release macos. It's the first time, I'm trying to integrate shorebird.

Image

Image

[1736932253614_shorebird.log](https://github.com/user-attachments/files/18421734/1736932253614_shorebird.log)

bryanoltman:
## Description

These were making `shorebird preview` unhappy – I think simply moving from a string split by spaces to an arg array (so we aren't splitting `/my/path/with spaces/somedir` into two separate args) is sufficient to resolve https://github.com/shorebirdtech/shorebird/issues/2764

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/executables/ditto.dart
+++ b/packages/shorebird_cli/lib/src/executables/ditto.dart
@@ -21,7 +21,7 @@ class Ditto {
     required String source,
     required String destination,
   }) async {
-    final result = await _exec(['-x', '-k', '"$source"', '"$destination"']);
+    final result = await _exec(['-x', '-k', source, destination]);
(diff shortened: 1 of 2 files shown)


eseidel (approved):
Yeah, the quotes are only needed for the shell, when passing directly via the args array they just end up as though they were escaped \" characters in the path.

PR #2782, opened 2025-01-17
feat(shorebird_cli): `shorebird patch <platform> --release-version=latest`

### LINKED ISSUE #2779: feat: add `latestRelease` option to shorebird patch
**Description**

depending on how your organisation triggers the patching process you might not have the release number of the latest shorebird release accessible. For instance if you trigger a patch based on a commit on a `patch` branch. 

So having a `shorebird patch android --onLatestRelease` flag would be really nice to have that

felangel:
## Description

- Closes #2779 

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -3,6 +3,7 @@ import 'dart:io';
 import 'package:collection/collection.dart';
 import 'package:mason_logger/mason_logger.dart';
 import 'package:meta/meta.dart';
+import 'package:pub_semver/pub_semver.dart';
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/artifact_manager.dart';
 import 'package:shorebird_cli/src/cache.dart';
@@ -77,8 +78,8 @@ class PatchCommand extends ShorebirdCommand {
       ..addOption(
         'release-version',
         help: '''
-The version of the associated release (e.g. "1.0.0"). This should be the version
-of the iOS app that is using this module.''',
+The version of the associated release (e.g. "1.0.0").
+To target the latest release (e.g. highest build number) use "latest".''',
       )
       ..addFlag(
         'allow-native-diffs',
@@ -186,6 +187,10 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
   /// Whether the patch is for the staging environment.
   bool get isStaging => track == DeploymentTrack.staging;
 
+  /// Whether the patch is targeting the latest release version
+  /// (--release-version=latest).
+  bool get useLatestRelease => results['release-version'] == 'latest';
+
   /// The deployment track to publish the patch to.
   DeploymentTrack get track {
     final channel = results['track'] as String;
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -77,8 +78,8 @@ class PatchCommand extends ShorebirdCommand {
       ..addOption(
         'release-version',
         help: '''
-The version of the associated release (e.g. "1.0.0"). This should be the version
-of the iOS app that is using this module.''',
```
bryanoltman, line 81: I don't think we want to completely remove this text, although how it was is confusing. Maybe:

```
The version of the associated release (e.g. "1.0.0").
To target the latest release (highest released version) use --release-version=latest.

If you are building an xcframework or aar, this number needs to match the host app's release version.
```

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -285,14 +290,33 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
...
+      final releases = await codePushClientWrapper.getReleases(appId: appId);
+      releases
+        ..removeWhere(
+          (release) => !release.platformStatuses.keys.contains(releasePlatform),
+        )
+        ..sort(
+          (a, b) => Version.parse(b.version).compareTo(
+            Version.parse(a.version),
+          ),
+        );
```
bryanoltman: I think this might be worth pulling out into an extension and unit testing separately. Looking at the pub_semver package, it's not clear how this would handle build numbers or versions that aren't technically semver (e.g., `1.0`)

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
felangel: thanks for the context, done 👍

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
felangel: Yup good call, updated to include dedicated unit tests for sort order. Regarding versions that aren't technically semver (e.g. 1.0) how would that be possible? We can have missing build numbers or weirdness like `1.0.0+1.0.0` but those are all still technically valid versions whereas `1.0` is not.

packages/shorebird_cli/test/src/commands/patch/patch_command_test.dart
```diff
@@ -1211,4 +1315,54 @@ Please re-run the release command for this version or create a new release.''',
...
+  group('sortByVersion', () {
+    test('sorts versions by semver', () {
+      // Sorts by major version
+      expect(
+        [
+          _FakeRelease(version: '1.0.0+1'),
+          _FakeRelease(version: '3.0.0+1'),
+          _FakeRelease(version: '2.0.0+1'),
+          _FakeRelease(version: '4.0.0+1'),
+        ]..sortByVersion(),
```
bryanoltman: ultra-nit: I usually expect sort to be ascending by default, so potentially

bryanoltman (approved):
One more small comment, but otherwise LGTM!

PR #2790, opened 2025-01-21
fix: detect macOS flavors in shorebird init

bryanoltman:
## Description

Updates `shorebird init` to detect flavors (i.e., schemes) in the macOS Xcode project. Because this logic is almost identical to its iOS counterpart, and because the iOS counterpart _only_ deals with flavor detection, this change also consolidates `platform/ios.dart` and `platform/macos.dart` into `platform/apple.dart` to share code for both platforms.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel (approved):
I might have consider hiding the shared-ness, behind separate ios and macos variables (both of which are actually Apples under the covers?), but this works too. Will just have some work to refactor when we end up splitting them again, unless some code will want to use *both* apple explicitly and ios/macos in parts then.  We'll cross that bridge when we come to it.  Thanks!

Issue #2711, opened 2024-12-18
feat: mac desktop support


eseidel:
Creating a bug to track the remaining work for Mac Desktop support.

- [x] Flavor Detection
- [x] Notarization
- [x] Intel Mac support
- [ ] Mac Add-to-App?

I'm sure there are more (feel free to edit), and I'm sure we don't need to solve all of those to resolve this.  Just starting a list so we have it.

bryanoltman:
I think this is done.

PR #2802, opened 2025-01-23
feat: add support for macos on intel

bryanoltman:
## Description

Switches macOS patch creation to use the simpler Android and Windows method of simply diffing the AOT snapshot against the release.

We still need to add validators to ensure that the correct app hardening settings are present in the macOS xcode project (specifically, if app hardening is enabled, that "allow unsigned executable memory" is checked):

Screenshot 2025-01-24 at 5 49 38 PM

I've verified with this that I can patch a mac release on both arm64 and intel machines, although there are some asset changes when using an intel mac to patch a release created on arm64.

Part of https://github.com/shorebirdtech/shorebird/issues/2711

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -338,7 +338,10 @@ class ArtifactBuilder {
       final stdout = stdoutLines.join('\n');
       final exitCode = await buildProcess.exitCode;
       if (exitCode != ExitCode.success.code) {
-        throw ArtifactBuildException('Failed to build: $stderr');
+        throw ArtifactBuildException('''
+Failed to build
+stdout: $stdout
```
bryanoltman, line 343: stdout occasionally contains useful failure messages that are not included in stderr

PR #2570, opened 2024-10-23
fix(shorebird_cli): support `--split-debug-info`

felangel:
## Description

- fix(shorebird_cli): support `--split-debug-info` (closes #2480)
  - when passing the `--split-debug-info` flag to the `flutter_tool`, the resulting `gen_snapshot` execution includes additional flags ('--dwarf-stack-traces', '--resolve-dwarf-paths', '--save-debugging-info') previously, we were proxying this flag when building the release but when building the optimized aot snapshot (as part of the linking step) we were not supplying the additional flags which would result in different vm sections of the generated snapshot.

## Type of Change

- [ ] ✨ New feature (non-breaking change which adds functionality)
- [X] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

PR #2807, opened 2025-01-27
feat: remove `--split-debug-symbols` handling from macOS patching

bryanoltman:
## Description

The `--split-debug-symbols` flag requires special handling on iOS due to linking considerations (see https://github.com/shorebirdtech/shorebird/pull/2570). macOS no longer has these considerations, so we can remove this code.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
@@ -2,7 +2,6 @@ import 'dart:io';
 
 import 'package:crypto/crypto.dart';
 import 'package:mason_logger/mason_logger.dart';
-import 'package:meta/meta.dart';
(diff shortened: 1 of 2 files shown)


eseidel (approved):
Remind me why we need special handling for split-debug-symbols?

bryanoltman:
We need it when building the AOT snapshot: https://github.com/shorebirdtech/shorebird/pull/2570. I'm not sure the reasoning there actually applies here, since we don't really care about the ordering of the snapshot on macOS. Maybe we can just remove this bit?

eseidel (approved):
I see, this is only needed when we do the multi-pass compile.  So we'll end up adding this back (everywhere) when we do support that.  Thanks!

PR #2808, opened 2025-01-27
fix: tighten Flutter revisions for macOS releasing and patching

bryanoltman:
## Description

Because we changed the macOS patching method in https://github.com/shorebirdtech/engine/pull/96, older macOS releases are no longer patchable. This change increases the minimum supported Flutter version to 3.27.3 (the latest stable version) and allowlists revisions of that version for patching to filter out previous 3.27.3 revisions that do not support the new patching mechanism. This allowlist should be removed when we can increase the minimum supported macOS Flutter version to 3.27.4 or higher.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -68,6 +68,7 @@ words:
   - nserror
   - Oltman
   - orri # Arm64 instruction, Or Register with Immediate
+  - patchability
   - pana
   - pkcs
   - podfile
@@ -107,6 +108,7 @@ words:
   - unmockable
   - Unpatchable
   - unsets
+  - unskipped
   - upvote
   - usbmuxd
   - USERPROFILE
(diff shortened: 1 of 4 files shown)


eseidel:
I guess this is OK since it's beta.  But is a bit awkward that any current user of `shorebird` + mac will no longer be able to patch.  That said, there aren't very many yet, so this seems reasonable.

eseidel (approved):

bryanoltman:
> I guess this is OK since it's beta. But is a bit awkward that any current user of shorebird + mac will no longer be able to patch. That said, there aren't very many yet, so this seems reasonable.

Yeah, it's unfortunate, but that's why we're in beta :)

PR #2811, opened 2025-01-28
feat: add support for releasing, previewing, and patching linux apps

bryanoltman:
## Description

Adds the `shorebird release linux` and `shorebird patch linux` as well as `shorebird preview` support for Linux.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/artifact_builder.dart
```diff
@@ -274,6 +274,45 @@ class ArtifactBuilder {
...
+        '--release',
+        if (flavor != null) '--flavor=$flavor',
+        if (target != null) '--target=$target',
+        ...args,
+      ];
+
+      final buildProcess = await process.start(
+        executable,
+        arguments,
+        runInShell: true,
```
eseidel: I think you'll want to avoid this whenever you can. :)

eseidel: What this does is pass the arguments array to `sh` rather than to the executable directly, which means they get processed (e.g. variable expansion, quote simplification, etc.) which is sometimes what you want?  But rarely.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -656,6 +656,35 @@ aar artifact already exists, continuing...''',
...
+    try {
+      await codePushClient.createReleaseArtifact(
+        appId: appId,
+        releaseId: releaseId,
+        artifactPath: zippedBundle.path,
+        arch: 'bundle',
+        platform: ReleasePlatform.linux,
+        hash: sha256.convert(await zippedBundle.readAsBytes()).toString(),
+        canSideload: true,
+        podfileLockHash: null,
```
eseidel, line 676: Both of these are meaningless in this context, I wonder if they end up with defaults 🤷

packages/shorebird_cli/lib/src/commands/patch/linux_patcher.dart
```diff
@@ -0,0 +1,167 @@
...
+    final flutterVersionString = await shorebirdFlutter.getVersionAndRevision();
+
+    final buildAppBundleProgress = logger.detailProgress(
+      'Building Linux app with Flutter $flutterVersionString',
+    );
+
+    final releaseDir = Directory(
+      p.join(
+        shorebirdEnv.getFlutterProjectRoot()!.path,
+        'build',
```
eseidel: I wonder if we end up with some sort of projectPath('foo', 'bar', 'baz') type helpers for these.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bryanoltman: `canSideload` is needed for previewing-we use this to answer the question of whether the artifact can be run ~on its own (obviously we often need to unzip, so "on its own" is a bit murky)

packages/shorebird_cli/lib/src/archive_analysis/linux_bundle_differ.dart
```diff
@@ -8,8 +8,13 @@ class LinuxBundleDiffer extends ArchiveDiffer {
   /// {@macro linux_bundle_differ}
   const LinuxBundleDiffer();
 
+  bool _isDirectoryPath(String path) {
```
eseidel, line 11: package:path probably gives you some of this.  Also just using Directory(path).existsSync() should tell you if you're dealing with a path on disk (I think that checks the type? but if not I'm sure there is a way to check the type)

packages/shorebird_cli/lib/src/archive_analysis/linux_bundle_differ.dart
```diff
@@ -8,8 +8,13 @@ class LinuxBundleDiffer extends ArchiveDiffer {
...
+    return path.endsWith('/');
+  }
+
   @override
   bool isAssetFilePath(String filePath) =>
+      !_isDirectoryPath(filePath) &&
```
eseidel, line 17: I guess this is for zip inspection? (e.g. you're not looking at real paths on disk.

bryanoltman: Yes, correct

packages/shorebird_cli/lib/src/archive_analysis/linux_bundle_differ.dart
bryanoltman: Maybe? We're getting paths as strings from zip files, so I'm not sure `io` would help us much here.

Issue #2856, opened 2025-02-06
SHOREBIRD_TOKEN not working in CodeMagic


eseidel:
> In my case i am getting 
> You must be logged in to run this command.
> If you already have an account, run shorebird login to sign in.
> 
> <img width="418" alt="Image" src="https://github.com/user-attachments/assets/25d28de8-87f3-442b-a293-13abcc03833b" />
> 
> even after adding token 

 _Originally posted by @Gaurav-CareMonitor in [#2855](https://github.com/shorebirdtech/shorebird/issues/2855#issuecomment-2638937238)_

eseidel:
That would suggest that SHOREBIRD_TOKEN is not set.  Can you tell us more about the setup involved?  This is on codemagic, correct?

Gaurav-CareMonitor:
Hi @eseidel ,
Thank you for creating the ticket.

Yes this is on codemagic.
I used token from shorebird login:ci.

Tried 2 times thinking that i might have copied wrong token, but had the same issue.

 I pasted the token in codemagic shorebird setup.

Also tried adding --flavor prod as build param but its not getting saved in code magic. Gets removed after refresh

I'll share more details tomorrow.

Gaurav-CareMonitor:
@eseidel 
Here are the logs

[Android Build Log (1).log](https://github.com/user-attachments/files/18700814/Android.Build.Log.1.log)

Moreover, i have tried adding the build args here, but it just works with initial build, when opening again it doesn't show prefilled for shorebird.

<img width="524" alt="Image" src="https://github.com/user-attachments/assets/74441241-40f9-4ae6-864d-87dc55c9ab00" />

bryanoltman:
This is the behavior I'd expect to see if SHOREBIRD_TOKEN were not set.

Gaurav-CareMonitor:
@bryanoltman can you share if i should change anything from my end?

sabin26:
Hi, I am facing the same issue. I am using a personal CodeMagic Account.

The docs at [https://docs.shorebird.dev/ci/codemagic/](https://docs.shorebird.dev/ci/codemagic/) only mentions about Teams.

I did add the token but still I am getting _"You must be logged in to run this command."_

eseidel:
We're still interested in helping, just our (small) team has been busy with other pressing concerns.  You can also reach codemagic through: https://github.com/orgs/codemagic-ci-cd/discussions

kamranbekirovyz:
[...]

PR #2885, opened 2025-02-12
feat: log when SHOREBIRD_TOKEN is detected and successfully parsed

bryanoltman:
## Description

Logs when the SHOREBIRD_TOKEN environment variable and, if parsable, successfully parsed.

Part of https://github.com/shorebirdtech/shorebird/issues/2856#issuecomment-2654851597

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -315,6 +315,8 @@ class Auth {
   void _loadCredentials() {
     final envToken = platform.environment[shorebirdTokenEnvVar];
     if (envToken != null) {
+      logger.info('$shorebirdTokenEnvVar detected');
+
       try {
         _token = CiToken.fromBase64(envToken.trim());
       } on FormatException catch (e) {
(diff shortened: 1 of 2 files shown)


eseidel (approved):

eseidel:
Thanks!

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -315,6 +315,8 @@ class Auth {
   void _loadCredentials() {
     final envToken = platform.environment[shorebirdTokenEnvVar];
     if (envToken != null) {
+      logger.info('$shorebirdTokenEnvVar detected');
```
felangel, line 318: shouldn't this be a logger.detail? info is probably too noisy for this type of diagnostic log imo 🤷

bryanoltman: I originally wrote this as a detail but opted for info because it's just two lines and will only ever print on CI. I don't feel strongly.

PR #2968, opened 2025-03-13
feat(shorebird_cli): add `inferredReleaseVersion` to `CreatePatchMetadata`

felangel:
## Description

- feat(shorebird_cli): add `inferredReleaseVersion` to `CreatePatchMetadata`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel (approved):
I might have done the opposite, and recorded the explicit version if set, but this seems fine.

felangel:
> I might have done the opposite, and recorded the explicit version if set, but this seems fine.

We already know the release version for the patch and if all we care about is how many people go through the fallback path it seemed this was the simpler approach. I felt the alternative would be more complex because we'd we have the following options:

1. Explicit Release Version (e.g. `--release-version=1.0.0+1`)
2. Prompt
3. Inferred Release Version

If all we care about is 3, then a boolean is all we really need 🤷 
Otherwise we'd have to name it something like `explicitReleaseVersion` and make it nullable or differentiate between prompt and explicit flag which seemed like just more complexity.

I was planning to revert this change once we collected the data we need and adjusted the behavior of the patch command anyway so I wanted to keep the changes as minimal as possible.

eseidel:
I support your approach, lgtm.

PR #2988, opened 2025-03-18
fix(shorebird_cli): windows executable sanitization

### LINKED ISSUE #2986: fix: failed to extract version name from app bundle
App ID: f990fe25-76a0-43d3-8352-96a1049e041a

**Description**

After running this command: 
```
shorebird release android --target .\lib\main_production.dart --flavor production --artifact=apk
```

Shorebird created both .aab and .apk files. But, it also throw an exception as below:

✗ Exception: Failed to extract version name from app bundle: 'C:\Program' is not recognized as an internal or external command,
operable program or batch file

**Steps To Reproduce**

1. Run command: shorebird release android --target .\lib\main_production.dart --flavor production --artifact=apk

**Expected Behavior**

Shorebird create both .aab and .apk files. This release also need to be seen on the shorebird console.

**Additional Context**

OS: Windows 11
Shorebird 1.6.28 • git@github.com:shorebirdtech/shorebird.git
Flutter 3.29.2 • revision 3f9cefb45389b72ff073ddf305fe0939f822143b
Engine • revision b85e118bef1e1289e845226f72f31a5abb95e8a1

Flutter Doctor:
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.29.1, on Microsoft Windows [Version 10.0.22000.2538], locale en-MY)
[√] Windows Version (11 Pro 64-bit, 21H2, 2009)
[√] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
[√] Chrome - develop for the web
[X] Visual Studio - develop Windows apps
    X Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components
[√] Android Studio (version 2022.2)
[√] VS Code, 64-bit edition (version 1.98.1)
[√] Connected device (4 available)
[√] Network resources

[1742264338999_shorebird.log](https://github.com/user-attachments/files/19304402/1742264338999_shorebird.log)

felangel:
## Description

- fix(shorebird_cli): windows executable sanitization (closes https://github.com/shorebirdtech/shorebird/issues/2986)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_process.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_process.dart
@@ -6,6 +6,7 @@ import 'package:meta/meta.dart';
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/engine_config.dart';
 import 'package:shorebird_cli/src/logging/logging.dart';
+import 'package:shorebird_cli/src/platform.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 
 /// A reference to a [ShorebirdProcess] instance.
@@ -13,6 +14,19 @@ final processRef = create(ShorebirdProcess.new);
 
 /// The [ShorebirdProcess] instance available in the current zone.
 ShorebirdProcess get process => read(processRef);
+
+/// Sanitizes the executable path on Windows.
+/// https://github.com/dart-lang/sdk/issues/37751
+String _sanitizeExecutablePath(String executable) {
+  if (executable.isEmpty) return executable;
+  if (!platform.isWindows) return executable;
+  if (executable.contains(' ') && !executable.contains('"')) {
+    // Use quoted strings to indicate where the file name ends and the arguments begin;
+    // otherwise, the file name is ambiguous.
+    return '"$executable"';
+  }
+  return executable;
+}
 
 /// A wrapper around [Process] that replaces executables to Shorebird-vended
 /// versions.
@@ -34,9 +48,6 @@ class ShorebirdProcess {
     Map<String, String>? environment,
     String? workingDirectory,
   }) async {
-    logger.detail(
-      '''[Process.stream] $executable ${arguments.join(' ')}${workingDirectory == null ? '' : ' (in $workingDirectory)'}''',
-    );
     final process = await start(
       executable,
       arguments,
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/shorebird_process.dart
```diff
@@ -34,9 +48,6 @@ class ShorebirdProcess {
     Map<String, String>? environment,
     String? workingDirectory,
   }) async {
-    logger.detail(
```
felangel, line 37: this was a superfluous log since this method just proxies to the underlying start call which also logs as detail

packages/shorebird_cli/lib/src/shorebird_process.dart
```diff
@@ -14,6 +15,19 @@ final processRef = create(ShorebirdProcess.new);
 /// The [ShorebirdProcess] instance available in the current zone.
 ShorebirdProcess get process => read(processRef);
 
+/// Sanitizes the executable path on Windows.
+/// https://github.com/dart-lang/sdk/issues/37751
+String _sanitizeExecutablePath(String executable) {
+  if (executable.isEmpty) return executable;
+  if (!platform.isWindows) return executable;
+  if (executable.contains(' ') && !executable.contains('"')) {
```
eseidel: What about single quote?

felangel: What do you mean?

eseidel:
Is this all caused by our use of run_in_shell?

felangel:
> Is this all caused by our use of run_in_shell?

I don't think so. I think spaces in the executable path would be problematic on windows regardless. These changes align with how the flutter_tool works

packages/shorebird_cli/lib/src/shorebird_process.dart
eseidel: This appears to be trying to quote strings which have spaces but aren't already quoted.  Do we need to check for single-quote as a quote as well?

eseidel:
> > Is this all caused by our use of run_in_shell?
> 
> I don't think so. I think spaces in the executable path would be problematic on windows regardless. These changes align with how the flutter_tool works

Seems plausible (but also super lame).  The whole point of not using run_in_shell (at least in unix) is so that the executable is passed as a single string that's never split.

https://man7.org/linux/man-pages/man3/exec.3.html

Maybe Windows doesn't have equivalent. 🤷

eseidel (approved):

packages/shorebird_cli/lib/src/shorebird_process.dart
felangel: I don't think so -- flutter_tool uses the same logic to sanitize executable paths on windows 🤷 
Going to leave it as is for now since it's consistent with flutter this way.

PR #3037, opened 2025-04-02
feat(redis_client): add `TS.RANGE`

felangel:
## Description

- feat(redis_client): add `TS.RANGE`

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/redis_client/lib/src/redis_client.dart
```diff
@@ -515,41 +515,6 @@ enum RedisTimeSeriesDuplicatePolicy {
...
-/// {@endtemplate}
-class RedisTimeSeriesClientClock extends RedisTimeSeriesClock {
-  /// {@macro redis_time_series_client_clock}
-  const RedisTimeSeriesClientClock();
-  @override
-  RedisTimeSeriesTimestamp now() {
-    return RedisTimeSeriesTimestamp(DateTime.timestamp());
-  }
-}
-
```
felangel, line 552: This didn't change, I just moved it further down in the file.

packages/redis_client/lib/src/redis_client.dart
```diff
@@ -596,6 +561,181 @@ class RedisTimeSeriesTimestamp {
...
+  /// {@macro redis_time_series_server_clock}
+  const RedisTimeSeriesServerClock();
+  @override
+  RedisTimeSeriesTimestamp now() => const RedisTimeSeriesTimestamp._('*');
+}
+
+/// {@template redis_time_series_client_clock}
+/// A [RedisTimeSeriesClock] that represents time on the client.
+/// {@endtemplate}
+class RedisTimeSeriesClientClock extends RedisTimeSeriesClock {
```
eseidel, line 590: I wonder if all these things need a "redis" prefix.  I suspect it could be removed everywhere and importers who care can just import it as 'redis'.

eseidel (approved):
lgtm.  I think you should (in a later change) consider removing the Redis prefix, and possibly even the TimeSeries prefix and just use nice short names and rely on import prefixes to disambiguate if needed.

felangel:
> lgtm. I think you should (in a later change) consider removing the Redis prefix, and possibly even the TimeSeries prefix and just use nice short names and rely on import prefixes to disambiguate if needed.

Sure that'd be a breaking change which we can make later. Were you thinking we'd go as far as to rename `RedisClient` -> `Client`?

We also have `RedisException` which we can't shorten to just `Exception` 🤷

eseidel:
I'd encourage you to ignore any pub usage (it's a non-goal ATM IMO) and just iterate as feels good.  We can always push breaking changes later.

Re: Client, Exception.  🤷   Maybe those stay RedisClient, RedisException.

package:postgres went through a similar migration for 3.0 (also after my urging) and I think they generally liked the results 🤷

PR #3070, opened 2025-04-22
feat: collect linker metadata if available

eseidel:
Runs `aot_tools link_metadata` if available.  Does not block the build if it is not.

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_framework_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_framework_patcher.dart
@@ -58,6 +58,10 @@ class IosFrameworkPatcher extends Patcher {
   @override
   ReleaseType get releaseType => ReleaseType.iosFramework;
 
+  /// The last build's link metadata.
+  @visibleForTesting
+  Map<String, dynamic>? lastBuildLinkMetadata;
+
   @override
   double? get linkPercentage => lastBuildLinkPercentage;
 
@@ -175,13 +179,22 @@ class IosFrameworkPatcher extends Patcher {
         patchSnapshotDir: shorebirdEnv.buildDirectory,
       );
 
-      await apple.runLinker(
+      final result = await apple.runLinker(
         kernelFile: File(_appDillCopyPath),
         releaseArtifact: releaseArtifactFile,
         splitDebugInfoArgs: IosPatcher.splitDebugInfoArgs(splitDebugInfoPath),
         aotOutputFile: File(_aotOutputPath),
         vmCodeFile: File(_vmcodeOutputPath),
       );
+      final linkPercentage = result.linkPercentage;
+      final exitCode = result.exitCode;
+      if (exitCode != ExitCode.success.code) throw ProcessExit(exitCode);
+      if (linkPercentage != null &&
+          linkPercentage < Patcher.linkPercentageWarningThreshold) {
+        logger.warn(Patcher.lowLinkPercentageWarning(linkPercentage));
+      }
+      lastBuildLinkPercentage = linkPercentage;
+      lastBuildLinkMetadata = result.linkMetadata;
     }
 
     final patchBuildFile =
@@ -249,6 +262,7 @@ class IosFrameworkPatcher extends Patcher {
     CreatePatchMetadata metadata,
   ) async => metadata.copyWith(
     linkPercentage: lastBuildLinkPercentage,
+    linkMetadata: lastBuildLinkMetadata,
     environment: metadata.environment.copyWith(
       xcodeVersion: await xcodeBuild.version(),
     ),
--- a/packages/shorebird_cli/lib/src/platform/apple.dart
+++ b/packages/shorebird_cli/lib/src/platform/apple.dart
@@ -327,7 +327,6 @@ $error''');
       }
     } on Exception catch (error) {
       logger.detail('[aot_tools] Failed to get link metadata: $error');
-      linkMetadata = null;
     }
 
     linkProgress.complete();
--- a/packages/shorebird_cli/lib/src/platform/apple.dart
+++ b/packages/shorebird_cli/lib/src/platform/apple.dart
@@ -26,7 +26,34 @@ enum ApplePlatform {
 
 /// A record containing the exit code and optionally link percentage
 /// returned by `runLinker`.
-typedef LinkResult = ({int exitCode, double? linkPercentage});
+class LinkResult {
+  /// Creates a new [LinkResult] with the given [exitCode] and optional
+  /// [linkPercentage].
+  const LinkResult({
+    required this.exitCode,
+    this.linkPercentage,
+    this.linkMetadata,
+  });
+
+  /// Creates a new [LinkResult] representing failure.
+  LinkResult.failure()
+    : exitCode = ExitCode.software.code,
+      linkPercentage = null,
+      linkMetadata = null;
+
+  /// Creates a new [LinkResult] representing success.
+  LinkResult.success({required this.linkPercentage, this.linkMetadata})
+    : exitCode = ExitCode.success.code;
+
+  /// The exit code of the linker process.
+  final int exitCode;
+
+  /// The percentage of code that was linked in the patch.
+  final double? linkPercentage;
+
+  /// Metadata from the linker, if available.
+  final Map<String, dynamic>? linkMetadata;
+}
 
 /// {@template missing_xcode_project_exception}
 /// Thrown when the Flutter project does not have iOS configured as a platform.
@@ -224,7 +251,7 @@ class Apple {
 
     if (!patch.existsSync()) {
       logger.err('Unable to find patch AOT file at ${patch.path}');
-      return (exitCode: ExitCode.software.code, linkPercentage: null);
+      return LinkResult.failure();
     }
 
     final analyzeSnapshot = File(
@@ -235,7 +262,7 @@ class Apple {
 
     if (!analyzeSnapshot.existsSync()) {
       logger.err('Unable to find analyze_snapshot at ${analyzeSnapshot.path}');
-      return (exitCode: ExitCode.software.code, linkPercentage: null);
+      return LinkResult.failure();
     }
 
     final genSnapshot = shorebirdArtifacts.getArtifactPath(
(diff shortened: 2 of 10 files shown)


packages/shorebird_cli/lib/src/platform/apple.dart
```diff
@@ -26,7 +27,30 @@ enum ApplePlatform {
...
+
+  /// Creates a new [LinkResult] representing success.
+  const LinkResult.success({required this.linkPercentage, this.linkMetadata})
+    : _exitCodeObject = ExitCode.success;
+
+  /// ExitCode.code isn't const, so store the actual object.
+  final ExitCode _exitCodeObject;
+
+  /// The exit code of the linker process.
+  int get exitCode => _exitCodeObject.code;
```
felangel, line 46: why do we need to hold onto the object instead of just the underlying int?

packages/shorebird_cli/lib/src/platform/apple.dart
```diff
@@ -26,7 +27,30 @@ enum ApplePlatform {
 
 /// A record containing the exit code and optionally link percentage
 /// returned by `runLinker`.
-typedef LinkResult = ({int exitCode, double? linkPercentage});
+@immutable
+class LinkResult {
+  /// Creates a new [LinkResult] representing failure.
+  const LinkResult.failure()
```
felangel, line 33: wouldn't we want to have an error reason (e.g. from `stderr`) in the case of a failure?

packages/shorebird_cli/lib/src/platform/apple.dart
eseidel: We don't.  The only advantage of this is to let us have const constructors.

packages/shorebird_cli/lib/src/platform/apple.dart
eseidel: We certainly could.  I was just matching what the existing code did, just moving it from being a record type to an explicit class with a named constructor (with appropriate defaults).

packages/shorebird_cli/lib/src/platform/apple.dart
```diff
@@ -289,12 +313,30 @@ $error''');
       );
     } on Exception catch (error) {
       linkProgress.fail('Failed to link AOT files: $error');
-      return (exitCode: ExitCode.software.code, linkPercentage: null);
+      return const LinkResult.failure();
     } finally {
       await dumpDebugInfo();
     }
+    Map<String, dynamic>? linkMetadata;
```
felangel, line 320:
```suggestion
    final Map<String, dynamic>? linkMetadata;
```

packages/shorebird_cli/lib/src/platform/apple.dart
eseidel: The reason why we can't have a const constructor if we hold onto the int directly is that ExitCode.code isn't const-compatible.

packages/shorebird_cli/test/src/executables/aot_tools_test.dart
```diff
@@ -950,6 +950,69 @@ Run "aot_tools help <command>" for more information about a command.
...
+            ).thenAnswer((_) => Stream.value(utf8.encode(stderr)));
+            return mockProcess;
+          });
+          when(
+            () => process.start(
+              any(),
+              any(),
+              workingDirectory: any(named: 'workingDirectory'),
+            ),
+          ).thenAnswer((invocation) async {
```
felangel:
```suggestion
          ).thenAnswer((_) async {
```

packages/shorebird_cli/test/src/executables/aot_tools_test.dart
```diff
@@ -950,6 +950,69 @@ Run "aot_tools help <command>" for more information about a command.
...
+            ).thenAnswer((_) => Stream.value(utf8.encode(stdout)));
+            when(
+              () => mockProcess.stderr,
+            ).thenAnswer((_) => Stream.value(utf8.encode(stderr)));
+            return mockProcess;
+          });
+        });
+
+        test('returns link metadata', () async {
+          stdout = '{}';
```
felangel, line 999: why is this stdout assignment needed?

packages/shorebird_cli/test/src/executables/aot_tools_test.dart
```diff
@@ -950,6 +950,69 @@ Run "aot_tools help <command>" for more information about a command.
           expect(result.existsSync(), isTrue);
         });
       });
+
+      group('getLinkMetadata', () {
+        late int exitCode;
+        late String stdout;
+        late String stderr;
+        setUp(() {
```
felangel:
```suggestion
        late String stderr;

        setUp(() {
```
nit

packages/shorebird_cli/lib/src/platform/apple.dart
eseidel, line 320: We can't as I've currently written it, since the analyzer can't be sure it' isn't set twice.

packages/shorebird_cli/test/src/executables/aot_tools_test.dart
```diff
@@ -950,6 +950,69 @@ Run "aot_tools help <command>" for more information about a command.
...
+              () => mockProcess.stdout,
+            ).thenAnswer((_) => Stream.value(utf8.encode(stdout)));
+            when(
+              () => mockProcess.stderr,
+            ).thenAnswer((_) => Stream.value(utf8.encode(stderr)));
+            return mockProcess;
+          });
+        });
+
+        test('returns link metadata', () async {
```
felangel:
```suggestion
        test('returns link metadata when aot_tools executes successfully', () async {
```

packages/shorebird_cli/lib/src/platform/apple.dart
eseidel: the `if (dumpDebugInfoDir == null)` branch seems to confuse it into believing it might be set twice.

packages/shorebird_cli/test/src/executables/aot_tools_test.dart
```diff
@@ -950,6 +950,69 @@ Run "aot_tools help <command>" for more information about a command.
...
+          stdout = '{}';
+          final result = await runWithOverrides(
+            () => aotTools.getLinkMetadata(debugDir: '/debug'),
+          );
+          expect(result, isA<Map<String, dynamic>>());
+        });
+
+        test('returns invalid json', () async {
```
felangel:
```suggestion
        test('throws FormatException when aot_tools outputs invalid json', () async {
```

felangel: Shouldn't we also have a test to ensure that when aot_tools exits with a non-zero exit code we treat it as a failure?

eseidel: We don't treat it as a failure.  If we can't get the metadata we just keep going.  It doesn't block the user.

packages/shorebird_cli/lib/src/platform/apple.dart
```diff
@@ -289,12 +313,30 @@ $error''');
...
+    try {
+      if (dumpDebugInfoDir == null) {
+        linkMetadata = null;
+      } else {
+        linkMetadata = await aotTools.getLinkMetadata(
+          debugDir: dumpDebugInfoDir.path,
+          workingDirectory: buildDirectory.path,
+        );
+      }
```
felangel:
```suggestion
      if (dumpDebugInfoDir != null) {
        linkMetadata = await aotTools.getLinkMetadata(
          debugDir: dumpDebugInfoDir.path,
          workingDirectory: buildDirectory.path,
        );
      }
```

packages/shorebird_cli/lib/src/platform/apple.dart
```diff
@@ -289,12 +313,30 @@ $error''');
...
+    } on Exception catch (error) {
+      logger.detail('Failed to get link metadata: $error');
```
felangel:
```suggestion
      logger.detail('[aot_tools] Failed to get link metadata: $error');
```
nit

packages/shorebird_cli/lib/src/commands/patch/ios_framework_patcher.dart
```diff
@@ -175,13 +179,22 @@ class IosFrameworkPatcher extends Patcher {
...
         vmCodeFile: File(_vmcodeOutputPath),
       );
+      final linkPercentage = result.linkPercentage;
+      final exitCode = result.exitCode;
+      if (exitCode != ExitCode.success.code) throw ProcessExit(exitCode);
+      if (linkPercentage != null &&
+          linkPercentage < Patcher.linkPercentageWarningThreshold) {
+        logger.warn(Patcher.lowLinkPercentageWarning(linkPercentage));
+      }
+      lastBuildLinkPercentage = linkPercentage;
```
felangel, line 196: nice so this was never being reported before?

packages/shorebird_cli/lib/src/platform/apple.dart
felangel: yeah makes sense, you can ignore my comment 👍

packages/shorebird_cli/lib/src/platform/apple.dart
felangel: Isn't the failure case's exit code always software? Can't we avoid passing an ExitCode altogether?

packages/shorebird_cli/lib/src/platform/apple.dart
felangel: yeah makes sense it's fine as is 👍

packages/shorebird_cli/lib/src/commands/patch/ios_framework_patcher.dart
eseidel: Correct, the ios_framework side was not reporting link percentages.

eseidel: Mostly all of this just needs to be pulled out of these leaf classes and into some shared logic.

felangel: probably should add a unit test to make sure there's no regression in the future 🤷

packages/shorebird_cli/lib/src/platform/apple.dart
eseidel: Do you mean just store a bool instead of the code?

packages/shorebird_cli/lib/src/commands/patch/ios_framework_patcher.dart
eseidel: Happy to.

eseidel: Done.

PR #3081, opened 2025-05-02
feat: add support for dispatch table supplement files

eseidel:


felangel (approved):

felangel:
Looks like tests are failing since you're using a logger now and the tests are injected a scoped mock logger.

packages/shorebird_cli/lib/src/platform/apple.dart
```diff
@@ -153,13 +153,19 @@ class Apple {
...
     // This uses maybeCopy because not all versions of gen_snapshot/aot_tools
     // use the same supplement files. At the `shorebird` level we don't know
     // which files should be present, so we just try to copy all.
     void maybeCopy(File file, Directory destDir, {String? newBaseName}) {
-      if (!file.existsSync()) return;
+      logger.detail('Copying supplement file ${file.path} to ${destDir.path}');
+      if (!file.existsSync()) {
+        logger.detail('Unable to find supplement file at ${file.path}');
+        return;
+      }
```
felangel, line 168: Can we also add a test to verify that if a file doesn't exist we continue and log?

eseidel: Yup, done.

felangel (approved):

PR #3113, opened 2025-05-24
feat: add shorebird flutter versions use command

Khurshidddbek:
# Add `shorebird flutter versions use` command

## Problem

Currently, there is no way to set a global Flutter version for Shorebird. Users must specify `--flutter-version` with every command when they want to use a specific Flutter version, even if it's the latest one. This is inconvenient and error-prone.

For example:
- When Flutter 3.32.0 is released, `shorebird flutter --version` shows an older version
- To use the new version, users must run: `shorebird release ios --flutter-version=3.32.0`
- This version parameter must be repeated for every command

## Solution

This PR introduces a new command: `shorebird flutter versions use <version>`

This command allows users to set a global Flutter version that will be used by all Shorebird commands.

### Usage

```bash
# List available Flutter versions
shorebird flutter versions list

# Set a specific Flutter version as the global version
shorebird flutter versions use 3.32.0

# Verify the change
shorebird flutter --version

# You can also use a git revision
shorebird flutter versions use abc123def456
```

### How it works

1. The command accepts either a semver version (e.g., `3.32.0`) or a git revision hash
2. It validates that the version exists in Shorebird's Flutter fork
3. If the version is not already installed, it downloads and installs it
4. It updates the `bin/internal/flutter.version` file with the git revision
5. All subsequent Shorebird commands will use this version by default

### Features

- ✅ Validates version existence before setting
- ✅ Shows helpful error messages if version not found
- ✅ Detects if the requested version is already active
- ✅ Supports both version numbers and git revision hashes
- ✅ Provides clear success feedback with verification instructions
- ✅ Comprehensive test coverage

### Implementation details

The implementation follows existing patterns in the Shorebird codebase:
- Uses the existing `shorebirdFlutter.resolveFlutterRevision` to handle version/revision resolution
- Uses the existing `shorebirdFlutter.installRevision` to ensure the Flutter version is installed
- Updates the same `bin/internal/flutter.version` file that Shorebird already uses
- Follows the same command structure as other `shorebird flutter versions` subcommands

### Testing

The PR includes comprehensive unit tests covering:
- Missing version arguments
- Invalid versions
- Already active versions
- Successful version changes
- Installation failures
- Git revision support

## Related issues

This addresses the need for a more convenient way to manage Flutter versions in Shorebird, as discussed in various community channels.

## Checklist

- [x] I've added tests for my changes
- [x] I've updated the documentation (if applicable)
- [x] I've followed the existing code style and patterns
- [x] All tests pass locally

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_command.dart
@@ -9,6 +9,7 @@ class FlutterVersionsCommand extends ShorebirdCommand {
   /// {@macro flutter_versions_command}
   FlutterVersionsCommand() {
     addSubcommand(FlutterVersionsListCommand());
+    addSubcommand(FlutterVersionsUseCommand());
   }
 
   @override
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_use_command.dart
@@ -1,0 +1,127 @@
+import 'dart:io';
+
+import 'package:mason_logger/mason_logger.dart';
+import 'package:path/path.dart' as p;
+import 'package:shorebird_cli/src/logging/logging.dart';
+import 'package:shorebird_cli/src/shorebird_command.dart';
+import 'package:shorebird_cli/src/shorebird_env.dart';
+import 'package:shorebird_cli/src/shorebird_flutter.dart';
+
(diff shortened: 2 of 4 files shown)


felangel:
Thanks for the pull request but we explicitly removed the functionality because it caused a lot confusion for users. We opted to have a more explicit mechanism for specifying flutter version to avoid having the cli maintain any system wide state. See https://github.com/shorebirdtech/shorebird/pull/1802

Closing this for now but if you feel strongly about this I’d recommend filing an issue instead and clearly explaining the problem you’re facing so we can have a discussion before jumping to any specific implementation.

PR #3117, opened 2025-05-28
chore: make Release type equatable

bryanoltman:


packages/shorebird_code_push_protocol/lib/src/models/release.dart
```diff
@@ -8,7 +9,7 @@ part 'release.g.dart';
 /// A release can have zero or more patches applied to it.
 /// {@endtemplate}
 @JsonSerializable()
-class Release {
+class Release extends Equatable {
```
eseidel, line 14: Should it also be `@immutable`?  Typically equatable is only "safe" if also immutable.

eseidel (approved):

packages/shorebird_code_push_protocol/lib/src/models/release.dart
bryanoltman: Added

packages/shorebird_code_push_protocol/test/src/models/release_test.dart
```diff
@@ -21,5 +21,47 @@ void main() {
         equals(release.toJson()),
       );
     });
+
+    test('is equatable', () {
```
eseidel, line 25: This is fine, but what I typically do (to save a little space) is also test toJson in the same kind of test.  That way it's 

final one = {..};
final roundTrip = One.fromJson(one.toJson());
expect(one, roundTrip);
final two = {..};
expect(one, isNot(two);

eseidel (approved):

packages/shorebird_code_push_protocol/test/src/models/release_test.dart
bryanoltman: Discussed offline, going to leave as-is for now

PR #3133, opened 2025-06-02
chore: bump very_good_analysis to 8.0.0, fix lints

bryanoltman:


packages/shorebird_cli/lib/src/android_sdk.dart
```diff
@@ -116,7 +116,7 @@ class AndroidSdk {
   }
 
   /// The path to the `adb` executable.
-  late final adbPath = getPlatformToolsPath(
+  late final String? adbPath = getPlatformToolsPath(
```
felangel, line 119: I don't think we should make all these type inference changes unless there's a clear reason. The analyzer can infer the types just fine so I don't think there's much value in arbitrarily specifying types in some places.

bryanoltman: Agreed about the analyzer, although there are some places where the code goes and the analyzer does not (reading on github, for example). I don't have super strong feelings either way.

felangel:
Closing for now in favor of #3135 
Happy to revisit if you feel strongly.

bryanoltman:
Nope, sg!

PR #3191, opened 2025-06-24
refactor: add initial shorebird_process_tools package

bryanoltman:
## Description

Pulls git logic and deps out of shorebird_cli for use in other packages.

This is still a bit proof-of-concepty and will leave the cli in an odd state where some classes (other executables) that live in the cli's executables folder.

If we like this approach, we can move those other executable support classes over to this package.

Also, very much not married to the package name.

## Type of Change

- [x] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -87,6 +87,11 @@ jobs:
               - ./.github/workflows/main.yaml
               - ./.github/actions/dart_package/action.yaml
               - packages/shorebird_code_push_protocol/**
+            shorebird_process_tools:
+              - ./.github/codecov.yml
+              - ./.github/workflows/main.yaml
+              - ./.github/actions/dart_package/action.yaml
+              - packages/shorebird_process_tools/**
             jwt:
               - ./.github/codecov.yml
               - ./.github/workflows/main.yaml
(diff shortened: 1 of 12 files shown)


felangel:
I think we might want to split this up into two packages: `shorebird_process` and `shorebird_git` to avoid having a catch-all package for any/every executable shorebird ever interacts with.

bryanoltman:
> I think we might want to split this up into two packages: `shorebird_process` and `shorebird_git` to avoid having a catch-all package for any/every executable shorebird ever interacts with.

I was thinking something similar, although I worry splitting them up too much leads us with a ton of tiny packages to maintain. Specifically worried about clutter, the work we'd need to do when upgrading our Dart version, etc. Any ideas how we could avoid having a everything-in-one package without an overwhelming number of packages?

felangel:
> > I think we might want to split this up into two packages: `shorebird_process` and `shorebird_git` to avoid having a catch-all package for any/every executable shorebird ever interacts with.
> 
> I was thinking something similar, although I worry splitting them up too much leads us with a ton of tiny packages to maintain. Specifically worried about clutter, the work we'd need to do when upgrading our Dart version, etc. Any ideas how we could avoid having a everything-in-one package without an overwhelming number of packages?

Dart workspaces should help here and I’m not really too worried about that. Also, might be worthwhile taking a bit of time to align since the approach I was exploring yesterday wouldn’t require us to do all this work.

felangel:
I think we can close this for now and revisit if needed.

PR #3192, opened 2025-06-24
refactor(shorebird_cli): update shorebird_cli to use shorebird_process_tools

bryanoltman:
Completes the refactor in the parent PR


felangel:
Per my comment on the parent PR I would prefer having `pkg:shorebird_process` which exposes our `Process` wrapper and a separate `pkg:shorebird_git` which exposes the public git APIs to avoid having a single catch-all package for all executables shorebird uses.

felangel:
I think we can close this for now and revisit if needed.

PR #3226, opened 2025-07-15
fix(shorebird_cli): don't append +0 to build numbers that don't include metadata

bryanoltman:
## Description

The changes in https://github.com/shorebirdtech/flutter/pull/92 update the engine to read the release version of a Windows executable as a string instead of as numbers. Previously, the label portion (i.e., the part of the version string following the "+") would default to `0` if missing. Now, because we read the entire version number as a string from the executable, we support release versions with no label.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/executables/powershell.dart
```diff
@@ -40,10 +40,6 @@ class Powershell {
 
     final result = await pwsh(['-Command', pwshCommand]);
 
-    var versionString = (result.stdout as String).trim();
-    if (!versionString.contains('+')) {
-      versionString += '+0';
-    }
-    return versionString;
+    return (result.stdout as String).trim();
```
felangel, line 43: I guess now it'll technically be possible for older versions of the CLI to use the new engine and in that case what will be the behavior?

bryanoltman: When releasing: older versions of the CLI will (almost certainly) be using an older version of the engine, so things will continue working as before.
When patching: the user needs to pick a release to patch in non-CI environments, so this shouldn't affect that flow. In CI environments, the latest version of shorebird should always be used.

bryanoltman: Although, thinking about it more, it would be possible for a CLI with this change to use an older Flutter engine that appends "+0" to the version number if no label is provided.

felangel: It's possible for an older version of the CLI to be used to release with the latest engine though right? In that case what would be the behavior?

felangel: Anyone who explicitly specifies the flutter version when releasing will use the latest engine revision

bryanoltman: Updated the minimum flutter version for windows releases to 3.32.6

bryanoltman: If a new release is created with an old version of the CLI and the latest Flutter version is specified (and their version number doesn't include a label), that release will not be patchable.

bryanoltman: I think that feels corner-case-y enough that I'm not worried about it? Trying to imagine the user who keeps up-to-date with Flutter but ignores "update available" messages from Shorebird.

felangel (approved):

Issue #2529, opened 2024-10-15
fix: users are confused by `shorebird` defaulting to latest Flutter


eseidel:
It's somewhat common that we get reports from users that their app doesn't build with Shorebird, but works fine with Flutter.  What's happening (in all the cases I've seen) is that `shorebird` is using the latest Flutter stable, and these users are otherwise using an older Flutter version.  Whether through fvm or their system flutter.

I believe we show a warning for users, but that's clearly not enough. One solution might be to show the warning right after printing any failure message.  Another might be to default `shorebird` to use the same flutter as is on their system path (or maybe even .fvm file).

https://discord.com/channels/1030243211995791380/1293167977667231774 is another recent example.

eseidel:
Another example: https://github.com/shorebirdtech/shorebird/issues/2434#issuecomment-2412688486

escamoteur:
I think it would be really helpful if it would respect FVM as CodeMagic does too.

eseidel:
I think there is just a question of how much of fvm to support.  I guess just reading the version number? https://github.com/shorebirdtech/shorebird/issues/1385

escamoteur:
Yep, I guess it's just that. Reading the version number and then use that for all Shorebird commands
Am 17. Dez. 2024, 17:54 +0100 schrieb Eric Seidel ***@***.***>:

felangel:
The main concern I have about supporting FVM is where do we draw the line? There are other Flutter version management tools like [Puro](https://puro.dev/) and I worry that as soon as we support one it's totally fair for customers to ask us to support any/all others. Long term, I think Shorebird should be your Flutter version manager so I'd personally rather move in that direction but would love to hear what others think.

I'd also love to understand how and why folks are using tools like fvm. For example, are you using it to easily cache/switch global flutter versions or are you using it at the project level exclusively?

felangel:
We've updated our shorebird release action to require a flutter version to help mitigate some of this confusion.

felangel:
Technically to fully resolve this issue, we'd need to push a breaking change to Shorebird CLI which makes `--flutter-version` required when running `shorebird release`.

lukemmtt:
Some thoughts on your questions, @felangel:

> The main concern I have about supporting FVM is where do we draw the line? There are other Flutter version management tools like Puro and I worry that as soon as we support one it's totally fair for customers to ask us to support any/all others. 

[...]

PR #3238, opened 2025-07-17
feat: support --flutter-version=fvm and --flutter-version=system

### LINKED ISSUE #1385: feat: support fvm
Some users don't install Flutter directly, rather they use fvm.  So `flutter` isn't even on their path, just `fvm` is and they might then use a shell alias to map `flutter` in the shell to `fvm flutter`.

One such user noted that this setup does not work with shorebird and is asking for potential fixes to `shorebird` to auto-detect/support fvm, and/or at least docs on how they're supposed to work together:
https://github.com/shorebirdtech/shorebird/issues/472#issuecomment-1754490057

eseidel:
I chose this approach so we didn't have to write any code to read .fvmrc files, and all we do is just know to run `fvm flutter` rather than `flutter` when determining the version from the system-available flutter.

Fixes https://github.com/shorebirdtech/shorebird/issues/1385

And is part of potentially fixing https://github.com/shorebirdtech/shorebird/issues/2529.

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_command.dart
@@ -263,7 +263,9 @@ of the iOS app that is using this module. (aar and ios-framework only)''',
     // This command handles logging, we don't need to provide our own
     // progress, error logs, etc.
     final app = await codePushClientWrapper.getApp(appId: appId);
-    final targetFlutterRevision = await resolveTargetFlutterRevision();
+    final targetFlutterRevision = await resolveTargetFlutterRevision(
+      flutterVersionArg,
+    );
     try {
       await shorebirdFlutter.installRevision(revision: targetFlutterRevision);
     } on Exception {
@@ -368,48 +370,6 @@ For more information see: ${supportedFlutterVersionsUrl.toLink()}''');
 
     // Ask the releaser to assert its own args are valid.
     await releaser.assertArgsAreValid();
-  }
-
-  /// Determines which Flutter version to use for the release. This will be
-  /// either the version specified by the user or the version provided by
-  /// [shorebirdEnv]. Will exit with [ExitCode.software] if the version
-  /// specified by the user is not found/supported.
-  Future<String> resolveTargetFlutterRevision() async {
-    if (flutterVersionArg == 'latest') return shorebirdEnv.flutterRevision;
-    if (flutterVersionArg == 'system') {
-      return flutterVersionFromSystemFlutter(process);
-    }
-    if (flutterVersionArg == 'fvm') {
-      return flutterVersionFromFVMFlutter(process);
-    }
-
-    final String? revision;
-    try {
-      revision = await shorebirdFlutter.resolveFlutterRevision(
-        flutterVersionArg,
-      );
-    } on Exception catch (error) {
-      logger.err('''
-Unable to determine revision for Flutter version: $flutterVersionArg.
-$error''');
-      throw ProcessExit(ExitCode.software.code);
-    }
-
-    if (revision == null) {
-      final openIssueLink = link(
-        uri: Uri.parse(
-          'https://github.com/shorebirdtech/shorebird/issues/new?assignees=&labels=feature&projects=&template=feature_request.md&title=feat%3A+',
-        ),
-        message: 'open an issue',
-      );
-      logger.err('''
-Version $flutterVersionArg not found. Please $openIssueLink to request a new version.
-Use `shorebird flutter versions list` to list available versions.
-''');
-      throw ProcessExit(ExitCode.software.code);
-    }
-
-    return revision;
   }
 
   /// Asserts that a release with version [version] can be released using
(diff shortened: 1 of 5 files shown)


packages/shorebird_cli/lib/src/executables/flutter_version.dart
```diff
@@ -0,0 +1,41 @@
...
+///
+/// The output of `flutter --version` is of the form:
+///
+/// ```text
+/// Flutter 3.32.4 • channel stable • https://github.com/flutter/flutter.git
+/// Framework • revision 6fba2447e9 (5 weeks ago) • 2025-06-12 19:03:56 -0700
+/// Engine • revision 8cd19e509d (5 weeks ago) • 2025-06-12 16:30:12 -0700
+/// Tools • Dart 3.8.1 • DevTools 2.45.1
+/// ```
+String parseFlutterVersionFromOutput(String output) {
```
felangel, line 22: I believe we already have this logic

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -369,6 +376,12 @@ For more information see: ${supportedFlutterVersionsUrl.toLink()}''');
   /// specified by the user is not found/supported.
   Future<String> resolveTargetFlutterRevision() async {
     if (flutterVersionArg == 'latest') return shorebirdEnv.flutterRevision;
+    if (flutterVersionArg == 'system') {
+      return flutterVersionFromSystemFlutter(process);
```
felangel:
```suggestion
      return shorebirdFlutter.getSystemVersion()
```

eseidel: That returns Future<String?> instead, so keeping the new version for now.

packages/shorebird_cli/lib/src/executables/flutter_version.dart
```diff
@@ -0,0 +1,48 @@
...
+///
+/// The output of `flutter --version` is of the form:
+///
+/// ```text
+/// Flutter 3.32.4 • channel stable • https://github.com/flutter/flutter.git
+/// Framework • revision 6fba2447e9 (5 weeks ago) • 2025-06-12 19:03:56 -0700
+/// Engine • revision 8cd19e509d (5 weeks ago) • 2025-06-12 16:30:12 -0700
+/// Tools • Dart 3.8.1 • DevTools 2.45.1
+/// ```
+String? maybeParseFlutterVersionFromOutput(String output) {
```
felangel, line 13: I still think we should consolidate this with `ShorebirdFlutter`. The intention of `ShorebirdFlutter` was to interface with the underlying Flutter installation. If we merged this we would have `flutter --version` execution and parsing in multiple places.

packages/shorebird_cli/lib/src/executables/flutter_version.dart
```diff
@@ -0,0 +1,48 @@
...
+    return parseFlutterVersionFromOutput(output.stdout as String);
+  } on Exception catch (e) {
+    throw Exception('Failed to get Flutter version from `flutter` on path: $e');
+  }
+}
+
+/// Returns the Flutter version from a FVM-installed Flutter if available.
+String flutterVersionFromFVMFlutter(ShorebirdProcess process) {
+  try {
+    final output = process.runSync('fvm', ['flutter', '--version']);
```
felangel, line 43: What if `fvm` isn't available on the PATH? I can imagine cases where users create aliases for `flutter` which proxy to the underlying `fvm` installation behind the scenes 🤷 

See also:
> Tighter VSCode integration, with configuration and settings management. FVM will now automatically configure VSCode to use the correct Flutter SDK version, triggering a terminal path update, so you can just use flutter, commands instead of fvm flutter

https://pub.dev/packages/fvm/changelog#300

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -6,6 +6,7 @@ import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/executables/executables.dart';
+import 'package:shorebird_cli/src/executables/flutter_version.dart';
```
felangel, line 9: This import shouldn't be needed if we export `flutter_version` from `executables/executables.dart`

Also the convention would be to create a class called `Flutter` under executables which exposes whatever we need from the underlying `flutter` tool. This is what the `ShorebirdFlutter` class was meant to do.

packages/shorebird_cli/test/src/commands/release/release_command_test.dart
```diff
@@ -64,6 +65,7 @@ void main() {
     late ShorebirdEnv shorebirdEnv;
     late ShorebirdFlutter shorebirdFlutter;
     late ShorebirdValidator shorebirdValidator;
+    late ShorebirdProcess process; // Only used by resolveTargetFlutterRevision.
```
felangel, line 68: Ideally we should just mock the direct dependency (e.g. `ShorebirdFlutter` or `Flutter`) and not have to mock a transitive dependency here.

eseidel:
Closing — branch is 230 commits behind main with conflicts in release_command.dart, original reviewer (@bryanoltman) is no longer at Shorebird, and there's been no movement since July 2025. Will reopen as a fresh PR if/when the fvm/system flutter-version support is needed again.

PR #3266, opened 2025-08-01
feat: add flutter_version_resolver package

bryanoltman:
## Description

Adds a minimal cli utility to resolve a Flutter version from an arbitrary Flutter app package.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/flutter_version_resolver/README.md
+++ b/packages/flutter_version_resolver/README.md
@@ -1,0 +1,4 @@
+# Flutter Version Resolver
+
+Analyzes Flutter packages to determine the Flutter version that should be used
+to build and test.
--- a/packages/flutter_version_resolver/bin/flutter_version_resolver.dart
+++ b/packages/flutter_version_resolver/bin/flutter_version_resolver.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:flutter_version_resolver/flutter_version_resolver.dart';
 import 'package:flutter_version_resolver/src/logger.dart';
+import 'package:mason_logger/mason_logger.dart';
 import 'package:scoped_deps/scoped_deps.dart';
 
 /// Resolves the Flutter version for a package and optionally writes it to a
@@ -11,13 +12,13 @@ import 'package:scoped_deps/scoped_deps.dart';
 /// ```sh
 /// dart run bin/flutter_version_resolver.dart <path-to-package> [<output-file>]
 /// ```
-Future<void> main(List<String> arguments) async {
-  return runScoped(() async {
+Future<int> main(List<String> arguments) async {
+  return await runScoped(() async {
     if (arguments.isEmpty || arguments.length > 2) {
       logger.err(
         'Usage: dart run bin/flutter_version_resolver.dart <path-to-package> [<output-file>]',
       );
-      return;
+      return ExitCode.usage.code;
     }
 
     final packageDirectory = Directory(arguments[0]);
@@ -25,7 +26,7 @@ Future<void> main(List<String> arguments) async {
       logger.err(
         'Package directory does not exist: ${packageDirectory.path}',
       );
-      return;
+      return ExitCode.usage.code;
     }
 
     final flutterVersion = resolveFlutterVersion(
@@ -39,5 +40,7 @@ Future<void> main(List<String> arguments) async {
         ..writeAsStringSync(flutterVersion);
       logger.info('Wrote flutter version to ${outputFile.path}');
     }
+
+    return ExitCode.success.code;
   }, values: {loggerRef});
 }
--- a/packages/flutter_version_resolver/bin/flutter_version_resolver.dart
+++ b/packages/flutter_version_resolver/bin/flutter_version_resolver.dart
@@ -1,0 +1,43 @@
+import 'dart:io';
+
+import 'package:flutter_version_resolver/flutter_version_resolver.dart';
+import 'package:flutter_version_resolver/src/logger.dart';
+import 'package:scoped_deps/scoped_deps.dart';
+
+/// Resolves the Flutter version for a package and optionally writes it to a
+/// file.
+///
+/// Usage:
+/// ```sh
+/// dart run bin/flutter_version_resolver.dart <path-to-package> [<output-file>]
+/// ```
+Future<void> main(List<String> arguments) async {
+  return runScoped(() async {
+    if (arguments.isEmpty || arguments.length > 2) {
+      logger.err(
+        'Usage: dart run bin/flutter_version_resolver.dart <path-to-package> [<output-file>]',
+      );
+      return;
+    }
+
+    final packageDirectory = Directory(arguments[0]);
+    if (!packageDirectory.existsSync()) {
+      logger.err(
+        'Package directory does not exist: ${packageDirectory.path}',
+      );
+      return;
+    }
+
+    final flutterVersion = resolveFlutterVersion(
+      packagePath: packageDirectory.path,
+    );
+    logger.info('Resolved Flutter version: $flutterVersion');
+
+    if (arguments.length > 1) {
+      final outputFile = File(arguments[1])
+        ..createSync(recursive: true)
+        ..writeAsStringSync(flutterVersion);
+      logger.info('Wrote flutter version to ${outputFile.path}');
+    }
+  }, values: {loggerRef});
+}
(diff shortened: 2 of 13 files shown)


packages/flutter_version_resolver/lib/flutter_version_resolver.dart
```diff
@@ -0,0 +1,63 @@
...
+/// pubspec.yaml file, or `null` if no version is specified.
+///
+/// A pubspec.yaml with the following will return `Version(3, 20, 0)`:
+/// ```yaml
+/// environment:
+///   sdk: ^3.8.1
+///   flutter: 3.20.0
+/// ```
+Version? flutterVersionFromPubspecEnvironment({required String packagePath}) {
+  final pubspecFile = File(p.join(packagePath, 'pubspec.yaml'));
```
felangel, line 69: We should try looking for the `pubspec.lock` first imo but I'm totally fine if we want to add that as a later step.

packages/flutter_version_resolver/test/flutter_version_resolver_test.dart
```diff
@@ -0,0 +1,130 @@
...
+        runWithOverrides(() {
+          expect(
+            resolveFlutterVersion(packagePath: packageDirectory.path),
+            equals('stable'),
+          );
+        });
+      });
+    });
+
+    group('when a flutter version is specified in the pubspec.yaml file', () {
```
felangel, line 49: should we add a case for `when a flutter version range is specified`

bryanoltman: You mean like on line 92? :P

packages/flutter_version_resolver/test/flutter_version_resolver_test.dart
```diff
@@ -0,0 +1,130 @@
...
+    group('when a flutter version range is specified', () {
+      setUp(() {
+        pubspecFile.writeAsStringSync('''
+environment:
+  sdk: ^3.8.1
+  flutter: ">=3.8.0 <4.0.0"
+''');
+      });
+
+      test('throws an exception', () {
```
felangel: is this just temporary? I'd expect we'd return `Version.range` or whatever the constructor is called.

bryanoltman: For now. I think we'll need to get more sophisticated to support ranges (i.e., we'll have to actually know what Flutter versions exist to be able to pick one)

bryanoltman: Although I think it could fail more gracefully. One sec

packages/flutter_version_resolver/lib/flutter_version_resolver.dart
bryanoltman: The lock file uses ranges, which we definitely should support, but as mentioned elsewhere, that requires us to be a bit more sophisticated than we are at present.

packages/flutter_version_resolver/README.md
```diff
@@ -0,0 +1,4 @@
+# Flutter Version Resolver
```
felangel, line 1: We should also update our ci (for now) to include this package 😛

felangel, line 1: We should also include this new package in the root `README`

packages/flutter_version_resolver/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
+name: flutter_version_resolver
+description: A CLI app that determines whether
```
felangel:
```suggestion
description: A Dart tool that resolves a Flutter version for a given Flutter application
```

packages/flutter_version_resolver/test/flutter_version_resolver_test.dart
felangel: yeah but we never explicitly cover `resolveFlutterVersion` with a version range as far as I can tell.

packages/flutter_version_resolver/test/flutter_version_resolver_test.dart
felangel: I was under the impression that we'd have two pieces: one that parses the pubspec lock and yaml etc. and returns the `Version` (which could be a range) and then a separate piece which resolves/solves to a specific explicit version

pubspec.yaml
```diff
@@ -5,6 +5,7 @@ environment:
 workspace:
   - packages/artifact_proxy
   - packages/discord_gcp_alerts
+  - packages/flutter_version_resolver
```
felangel, line 8: might be nice to split this into `flutter_version_parser` and `flutter_version_solver` 🤷 
we can always do that later but I think there are conceptually two independent pieces here at least the way I was approaching this problem.

packages/flutter_version_resolver/README.md
felangel, line 1: We should also add a LICENSE and CONTRIBUTING file + `example/main.dart`

packages/flutter_version_resolver/bin/flutter_version_resolver.dart
```diff
@@ -0,0 +1,43 @@
...
+      );
+      return;
+    }
+
+    final packageDirectory = Directory(arguments[0]);
+    if (!packageDirectory.existsSync()) {
+      logger.err(
+        'Package directory does not exist: ${packageDirectory.path}',
+      );
+      return;
```
felangel: we should probably exit with a non-zero exit code in this case.

packages/flutter_version_resolver/bin/flutter_version_resolver.dart
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    final flutterVersion = resolveFlutterVersion(
+      packagePath: packageDirectory.path,
+    );
+    logger.info('Resolved Flutter version: $flutterVersion');
+
+    if (arguments.length > 1) {
```
felangel, line 37: shouldn't we fail if arguments.length isn't exactly 1 with a usage exception?

packages/flutter_version_resolver/bin/flutter_version_resolver.dart
```diff
@@ -0,0 +1,43 @@
...
+/// ```sh
+/// dart run bin/flutter_version_resolver.dart <path-to-package> [<output-file>]
+/// ```
+Future<void> main(List<String> arguments) async {
+  return runScoped(() async {
+    if (arguments.isEmpty || arguments.length > 2) {
+      logger.err(
+        'Usage: dart run bin/flutter_version_resolver.dart <path-to-package> [<output-file>]',
+      );
+      return;
```
felangel: same here about exiting with a non-zero exit code (usage exception).

packages/flutter_version_resolver/test/flutter_version_resolver_test.dart
bryanoltman: Done

packages/flutter_version_resolver/README.md
bryanoltman: Yep, that's next on the agenda :)

bryanoltman: Done

bryanoltman: Added license and contributing. I think I'm going to punt on the example for now, as this is currently just being build as a CLI utility.

packages/flutter_version_resolver/bin/flutter_version_resolver.dart
bryanoltman: Done

packages/flutter_version_resolver/lib/flutter_version_resolver.dart
felangel: Sounds good I guess I'm unclear on what the scope of this package is

packages/flutter_version_resolver/bin/flutter_version_resolver.dart
bryanoltman: Nope, because we want to be able to output the version to a file. My hope was to provide really detailed logs for the process of resolving the version and enable writing a single string to a file (e.g., `resolved.flutter.version`) that our CI will consume.

packages/flutter_version_resolver/bin/flutter_version_resolver.dart
bryanoltman: Done.

pubspec.yaml
bryanoltman: Yeah, we can split this out in the future when we actually add any sort of resolution logic, rather than just rummaging through various files to find an exact version number.

PR #3295, opened 2025-08-27
feat: add a flutter versions precache command

### LINKED ISSUE #3294: feat: Would like a way to pre-cache flutter within a docker container
Shorebird used to have `shorebird versions use` which would download/install a flutter version and customers used this to pre-cache Shorebird artifacts within a docker container.
https://github.com/sncf-connect-tech/shorebird-docker/blob/main/Dockerfile-shorebird-android

We've removed `shorebird versions use`, so they'd like an equivalent way to do this pre-caching.

e.g. maybe `shorebird versions precache X.Y.Z` since we still have the "versions" subcommand?  This could even call `flutter precache` as well, although my understanding is `flutter precache` doesn't actually install *everything* it needs, just most things (e.g. it won't do the gradle downloads for you).

eseidel:
wip, this isn't ready for review (no tests yet, not sure the API is correct).

Fixes https://github.com/shorebirdtech/shorebird/issues/3294

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_precache_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/flutter/versions/flutter_versions_precache_command.dart
@@ -1,0 +1,50 @@
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/cache.dart';
+import 'package:shorebird_cli/src/logging/logging.dart';
+import 'package:shorebird_cli/src/shorebird_command.dart';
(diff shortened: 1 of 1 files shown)


felangel:
imo the correct solution to the linked issue isn't to include the shorebird artifacts in the docker image itself and instead to mount a shared volume where the shorebird installation (along with any artifacts lives). This will keep the image small and easier to maintain and pull while avoiding having to repeatedly install the same artifacts over and over again.

felangel:
Closing this for now

PR #3297, opened 2025-08-28
feat: add a script to check xcframework signatures

eseidel:
We've had a few scripts like this in build_engine before
but seems like we should just put these in our public repo.

Output:

❌ 3.16.9 (eng e8988540)
✅ 3.19.0 (eng 62fa4e8b)
cache/engine/66955014ffcb210ed7d1e82aa194d232f4aa0210/ios-release/Flutter.xcframework: code object is not signed at all

❌ 3.19.0-0.4.pre (eng 66955014)
✅ 3.19.1 (eng 62fa4e8b)
✅ 3.19.2 (eng 62fa4e8b)
✅ 3.19.3 (eng f0d3a823)
✅ 3.19.4 (eng 9a4a37dd)
✅ 3.19.5 (eng 255003dc)
✅ 3.19.6 (eng ef2c5cd9)
✅ 3.22.0 (eng fb9c4108)
✅ 3.22.1 (eng e5fdf87c)
✅ 3.22.2 (eng 2a9f67b0)
✅ 3.22.3 (eng b633ef95)
✅ 3.24.0 (eng 96854a8f)
✅ 3.24.1 (eng 5d31e777)

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_env.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_env.dart
@@ -51,7 +51,18 @@ class ShorebirdEnv {
   ///
   /// Assumes we are running from $ROOT/bin/cache.
   Directory get shorebirdRoot {
-    return File(platform.script.toFilePath()).parent.parent.parent;
+    final script = platform.script.toFilePath();
+    final lastPathComponent = p.basename(p.dirname(script));
+    // We're being run directly (e.g. `dart run shorebird_cli`)
+    if (lastPathComponent == 'bin' || lastPathComponent == 'tool') {
+      return File(script).parent.parent.parent.parent;
+    }
+    if (lastPathComponent != 'cache') {
+      throw StateError('Unexpected script path: $script');
+    }
+    // We're being run from within the `shorebird` shell script.
+    // and we're in $ROOT/bin/cache.
+    return File(script).parent.parent.parent;
   }
 
   /// The Shorebird engine revision.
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/shorebird_env.dart
```diff
@@ -51,7 +51,18 @@ class ShorebirdEnv {
...
-    return File(platform.script.toFilePath()).parent.parent.parent;
+    final script = platform.script.toFilePath();
+    final lastPathComponent = p.basename(p.dirname(script));
+    // We're being run directly (e.g. `dart run shorebird_cli`)
+    if (lastPathComponent == 'bin' || lastPathComponent == 'tool') {
+      return File(script).parent.parent.parent.parent;
+    }
+    if (lastPathComponent != 'cache') {
+      throw StateError('Unexpected script path: $script');
+    }
```
felangel, line 62: Not sure it's worth supporting multiple entrypoints.
If you want to run a local version you can `dart pub global activate --source path ./path/to/shorebird_cli` rather than changing the assumptions to also support `dart run shorebird_cli/bin/...`

eseidel: The current entrypoint is just confusing.  If you try to run shorebird.dart as a normal dart package it fails.

felangel: that's because it's not meant to be used as a normal dart package -- it's meant to be used as a precompiled executable 🤷

PR #3319, opened 2025-09-18
fix(windows): select correct executable when multiple executables are generated

kingdomseed:
Diagnostics

- Add selector to choose application exe
  - Prefer exe by pubspec projectName; fallback preserves legacy behavior
- Wire selector into Windows releaser/patcher; log candidates and selection under --verbose
- Powershell: log inspected exe path and command
- Rename local var to projectName for clarity; add doc comments
- Add unit test for selector
- Windows-only; no changes to other platforms

## Description

On Windows, version detection could incorrectly select a helper binary (e.g., Sentry’s crashpad handler), returning the wrong ProductVersion. This PR adds a Windows exe selector that:
- Selects the application exe using the pubspec projectName (exact, then contains), and falls back to the first candidate to preserve legacy behavior.

It wires the selector into the Windows releaser and patcher, and adds verbose diagnostics showing exe candidates, the projectName used, the selected exe, and the exact PowerShell command used for ProductVersion inspection. Concise doc comments were added. Includes a unit test covering exclusions, matching, and fallback.

## Type of Change

- [ ] 🧹 Code refactor
- [x] 📝 Documentation
- [ ] 🗑️ Chore
- [x] 🧪 Tests


packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
```diff
@@ -139,14 +141,32 @@ class WindowsPatcher extends Patcher {
...
+    final projectName = shorebirdEnv.getPubspecYaml()?.name;
+    final exesFound = outputDirectory
+        .listSync()
+        .whereType<File>()
+        .where((f) => p.extension(f.path).toLowerCase() == '.exe')
+        .map((f) => p.basename(f.path))
+        .toList();
+    logger
+      ..detail(
+        '[windows_patcher] EXEs found in archive root: \\${exesFound.join(', ')}',
```
bryanoltman: Why the `\\` here and throughout?

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
```diff
@@ -139,14 +141,32 @@ class WindowsPatcher extends Patcher {
...
+      )
+      ..detail('[windows_patcher] projectName: \\$projectName');
+    final exeFile = selectWindowsAppExe(
+      outputDirectory,
+      projectNameHint: projectName,
```
bryanoltman: This logic is identical in both the patcher and the releaser. It should probably exist in just one place

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
```diff
@@ -0,0 +1,63 @@
...
+    throw Exception('No .exe found in release artifact');
+  }
+
+  const excludedNames = <String>{
+    'crashpad_handler.exe',
+    'crashpad_wer.exe',
+    'dump_syms.exe',
+    'symupload.exe',
+    'flutter_tester.exe',
+  };
```
bryanoltman: In your project, what is the name of your app's exe? I wonder if, instead of creating a list of exe names to ignore, we could intelligently identify the right one so we aren't updating this each time we discover a new exe we want to ignore

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
kingdomseed: I actually am not sure. I think I got crossed eyed.

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
kingdomseed: Ah this is a good point. I can collapse it.

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
kingdomseed: My app is mythicgme2e. Crash pad was the one interfering. I think the exclusion list is just a backup for known or potential problems. But I agree--it's probably nicer to build this list (or no list) programmatically. Or have the list in a JSON file or something.

bryanoltman: If it's just the name from the `pubspec.yaml` (sounds like it is?), let's just search for that and we can probably call it day

kingdomseed: Sounds good. Yeah that's what we are searching for.

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
kingdomseed: Updated

packages/shorebird_cli/lib/src/commands/patch/windows_patcher.dart
kingdomseed: Removed.

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
kingdomseed: Updated.

packages/shorebird_cli/lib/src/executables/powershell.dart
```diff
@@ -34,9 +35,14 @@ class Powershell {
 
   /// Returns the version string of the given executable file.
   Future<String> getExeVersionString(File exeFile) async {
+    // Reads the Windows ProductVersion for the provided executable using
+    // PowerShell's Get-Item cmdlet.
```
bryanoltman:
```suggestion
```

packages/shorebird_cli/lib/src/executables/powershell.dart
```diff
@@ -34,9 +35,14 @@ class Powershell {
...
     final exePath = exeFile.path;
     final pwshCommand =
         "(Get-Item -Path '$exePath').VersionInfo.ProductVersion";
+    logger
+      ..detail('[powershell] Inspecting EXE for ProductVersion: $exePath')
+      ..detail('[powershell] Command: powershell.exe -Command $pwshCommand');
```
bryanoltman: I would move this `powershell.exe -Command $pswhCommand` log to the `pwsh` method if we want to keep it

packages/shorebird_cli/lib/src/windows/windows_app_version.dart
```diff
@@ -0,0 +1,34 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:shorebird_cli/src/executables/executables.dart';
+import 'package:shorebird_cli/src/logging/logging.dart';
+import 'package:shorebird_cli/src/windows/windows_exe_selector.dart';
+
+/// Returns the Windows application ProductVersion for the app executable in
+/// [releaseDir]. Uses [projectNameHint] to disambiguate if multiple `.exe`
+/// files are present. Logs helpful details with the provided [logTag].
```
bryanoltman: What is `releaseDir` in this context? How is `projectNameHint` used to disambiguate?

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
```diff
@@ -0,0 +1,48 @@
+import 'dart:io';
+import 'package:path/path.dart' as p;
+
+/// Selects the most likely application executable from a Windows release
+/// directory.
+///
+/// Selects the application executable by preferring matches based on
+/// [projectNameHint]. Falls back to the first `.exe` to preserve legacy
+/// behavior when no hint match is found.
+File selectWindowsAppExe(
```
bryanoltman:
```suggestion
File windowsAppExe(
```

bryanoltman: Same Qs here re: what a "projectNameHint" is and how it is used.

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
```diff
@@ -0,0 +1,48 @@
...
+        contains ??= f;
+      }
+    }
+
+    if (exact != null) return exact;
+    if (contains != null) return contains;
+  }
+
+  return candidates.first;
+}
```
bryanoltman:
```suggestion
bool _pathMatchesHint(String path, String hint) =>
    p.basename(path) == '$hint.exe' || p.basename(path) == '${hint}_win64.exe';

File windowsAppExe(Directory releaseDir, {String? projectNameHint}) {
  final exes = releaseDir
      .listSync()
      .whereType<File>()
      .where((f) => p.extension(f.path).toLowerCase() == '.exe')
      .toList();
  if (projectNameHint == null) {
    return exes.first;
  }

  final exactMatch = exes.singleWhereOrNull(
    (f) => _pathMatchesHint(f.path, projectNameHint),
  );

  if (exactMatch != null) {
    return exactMatch;
  }

  final fuzzyMatch = exes.firstWhereOrNull(
    (f) => p.basename(f.path).contains(projectNameHint),
  );

  if (fuzzyMatch != null) {
    return fuzzyMatch;
  }

  return exes.first;
}
```

bryanoltman: This presumes that if the hint has an exact match, it will have *exactly one* exact  match. I'd leave a comment about the _win64.exe suffix, as I'm not sure when that gets generated or why.

bryanoltman:
Looking good! fwiw, we generally prefer to avoid top-level functions, as that makes them impossible to mock with our mocking framework. I'd take a look at our `Apple` class or our `Linux` class to see how we've done this for other host platform-specific code.

packages/shorebird_cli/lib/src/executables/powershell.dart
kingdomseed: Got it. Moving it.

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
kingdomseed: Changing this to projectName because I think it's clearer. Or maybe it could be pubspecProjectName if you want to really be explicit.

packages/shorebird_cli/lib/src/windows/windows_app_version.dart
kingdomseed: projectNameHint is just the name of the project. The variable name is not great. Dropping the word "hint" is probably appropriate. This would be the name pulled from pubspec. So probably this comment should read uses projectNameHint/projectName to select the correct exe.

[releaseDir] is the root of the Windows release output that directly contains the app .exe (non-recursive), for example build/windows/x64/runner/Release, or the equivalent extracted directory from a zipped release.

I've made a couple of changes and I'll commit those.

packages/shorebird_cli/lib/src/windows/windows_exe_selector.dart
kingdomseed: I think there's no wind64 suffix. Just digging a bit I couldn't find one so I'll remove this.

kingdomseed:
> Looking good! fwiw, we generally prefer to avoid top-level functions, as that makes them impossible to mock with our mocking framework. I'd take a look at our `Apple` class or our `Linux` class to see how we've done this for other host platform-specific code.

Ok let me try this in it's own commit so you can review. 

Update: made the changes to `windows.dart`

felangel:
@kingdomseed thanks for the contribution! I can try to take a closer look at this tomorrow and hopefully should be able to get it merged shortly thereafter 👍

bryanoltman:
Pushed a change to fix some tests and make cspell happy, but it looks like there's some missing coverage. Should be an easy enough test to add

kingdomseed:
Sounds good. I'll add it in the morning.

kingdomseed:
Ok great I added the missing test.

bryanoltman:
Looks good! Have you tested to verify that it works as expected? If so, I'll merge in the morning

packages/shorebird_cli/lib/src/windows/windows_app_version.dart
```diff
@@ -0,0 +1,39 @@
...
+
+/// Returns the Windows application ProductVersion for the .exe in [releaseDir].
+/// Uses [projectName] (the pubspec name) to select the exe: prefers an exact
+/// match on '[projectName].exe', then a filename containing '[projectName]',
+/// otherwise falls back to the first .exe. Logs details with [logTag].
+/// [releaseDir] is the directory containing the release artifacts. In the
+/// release flow, [WindowsReleaser.getReleaseVersion()] passes the [Directory]
+/// returned by [artifactBuilder.buildWindowsApp()] directly here.
+/// Only top-level .exe files in [releaseDir] are considered (non-recursive).
+Future<String> getWindowsAppVersionFromDir(
```
felangel: Why isn't this part of `windows.dart`?

felangel:
I'll take this over and get this merged today, thanks!

bryanoltman (approved):

felangel (approved):

kingdomseed:
Woo! Very cool and fun to contribute. Thanks for all the reviews and questions!

Issue #3348, opened 2025-10-09
feat: more helpful message when shorebird.yaml file missing on patch promote


Peetee06:
**Description**
When running `shorebird patches promote` in a directory that doesn't contain a `shorebird.yaml`, a generic `Null check operator used on a null value` error message is thrown.

It would make debugging easier to have a helpful error message.

**Requirements**

- [ ] When shorebird.yaml can't be found when running `shorebird patches promote`, a helpful error message is printed that names the issue

**Additional Context**
Logs from the failing GitHub Actions workflow step:
```
Run SHOREBIRD_TOKEN=*** shorebird patches promote --patch-number=6 --release-version=1.41.0+1694 --verbose
Null check operator used on a null value
#0      PromoteCommand.run (package:shorebird_cli/src/commands/patches/promote_command.dart:45:50)
#1      CommandRunner.runCommand (package:args/command_runner.dart:212:27)
#2      CompletionCommandRunner.runCommand (package:cli_completion/src/command_runner/completion_command_runner.dart:89:18)
#3      ShorebirdCliCommandRunner.runCommand (package:shorebird_cli/src/shorebird_cli_command_runner.dart:205:32)
#4      ShorebirdCliCommandRunner.run.<anonymous closure> (package:shorebird_cli/src/shorebird_cli_command_runner.dart:132:19)
#5      _rootRun (dart:async/zone.dart:1525:13)
#6      _CustomZone.run (dart:async/zone.dart:1422:19)
#7      _runZoned (dart:async/zone.dart:2034:6)
#8      runZoned (dart:async/zone.dart:1960:10)
#9      runScoped (package:scoped_deps/src/scoped_deps.dart:57:10)
#10     ShorebirdCliCommandRunner.run (package:shorebird_cli/src/shorebird_cli_command_runner.dart:131:20)
#11     main.<anonymous closure>.<anonymous closure> (file:///home/runner/.config/shorebird/packages/shorebird_cli/bin/shorebird.dart:48:49)
#12     _rootRun (dart:async/zone.dart:1525:13)
#13     _CustomZone.run (dart:async/zone.dart:1422:19)
#14     _runZoned (dart:async/zone.dart:2034:6)
#15     runZoned (dart:async/zone.dart:1960:10)
#16     runScoped (package:scoped_deps/src/scoped_deps.dart:57:10)
[...]

PR #3349, opened 2025-10-09
feat(shorebird_cli): handle missing shorebird.yaml file gracefully when promoting a patch

Peetee06:
## Description

Prints a helpful error message when `shorebird.yaml` isn't found while promoting a patch and exits with `unavailable` code.
Resolves issue #3348

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/commands/patches/promote_command.dart
```diff
@@ -42,7 +42,14 @@ class PromoteCommand extends ShorebirdCommand {
...
     final patchNumber = int.parse(results['patch-number'] as String);
     final flavor = results.findOption('flavor', argParser: argParser);
-    final appId = shorebirdEnv.getShorebirdYaml()!.getAppId(flavor: flavor);
+    final shorebirdYaml = shorebirdEnv.getShorebirdYaml();
+    if (shorebirdYaml == null) {
+      logger.err(
+        '''Unable to find shorebird.yaml. Are you in a shorebird app directory?''',
+      );
+      return ExitCode.unavailable.code;
+    }
```
bryanoltman: We have a `ShorebirdValidator` that does exactly this. You can an example of it at https://github.com/shorebirdtech/shorebird/blob/main/packages/shorebird_cli/lib/src/commands/release/android_releaser.dart#L54-L62.

You probably want something like this instead:

```dart
    try {
      await shorebirdValidator.validatePreconditions(
        checkUserIsAuthenticated: true,
        checkShorebirdInitialized: true,
      );
    } on PreconditionFailedException catch (e) {
      throw ProcessExit(e.exitCode.code);
    }
```

Peetee06: ah nice, thanks for the hint!

I refactored it to use the validator instead 👍

packages/shorebird_cli/test/src/commands/patches/promote_command_test.dart
```diff
@@ -107,13 +111,41 @@ void main() {
...
 
+    group('when validation fails', () {
+      test('exits with exit code from validation error', () async {
+        final exception = ShorebirdNotInitializedException();
+        when(
+          () => shorebirdValidator.validatePreconditions(
+            checkUserIsAuthenticated: any(named: 'checkUserIsAuthenticated'),
+            checkShorebirdInitialized: any(named: 'checkShorebirdInitialized'),
+          ),
+        ).thenThrow(exception);
```
bryanoltman, line 136: This `when` should be in a `setUp` block above this test

bryanoltman (commented):
Small comment, but otherwise looks good!

packages/shorebird_cli/test/src/commands/patches/promote_command_test.dart
Peetee06: updated :)

bryanoltman (approved):

Issue #3302, opened 2025-09-03
fix: flavor flag not respected on Windows


bryanoltman:
App ID: `6e8e4160-8c56-478b-9c94-3f71e75a4e8a`

**Description**

Flutter does not support flavors on Windows. When an app uses Windows in conjunction with a platform that does support flavors (macOS, for example), this results in unexpected behavior. Specifically, given the following shorebird.yaml:

```yaml
app_id: 8ae100a2-100e-42b3-a8b0-b965ebec781d
flavors:
  development: f8b556fe-e0ed-4796-8bf2-7f43f4b93c7f
  production: 6e8e4160-8c56-478b-9c94-3f71e75a4e8a
```

and running `shorebird release windows --flavor production`, the Windows release will always check for patches using the default app ID, ex:

```
2025-09-03T12:12:41.623Z INFO  [updater::network] [shorebird] Sending patch check request: PatchCheckRequest { app_id: "8ae100a2-100e-42b3-a8b0-b965ebec781d", channel: "stable", release_version: "2.0.0+5002", platform: "windows", arch: "x86_64" }
```

bryanoltman:
Closed by https://github.com/shorebirdtech/shorebird/pull/3352. We now print a log message explicitly saying which app id will be used for the patch/release.

PR #3352, opened 2025-10-14
fix(shorebird_cli): warn instead of error when no flavor provided to app that has flavors

### LINKED ISSUE #3248: fix: allow iOS release without --flavor when only Android has flavors
### App ID:

`com.zaper.ai.dev`

### **Description**

We're using Shorebird for Flutter iOS builds. Our project is configured with **flavors for Android only** (e.g., `dev`, `prod`), and **iOS does not use any flavoring**.

When attempting to run the following command for iOS release:

```
shorebird release ios --dart-define=apiMode=live --dart-define=enableLogging=false
```

we receive the following error:

```
The project has flavors (dev, prod), but no --flavor argument was provided
Instance of 'ValidationFailedException'
```

This appears to be a **false-positive validation**, as iOS does not require or use flavors in this project. The same build works perfectly using Flutter’s default command:

```
flutter build ipa --release --dart-define=apiMode=live --dart-define=enableLogging=false
```

### **Steps To Reproduce**

1. Create a Flutter project with Android flavors (`dev`, `prod`) defined.
2. Do **not** configure any iOS flavor targets in Xcode.
3. Run:

   ```bash
   shorebird release ios --dart-define=apiMode=live --dart-define=enableLogging=false
   ```
4. Observe the error.

### **Expected Behavior**

Shorebird should allow iOS builds even if no `--flavor` is provided, as long as **iOS itself has no flavor configuration**.

If the flavors only apply to Android, the iOS build should proceed without requiring a `--flavor` argument.

### **Additional Context**

* `flutter build ipa` works fine in this setup.
* This issue may block CI/CD for teams who rely on platform-specific flavor configurations (Android only).
* Log file: *available on request if needed.*

bryanoltman:
## Description

Some apps have flavors for one platform, but not others. This change updates the behavior of `shorebird release` and `shorebird patch` for apps with flavors when no `--flavor` argument is provided. Instead of an error (current behavior), this will instead print a warning, telling the user which app id will be targeted by the release or patch.

Screenshot 2025-10-14 at 3 24 29 PM

Fixes https://github.com/shorebirdtech/shorebird/issues/3248

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_validator.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_validator.dart
@@ -175,12 +175,23 @@ To fix, update your pubspec.yaml to include the following:
 
       throw ValidationFailedException();
     }
+
+    if (validationIssuesContainsWarning(issues)) {
+      for (final issue in issues) {
+        logger.warn(issue.message);
+      }
+    }
   }
 
   /// Whether any [ValidationIssue]s have a severity of
   /// [ValidationIssueSeverity.error].
   bool validationIssuesContainsError(List<ValidationIssue> issues) =>
       issues.any((issue) => issue.severity == ValidationIssueSeverity.error);
+
+  /// Whether any [ValidationIssue]s have a severity of
+  /// [ValidationIssueSeverity.warning].
+  bool validationIssuesContainsWarning(List<ValidationIssue> issues) =>
+      issues.any((issue) => issue.severity == ValidationIssueSeverity.warning);
 
   /// Logs a message indicating that validation failed. If any of the issues
   /// can be automatically fixed, this also prompts the user to run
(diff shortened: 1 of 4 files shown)


eseidel:
Will this have any effect on https://github.com/shorebirdtech/shorebird/issues/3302 by chance?

eseidel (approved):

bryanoltman:
> Will this have any effect on #3302 by chance?

It shouldn't. For platforms that don't support flavors, we error out before we get to the logic added in this PR. I think that issue probably needs to be reworked or closed.

bryanoltman:
Correction: this makes the windows behavior a bit more obvious

PR #3363, opened 2025-10-28
chore(shorebird_cli): update path to android local engine

bryanoltman:


eseidel (approved):
Wild.  That's building for arm64 from arm64?  Why the double? (that others don't have)  Does this need/want a comment?

bryanoltman:
> Wild. That's building for arm64 from arm64? Why the double? (that others don't have) Does this need/want a comment?

It's a target/host specification, I believe (similar to how we have host_release and host_release_arm64). I've updated it to be a bit smarter about detecting host arch

Issue #3354, opened 2025-10-17
fix: shorebird patch ios consistently crashes


KIRRAA0:
**App ID:** bd50d03b-174a-4b3a-a39a-e1f402e19d18

## Description

`shorebird patch ios` consistently crashes with a segmentation fault during AOT compilation at `dart::kernel::TranslationHelper::DartString()`, while `flutter build ios` and `shorebird patch android` work perfectly with the same codebase. The crash occurs when processing string constants in a large localization file.

## Steps To Reproduce

1. Create an iOS release with Shorebird:
   ```bash
   shorebird release ios --no-confirm
   ```
   ✅ This succeeds

2. Make Dart-only code changes by adding a large localization file containing:
   - 106 new string key-value pairs (English + Arabic)
   - Strings with escaped apostrophes (e.g., `'We\'ll send you...'`)
   - Total ~1200 lines in `Map<String, Map<String, String>>`

3. Try to create a patch:
   ```bash
   shorebird patch ios --release-version=2.2.2+8 --no-confirm
   ```
   ❌ This crashes with segmentation fault

4. Verify the same code builds with standard Flutter:
   ```bash
   flutter clean
   flutter build ios --release --no-codesign
   ```
   ✅ This succeeds

5. Verify Android patching works:
   ```bash
   shorebird patch android --release-version=2.2.2+8 --no-confirm
   ```
   ✅ This succeeds and publishes patch

## Expected Behavior

The iOS patch should compile successfully and create a patch, just like:
- Android patches (which work fine with identical code changes)
- Standard Flutter iOS builds (which compile without errors)
- iOS full releases with `shorebird release ios` (which also work fine)

## Actual Behavior

The command crashes with a segmentation fault:

```
Building App Store IPA...                                          28.3s
✓ Built IPA to build/ios/ipa (40.7MB)

===== CRASH =====
si_signo=Segmentation fault: 11(11), si_code=SEGV_ACCERR(2), si_addr=0xdfe8009c
version=3.9.2 (stable) (Thu Sep 4 15:29:30 2025 -0700) on "macos_simarm64"
pid=53310, thread=259, isolate_group=isolate(0x803018000), isolate=isolate(0x103990290)
os=macos, arch=arm64, comp=no, sim=yes
isolate_instructions=0, vm_instructions=0
fp=16d4c07d0, sp=16d4c0780, pc=102dbfa38

[...]

PR #3403, opened 2025-11-25
fix: Seg fault issue

Dario-Maselli:
## Description
There is SEG Fault issue that occurs when trying to Patch IOS. I do see that the Original Issue is now closed but thought I would still add my **half** solution/bandaid here. 

There is a bit more logging to get down to the issue with some retries. I'll see if I can get some outputs soon to show what the console now looks like.

I eventually got it to move past the error and use the MacOS snapshot to make a patch but it only works for x86 patches, not ARM. I did write tests as I was hoping to fix this properly, but I'm not confident with this solution, so just leaving my thought process here for the interim if this issue persists. 

Making this a draft PR for now with a disclaimer, I have not tried to patch on the latest version (v1.6.68) but the previous version did have this issue still (v1.6.67), so hopefully this can assist with either a temporary solution or a permanent fix.

Issue Link: https://github.com/shorebirdtech/shorebird/issues/3354
Discord Issue Thread: https://discord.com/channels/1030243211995791380/1437710833475129386

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


eseidel:
I have some idea what might be wrong.  I think I can work on this tomorrow.

Dario-Maselli:
Closing this PR as it is not related to the cause of the Seg Fault issue. 

The actual root cause of the issue was in the XCode Project -> Runner -> Build Phases where one of the build scripts was overriding the Flutter Version so the internal shorebird flutter version was not being used for IOS.

^ For those experiencing the same issue, this is a good place to start debugging.

Issue #2788, opened 2025-01-21
shorebird patch ios crash (gen_snapshot version mismatch)


devmfouad:
i build ipa from my app using shorebird release ios
and it build successfully

i created update to my app and run shorebird patch ios

but it crach and here is log content

```
Command: shorebird patch ios --verbose

2025-01-21T13:53:51.932151 [90mNo checksum provided for patch, skipping file corruption validation[0m
2025-01-21T13:53:52.274277 [90mNo checksum provided for aot-tools.dill, skipping file corruption validation[0m
2025-01-21T13:53:52.281898 [90m[HTTP] GET https://api.shorebird.dev/api/v1/apps[0m
2025-01-21T13:53:52.360260 [?7l[2K
[92mâ ™[0m Fetching apps... [90m(82ms)[0m
2025-01-21T13:53:52.440073 [?7l[2K
[92mâ ¹[0m Fetching apps... [90m(0.2s)[0m
2025-01-21T13:53:52.520547 [?7l[2K
[92mâ ¸[0m Fetching apps... [90m(0.2s)[0m
2025-01-21T13:53:52.599692 [?7l[2K
[92mâ ¼[0m Fetching apps... [90m(0.3s)[0m
2025-01-21T13:53:52.680813 [?7l[2K
[92mâ ´[0m Fetching apps... [90m(0.4s)[0m
2025-01-21T13:53:52.759848 [?7l[2K
[92mâ ¦[0m Fetching apps... [90m(0.5s)[0m
2025-01-21T13:53:52.840558 [?7l[2K
[92mâ §[0m Fetching apps... [90m(0.6s)[0m
2025-01-21T13:53:52.883167 [?7h[2K
[92mâœ“[0m Fetching apps [90m(0.6s)[0m

............

===== CRASH =====
si_signo=Segmentation fault: 11(11), si_code=SEGV_MAPERR(1), si_addr=0xb61cd09c
version=3.6.0 (stable) (Thu Dec 12 11:58:55 2024 -0600) on "macos_simarm64"
pid=59607, thread=259, isolate_group=isolate(0x7fa33181f400), isolate=isolate(0x7fa331822200)
os=macos, arch=arm64, comp=no, sim=yes
isolate_instructions=0, vm_instructions=0
fp=3051de150, sp=3051de110, pc=100829db4
[...]

PR #3406, opened 2025-11-26
feat: add doctor check for FLUTTER_* overrides in Xcode project

eseidel:
Adds a doctor/init validator that flags any `FLUTTER_`-prefixed build setting assigned in `ios/Runner.xcodeproj/project.pbxproj`. `FLUTTER_*` variables are set by Flutter itself — either at runtime or via `ios/Flutter/Generated.xcconfig` (which this check does *not* scan) — so hard-coding them in the Xcode project can cause different Flutter versions to be used across build phases.

Believed to address https://github.com/shorebirdtech/shorebird/issues/2788.

Runs during `shorebird doctor`, `shorebird init`, and as part of the iOS command validators (`shorebird release/patch ios`). On a hit, the error message lists the offending names (e.g. `FLUTTER_ROOT`, `FLUTTER_BUILD_DIR`).

### Review follow-ups

- Fixed the regex comment: the `(?!\s*//)` at the start is a negative **lookahead**, not a lookbehind (thanks @bdero).
- Noted that `/* ... */` block comments are intentionally not stripped — a FLUTTER_* assignment inside one still signals intent.
- Added a `validate()` test covering the `getFlutterProjectRoot() == null` branch to get patch coverage to 100%.

### DIFF
--- a/packages/shorebird_cli/lib/src/validators/xcodeproj_flutter_override_validator.dart
+++ b/packages/shorebird_cli/lib/src/validators/xcodeproj_flutter_override_validator.dart
@@ -4,11 +4,13 @@ import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/shorebird_env.dart';
 import 'package:shorebird_cli/src/validators/validators.dart';
 
-/// Checks that ios/Runner.xcodeproj/project.pbxproj does *not* contain
-/// FLUTTER_ROOT environment overrides.  Doing so could cause different
-/// versions of Flutter to be used during different parts of the build process
-/// potentially causing build failures.
-class XcodeprojFlutterRootValidator extends Validator {
+/// Checks that ios/Runner.xcodeproj/project.pbxproj does *not* assign any
+/// `FLUTTER_`-prefixed build setting. Flutter itself populates these at
+/// runtime or via the generated `ios/Flutter/Generated.xcconfig` (which we
+/// do not scan), so a hard-coded assignment in `project.pbxproj` can cause
+/// different versions of Flutter to be used during different parts of the
+/// build, potentially causing build failures.
+class XcodeprojFlutterOverrideValidator extends Validator {
   static final String _iosRunnerXCodeProjPath = p.join(
     'ios',
     'Runner.xcodeproj',
@@ -22,7 +24,7 @@ class XcodeprojFlutterRootValidator extends Validator {
 
   @override
   String get description =>
-      'Xcode project does not override FLUTTER_ROOT environment variable';
+      'Xcode project does not override FLUTTER_ build settings';
 
   @override
   bool canRunInCurrentContext() =>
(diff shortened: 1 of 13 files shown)


eseidel:
This still needs more tests to land.

eseidel:
I'm not sure why we didn't get coverage info.  🤔  https://app.codecov.io/gh/shorebirdtech/shorebird/pull/3406

bryanoltman (approved):

eseidel:
This is one of multiple things that could cause https://github.com/shorebirdtech/shorebird/issues/2788 I believe.  I'm still not sure we should ship this particular fix (since trying to replicate the direct cause I got a different error), but maybe we should broaden this fix to detect any FLUTTER_* or a known set of FLUTTER_ environment variables set in the pbxproj directly, since all of those cases would be wrong.

One of the gen_snapshot crashes I helped with last week was overriding the FLUTTER_ROOT within the xcode_backend.sh build script and just happened to be working locally but was (understandably) then failing with shorebird.

I wish I had a more systemic fix (like detecting somehow when the backend scripts arent matching), but I don't know how to write that fix.  Maybe opening a side socket in a known location and having the backend look for that socket and confirm versions or sometihng 🤷

packages/shorebird_cli/lib/src/validators/xcodeproj_flutter_override_validator.dart
```diff
@@ -0,0 +1,104 @@
...
+  Set<String> _flutterOverridesIn(File pbxProjFile) {
+    // We could consider using package:xcode_parser, but for now this should
+    // be sufficient.
+    final contents = pbxProjFile.readAsStringSync();
+    // Match FLUTTER_<NAME> assignments in various formats:
+    // - FLUTTER_FOO = value;
+    // - FLUTTER_FOO = "value";
+    // - FLUTTER_FOO = $(FLUTTER_FOO);
+    // Allow flexible spacing around the equals sign.
+    // Use a negative lookbehind on `//` to avoid commented-out lines, and
```
bdero:
```suggestion
    // Use a negative lookahead on `//` to avoid commented-out lines, and
```

packages/shorebird_cli/lib/src/validators/xcodeproj_flutter_override_validator.dart
```diff
@@ -0,0 +1,104 @@
...
+    // require the assignment to terminate with `;` on the same line.
+    final matcher = RegExp(
+      r'^(?!\s*//).*?(FLUTTER_\w+)\s*=\s*[^;\n]+;',
```
bdero, line 100: pbxproj files support block comments too, apparently. So something like `/* FLUTTER_ROOT = /old/path; */` won't be ignored. Probably fine though.

bdero (approved):

PR #3452, opened 2025-12-31
fix: remove confirmation in shorebird release

### LINKED ISSUE #3439: fix: remove y/N prompt for `shorebird release`
It doesn't really serve a purpose anymore and is just a foot-gun.
```
Would you like to continue? (y/N) Yes
```

It's too easy to hit enter and have it stop the whole release and have to start over again.

eseidel:
I'm not sure why we added this originally, but currently it feels more like a footgun than a help.

A smaller change would be to just change the default to "Y" instead of "N", but my gut says we can just remove this entirely and no one will complain.

Fixes https://github.com/shorebirdtech/shorebird/issues/3439.

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -186,9 +186,6 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
   /// Whether to allow changes in native code (--allow-native-diffs).
   bool get allowNativeDiffs => results['allow-native-diffs'] == true;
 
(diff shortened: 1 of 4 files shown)


eseidel:
I don't feel super strongly here?  But I also feel like I've accidentally canceled a release much more often than this prompt has helped.

bryanoltman (changes requested):
This should also remove the `--no-confirm` option from our args list

bryanoltman:
We should probably also remove this from the patch command

bryanoltman:
Created https://github.com/shorebirdtech/shorebird/pull/3454 with these changes. Feel free to merge into this branch if they look OK.

bryanoltman (approved):

PR #3454, opened 2025-12-31
fix(shorebird_cli): remove `--no-confirm` flag from patch and release commands

bryanoltman:
## Description

Removes the `--no-confirm` option from the patch and release commands

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -103,6 +103,11 @@ To target the latest release (e.g. the release that was most recently updated) u
 [DEPRECATED] Whether to publish the patch to the staging environment. Use --track=staging instead.''',
         hide: true,
       )
+      ..addFlag(
+        CommonArguments.noConfirmArg.name,
+        help: CommonArguments.noConfirmArg.description,
+        negatable: false,
+      )
       ..addOption(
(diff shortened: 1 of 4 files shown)


eseidel:
I worried removing the arg would break existing scripts.  I could have marked it deprecated I guess.

I also think we need to keep the confirmation on patch for now, since what it's really confirming is "promote to stable and deploy to your users".

eseidel (changes requested):
I don't think this will be safe to ship as-is.  Both because it will break scripts that have --no-confirm and because it will cause `shorebird patch` to instantly ship (which maybe is what the user wants anyway and should be solved differently?)

bryanoltman:
> I worried removing the arg would break existing scripts. I could have marked it deprecated I guess.

I don't think we need to worry about this breaking scripts, as we don't ask for confirmation on CI (which, I imagine, is something like 99% of scripted invocations of the command).

> I also think we need to keep the confirmation on patch for now, since what it's really confirming is "promote to stable and deploy to your users".

I feel like the y/N confirm dialog is a footgun any time we're building, but don't feel strongly about removing it from patch.

eseidel (approved):
OK, I'm convinced.

Leave the --no-confirm flag to avoid breaking scripts.

PR #3465, opened 2026-01-13
feat: track git commit hash for releases and patches (#3443)

### LINKED ISSUE #3443: feat: would like a way to know what git commit a release was made with
This may just be https://github.com/shorebirdtech/shorebird/issues/1288

But I would like to be able to go back to a release and know what git hash it came from (optionally, since not everyone uses git).

realmeylisdev:
## Summary
Add optional tracking of the project's git commit hash when creating releases and patches. This allows users to trace a release or patch back to the exact source code commit it was built from, enabling better debugging and reproducibility.

## Changes
- Add `projectGitHash` field to `BuildEnvironmentMetadata`
- Update release command to capture git hash
- Update patch command to capture git hash
- Add comprehensive tests for new functionality

## Implementation Details
The git hash is captured on a best-effort basis:
- Returns the current HEAD commit hash if the project is in a git repo
- Returns `null` if not in a git repo or git is unavailable
- Release/patch continues successfully even if git hash capture fails
- Field is optional and backward compatible

## Testing
✅ All tests passing (147+ tests)
- Metadata tests: 17/17
- Release command tests: 23/23
- Patch command tests: 45/45
- iOS releaser/patcher tests: 62/62

Fixes #3443

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -12,6 +12,7 @@ import 'package:shorebird_cli/src/commands/patch/patch.dart';
 import 'package:shorebird_cli/src/common_arguments.dart';
 import 'package:shorebird_cli/src/config/config.dart';
 import 'package:shorebird_cli/src/deployment_track.dart';
+import 'package:shorebird_cli/src/executables/git.dart';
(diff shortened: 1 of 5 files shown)


eseidel:
We should do this.  But I don't think it can be done without also modifying our backend.  So I believe @easymac will be tackling this as part of the "correlation keys" work he's landing this week.

PR #3466, opened 2026-01-14
fix(shorebird_cli): use correct local-engine out dirs for Android on macOS

bdero:
## Description
Removes some special sauce originally added in https://github.com/shorebirdtech/shorebird/pull/3363.

Flutter's GN script takes an `--android-cpu {arm,x64,x86,arm64,riscv64}` param which [determines the suffix](https://github.com/flutter/flutter/blob/9aea438bfcb9b8041106650c847d29d49ffbf120/engine/src/flutter/tools/gn#L38-L39) of the build output directory. However, if you look at the surrounding code, you'll notice it'll blindly append more stuff if you accidentally include multiple CPU parameters, which I assume is what caused the confusion here. :)

So if your GN args are `--android --android-cpu=arm64 --mac-cpu=arm64 --simulator-cpu=arm64 --linux-cpu=x86 --windows-cpu=x86 --runtime-mode=release`, then you'll get a pretty hilarious output directory: `android_release_arm64_arm64_arm64_x86_x86`
We should fix this in the engine. 🤣 

This patch also adds a clear error message when no matching build architectures are available.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:collection/collection.dart';
 import 'package:mason_logger/mason_logger.dart';
 import 'package:shorebird_cli/src/artifact_builder/artifact_builder.dart';
 import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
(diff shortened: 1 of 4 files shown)


eseidel (approved):
lgtm!

Issue #1722, opened 2024-02-09
feat: issue our own CI tokens (don't use oauth tokens)


eseidel:
Moving to our own tokens would make it so we can:
1. revoke them (and give them lifetimes, etc.)
2. vend them from console.shorebird.dev

Right now our CI instructions ask you to run `shorebird login:ci` to get the auth token.

https://docs.shorebird.dev/ci/github

We could also expose that from the console and avoid users having to install `shorebird` at all?

lambdachad:
Would be awesome to have this. Makes it possible to use `shorebird` with Github Actions without setting up all the tooling.

eseidel:
Erick wrote this and landed it but we since reverted it.  We currently use two separate "audiences" for our oauth tokens (one for the console and one for the cli).  Getting the console to be able to vend a "cli" auth token would be more complexity than its worth.  So the plan is to wait until we move to our own CI token system (separate from oauth) to fix this.

eseidel:
Thinking about this more, I think this is really about us taking ownership of our own session management.  Right now our login is just that we store the JWT identity token as our session value and thus are at the mercy of the expiry of those tokens from Google/Microsoft for our web sessions.  We store the JWT refresh token on the CLI and thus those logins stay forever. Neither of those is what we'll want to end up with, rather we move to using our own explicit sessions and then we can also handle "service accounts" (which is essentially what CI tokens should be) separately, and it also sets us up for handling SAML which right now we can't easily since we don't have our own session to create after the SAML attestation is done.

eseidel:
I committed to a customer this afternoon that we intend to land this within the next two weeks.  Most of the patches are written already, but will be some work to land it.  Work may bleed past the next two weeks since we're currently busy rolling out another feature, but I expect we should have the bulk of this live by then.

bdero:
Shipped in 1.6.91 as API keys: https://docs.shorebird.dev/account/api-keys/. `login:ci` is deprecated. Closing.

PR #3483, opened 2026-01-29
feat: migrate CLI to auth.shorebird.dev

eseidel:
Part of https://github.com/shorebirdtech/shorebird/issues/1722.

This removes the --provider option from `shorebird login` instead asking the user to pick at auth.shorebird.dev.  We *could* keep the --provider option and just pass through to auth.shorebird.dev but I'm choosing the simpler UX for now.

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -338,49 +338,18 @@ class Auth {
     }
   }
 
-  /// Logs in the user.
-  Future<void> login(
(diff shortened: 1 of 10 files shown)


eseidel:
This is waiting on the deployment of auth.shorebird.dev to land, which is itself just blocked on our engineers finishing other projects.  I expect we'll land this in another 2 weeks or so.

eseidel:
I think you landed this otehr ways and we can close this @easymac ?

easymac:
Yes, this is live already 👍

PR #3492, opened 2026-01-31
fix: select correct executable in Windows preview

### LINKED ISSUE #3450: fix: previewing patches on Windows isn't possible
App ID: d618e438-c233-4fc5-b8bd-9fafb43e1a9c

**Description**

When trying to preview an app on Windows, this error appears. Apologies for the screenshot and no text.

Image

It seems the crashpad_handler.exe which is launched first doesn't get launched properly.

**Steps To Reproduce**

1. Make a release for Windows
2. Try to preview that release (maybe this just occurs for non-stable tracks?)

**Expected Behavior**

It should preview.

**Additional Context**

In combination with #3247, this seems to make it impossible to preview a patch when developing only with Windows and Android without pushing the patch straight to stable.

eseidel:
## Summary
- Fixes #3450
- The preview command used a naive "first `.exe` found" approach which could pick `crashpad_handler.exe` instead of the actual app executable
- Now uses `Windows.findExecutable()` (already used by release/patch commands) which matches by project name first, then falls back to most recently modified exe
- Filters out known non-app executables like `crashpad_handler.exe`
- Makes `projectName` optional and `ignoredExecutables` overridable on `findExecutable()`

## Test plan
- [x] Existing preview command tests pass
- [x] Existing Windows platform tests pass
- [x] Analyzer passes with no warnings


eseidel:
We shouldn't commit this until we have a chance to smoke-test this on windows.

easymac (approved):

PR #3626, opened 2026-02-24
feat(shorebird_cli): make --obfuscate a first-class CLI flag

### LINKED ISSUE #1619: feat: support --obfuscate on iOS
**Description**

Running `shorebird patch ios-alpha` on an app that uses flavors fails with the following message 

```
Failed to link AOT files: Exception: Failed to link: base and patch snapshots have differing VM sections
```

I created a repo and manually added flavors/schemes to iOS (development and production). The repo I used to reproduce the issue can be found [here](https://github.com/psatler/shorebird_flavors). 
I was also able to reproduce this issue using the [very good CLI](https://pub.dev/packages/very_good_cli) which automatically sets up an app with flavors from the get-go when running `very_good create flutter_app <app-name>`

Android `release` and `patch` are working fine.

**Steps To Reproduce**

On a project that uses flavors/schemes

1. Trigger a release for a given flavor, let's say `development`:
```bash
shorebird release ios-alpha \
        --target lib/main_development.dart \
        --flavor development -- \
        --obfuscate --split-debug-info=./build/app/outputs/symbols
```
2. After that release is published, trigger a patch release this time

```bash
shorebird patch ios-alpha \
        --target lib/main_development.dart \
        --flavor development -- \
        --obfuscate --split-debug-info=./build/app/outputs/symbols
```
3. Verify that the patch fails 

Again, a reproducible repo was created [here](https://github.com/psatler/shorebird_flavors). 

**Expected Behavior**

To be able to publish a patch release for the respective flavor/target.

**Screenshots**

- initializing shorebird
image

- triggering a release
image

- triggering a patch
image

**Additional Context**

- `shorebird doctor`: 
```bash
Shorebird 0.21.1 • git@github.com:shorebirdtech/shorebird.git
Flutter 3.16.5 • revision 4e8a7c746ae6f10951f3e676f10b82b21d7300a5
Engine • revision 001b60083b642c134c62e7c10bc34f188b0a2869

✓ Shorebird is up-to-date (0.6s)
✓ Flutter install is correct (0.5s)
✓ AndroidManifest.xml files contain INTERNET permission (35ms)
✓ Has access to storage.googleapis.com (0.2s)

No issues detected!
```
- `flutter doctor`
```bash
❯ flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.16.5, on macOS 14.2.1 23C71 darwin-arm64, locale en-BR)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
[✓] Xcode - develop for iOS and macOS (Xcode 15.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.3)
[✓] VS Code (version 1.85.1)
[✓] Connected device (4 available)
[✓] Network resources

• No issues found!
```

bdero:
## Summary

- Adds `--obfuscate` as a first-class flag to both `release` and `patch` commands across all platforms
- When `--obfuscate` is used without `--split-debug-info`, automatically defaults to `build/shorebird/symbols/`
- Users can now run `shorebird release ios --obfuscate` instead of `shorebird release ios -- --obfuscate --split-debug-info=symbols/`

Closes https://github.com/shorebirdtech/shorebird/issues/1619

## Dependencies

This PR depends on the corresponding dart-sdk changes to land and the engine artifacts to be bumped:
- https://github.com/shorebirdtech/dart-sdk/pull/757

## Test plan

- [x] Bump engine artifacts after dart-sdk PR merges
- [x] `shorebird release android --obfuscate` creates a release successfully
- [x] `shorebird patch android` creates a patch successfully against an obfuscated release
- [x] `shorebird release ios --obfuscate` creates a release successfully
- [x] `shorebird patch ios` creates a patch successfully against an obfuscated release
- [x] Verify user-provided `--split-debug-info` is preserved when both flags are used
- [x] Run unit tests for affected files


packages/shorebird_cli/lib/src/commands/patch/aar_patcher.dart
```diff
@@ -74,10 +74,20 @@ class AarPatcher extends Patcher {
   );
 
   @override
-  Future<File> buildPatchArtifact({String? releaseVersion}) async {
+  Future<File> buildPatchArtifact({
+    String? releaseVersion,
+    String? obfuscationMapPath,
```
eseidel: This feels wrong?  This should come from the db record?

bdero: so this path is local. We upload the obfuscation map to the release artifacts and then check for the presence of the map via `maybeGetReleaseArtifact(arch: "ios_obfuscation_map", ...)` at patch build time. THen we can grab it and pass the local path of the downloaded map here

eseidel: I guess?  It just seems odd that this one is singled out this way.  I guess we don't want to doctor the forwarded args and this is our first time where we're re-interpreted a forwarded arg into a different form?

bdero:
Smoke tested this end-to-end and it's working! Dart-side changes are merged and a new version of the engine artifacts is in progress right now for the version bump. We gotta bump the flutter version before merging this, of course.

packages/shorebird_cli/lib/src/commands/patch/aar_patcher.dart
bdero: I have no qualms about hacking up the forwarded args in patch_command. 🤷‍♂️  Made some changes to reduce the code bloat. Please take a look.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -537,6 +538,30 @@ aab artifact already exists, continuing...''');
...
     }
 
+    if (obfuscationMapPath != null) {
+      final obfuscationMapFile = File(obfuscationMapPath);
+      try {
+        await codePushClient.createReleaseArtifact(
+          appId: appId,
+          releaseId: releaseId,
+          artifactPath: obfuscationMapFile.path,
+          arch: 'android_obfuscation_map',
```
eseidel: Is this supposed to be android_?  This is generic code, no?

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -732,6 +758,30 @@ aar artifact already exists, continuing...''');
...
+          podfileLockHash: null,
+        );
+      } catch (error) {
+        _handleErrorAndExit(
+          error,
+          progress: createArtifactProgress,
+          message: 'Error uploading obfuscation map: $error',
+        );
+      }
+    }
```
eseidel: This seems repeated with the above?  should we share code?

packages/shorebird_cli/lib/src/commands/release/macos_releaser.dart
```diff
@@ -51,17 +62,6 @@ To change the version of this release, change your app's version in your pubspec
...
-    if (argResults.rest.contains('--obfuscate')) {
-      // Obfuscated releases break patching, so we don't support them.
-      // See https://github.com/shorebirdtech/shorebird/issues/1619
-      logger
-        ..err('Shorebird does not currently support obfuscation on macOS.')
-        ..info(
-          '''We hope to support obfuscation in the future. We are tracking this work at ${link(uri: Uri.parse('https://github.com/shorebirdtech/shorebird/issues/1619'))}.''',
-        );
-      throw ProcessExit(ExitCode.unavailable.code);
-    }
```
eseidel, line 64: I think we need to keep this for old versions of Flutter?  Since we don't support --obfuscate on those w/o your changes, which are only in 3.41.2 and later?

eseidel (approved):
I think you will want to:
- move to using the ios_supplement pattern, to avoid making more architectures.
- revive the warning/error for --obfuscate for older versions of flutter.
But otherwise I think you're good to ship.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bdero: Fixed

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
bdero: Done.

packages/shorebird_cli/lib/src/commands/release/macos_releaser.dart
bdero: Done, w/ guard for 3.41.2

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -152,6 +153,11 @@ To target the latest release (e.g. the release that was most recently updated) u
         CommonArguments.splitDebugInfoArg.name,
         help: CommonArguments.splitDebugInfoArg.description,
       )
+      ..addFlag(
+        CommonArguments.obfuscateArg.name,
+        help: CommonArguments.obfuscateArg.description,
+        negatable: false,
+      )
```
eseidel, line 162: So this is just entirely for CI compatibility?  Should this be a hidden flag since it just warns if you pass it?

packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
```diff
@@ -164,6 +177,8 @@ Please comment and upvote ${link(uri: Uri.parse('https://github.com/shorebirdtec
       architectures: architectures,
       flavor: flavor,
     );
+
+    await uploadSupplementArtifact(appId: appId, releaseId: release.id);
```
eseidel, line 181: Could we ever get interrupted between these two and have the db in a confused state?

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -117,13 +129,14 @@ class IosFrameworkReleaser extends Releaser {
...
-    return codePushClientWrapper.createIosFrameworkReleaseArtifacts(
+  }) async {
+    await codePushClientWrapper.createIosFrameworkReleaseArtifacts(
       appId: appId,
       releaseId: release.id,
       appFrameworkPath: p.join(releaseDirectory.path, 'App.xcframework'),
-      supplementPath: artifactManager.getIosReleaseSupplementDirectory()?.path,
     );
+
+    await uploadSupplementArtifact(appId: appId, releaseId: release.id);
```
eseidel, line 139: Future code-sharing between platform impls I guess.  There is probably a lot of similar in shorebird_cli.

packages/shorebird_cli/lib/src/commands/release/releaser.dart
```diff
@@ -1,16 +1,26 @@
...
 import 'package:shorebird_cli/src/shorebird_env.dart';
+import 'package:shorebird_cli/src/shorebird_flutter.dart';
+import 'package:shorebird_cli/src/third_party/flutter_tools/lib/flutter_tools.dart';
 import 'package:shorebird_code_push_client/shorebird_code_push_client.dart';
 import 'package:shorebird_code_push_protocol/shorebird_code_push_protocol.dart';
 
 export 'package:pub_semver/pub_semver.dart';
 
+/// Minimum Flutter version for obfuscation support across all platforms.
+final _minimumObfuscationFlutterVersion = Version(3, 41, 2);
```
eseidel, line 22: I wonder if we eventually move these to a central list or something. I wouldn't be able to tell you where all our various version constraints are today.

We will also update https://docs.shorebird.dev/getting-started/flutter-version/ when this launches.

packages/shorebird_cli/lib/src/commands/release/releaser.dart
```diff
@@ -87,4 +97,117 @@ abstract class Releaser {
...
+  /// mapping and --strip to remove unobfuscated DWARF debugging information
+  /// from the compiled snapshot (the DWARF sections would otherwise leak
+  /// identifiers that obfuscation was meant to hide).
+  void addObfuscationMapArgs(List<String> buildArgs) {
+    if (!useObfuscation) return;
+    final mapDir = Directory(p.dirname(obfuscationMapPath));
+    if (!mapDir.existsSync()) mapDir.createSync(recursive: true);
+    buildArgs.addAll([
+      '--extra-gen-snapshot-options=--save-obfuscation-map=$obfuscationMapPath',
+      '--extra-gen-snapshot-options=--strip',
```
eseidel, line 134: I feel like there was another place where these were added (with nice comments!) above?

packages/shorebird_cli/lib/src/artifact_manager.dart
```diff
@@ -368,44 +368,49 @@ class ArtifactManager {
...
+  /// Returns the shorebird release supplement directory for the given
+  /// [platformSubdir] (e.g. 'ios', 'macos', 'android').
   ///
-  /// Returns null if there is no supplement directory
-  /// (e.g. when using older Flutter revisions).
-  Directory? getIosReleaseSupplementDirectory() {
-    final projectRoot = shorebirdEnv.getShorebirdProjectRoot()!;
+  /// If [create] is true, the directory is created if it doesn't exist.
+  /// Returns null if the directory doesn't exist and [create] is false.
+  Directory? getReleaseSupplementDirectory({
```
eseidel, line 376: Nice cleanup!

packages/shorebird_cli/lib/src/common_arguments.dart
```diff
@@ -137,6 +137,16 @@ Command that reads data from stdin and outputs a base64 signature to stdout.
...
   );
 
+  /// The Flutter --obfuscate argument.
+  static const obfuscateArg = ArgumentDescriber(
+    name: 'obfuscate',
+    description:
+        'In a release build, this flag removes identifiers and replaces '
+        'them with randomized values for the purposes of source code '
+        'obfuscation. This flag must always be combined with '
+        '"--split-debug-info" command, to not break crash reports.',
```
eseidel, line 147: I think over time we just simply this, and imply the latter with the former.  Also when we launch our own crash reporting we'll just have it "do the right thing".

eseidel (approved):
lgtm!  Exciting!

misterGFTT:
Guys, when will you publish this new version? I can't wait to use it. 😂

packages/shorebird_cli/lib/src/common_arguments.dart
bdero: Agreed

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
bdero: Yup, let's hide it. Done.

packages/shorebird_cli/lib/src/commands/release/android_releaser.dart
bdero: Yeah, I'd normally say it's a pre-existing concern with the supplement artifact. But it's the first time we're actually using it makes a consequential decision; whether patches get obfuscated or not... so we should probably figure out a way to make this better and maybe automatically invalidate the release when this kind of thing happens. Added a comment and filed a bug: https://github.com/shorebirdtech/shorebird/issues/3630

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
bdero: Yeah, I tried to reduce some of the repeated logic in this PR, but I think there's more we can do. File a bug: https://github.com/shorebirdtech/shorebird/issues/3631

packages/shorebird_cli/lib/src/commands/release/releaser.dart
bdero: Yep, we gotta do this for both patches and releases.

packages/shorebird_cli/lib/src/commands/release/releaser.dart
bdero: Yup, that's a super good comment and I agree we should put all our version constraints into a centralized constants list somewhere. Did a little Claude sleuthing and filed a bug here so we don't forget: https://github.com/shorebirdtech/shorebird/issues/3632

bdero:
@misterGFTT this is released now in Shorebird 1.6.84 for Flutter 3.41.2 builds!

PR #3628, opened 2026-02-27
feat: add --confirm flag to release and patch commands

eseidel:
## Summary
- Adds a hidden `--confirm` flag to `shorebird release` and `shorebird patch` that restores the interactive confirmation prompt when explicitly passed
- Default behavior (no confirmation) is unchanged; `--no-confirm` preserved for backwards compatibility
- When `--confirm` is passed and the terminal supports input, prompts "Would you like to continue?" before proceeding

## Test plan
- [x] `dart test packages/shorebird_cli/test/src/commands/patch/patch_command_test.dart` — all 48 tests pass
- [x] `dart test packages/shorebird_cli/test/src/commands/release/release_command_test.dart` — all 27 tests pass
- [x] Tests cover: `--confirm` + user confirms (continues), `--confirm` + user declines (aborts), no `--confirm` (no prompt)

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -108,6 +108,11 @@ To target the latest release (e.g. the release that was most recently updated) u
         help: CommonArguments.noConfirmArg.description,
         negatable: false,
       )
+      ..addFlag(
+        'confirm',
+        negatable: false,
+        hide: true,
+      )
       ..addOption(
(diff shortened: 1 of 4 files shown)


bdero (approved):
Looks fine to me, but instead of these controlling separate flags, maybe we should have --confirm and --no-confirm control a single flag and only allow one to be present at a time (I think that's what making --confirm negatable would do)? I think right now both of them could be passed in, which is kind of confusing.

eseidel:
Oops, yeah, they should control the same falg, fixing.

PR #3629, opened 2026-02-27
refactor: use single negatable --confirm flag instead of separate flags

eseidel:
## Summary
- Replaces the separate `--no-confirm` (non-negatable) and `--confirm` (non-negatable) flags with a single negatable `--confirm` flag
- `--no-confirm` continues to work as the negated form of `--confirm`, preserving backwards compatibility
- Follows up on #3628

## Test plan
- [x] `dart test packages/shorebird_cli/test/src/commands/patch/patch_command_test.dart` — all 48 tests pass
- [x] `dart test packages/shorebird_cli/test/src/commands/release/release_command_test.dart` — all 27 tests pass


bdero (approved):
lgtm

PR #3638, opened 2026-03-04
feat: Support new Auth service in CLI

easymac:
## Description

- Add support for Shorebird's own auth service as a first-party login provider alongside Google and Microsoft
- Implement a loopback OAuth flow `shorebird_oauth.dart` that binds a local HTTP server, directs the user to the auth service login page, receives an auth code callback, and exchanges it for JWT access/refresh tokens
- Make Shorebird the default auth provider for both `login` and `login_ci` commands (previously defaulted to Google with an interactive prompt)
- Add `ShorebirdAuthEndpoints`, `authServiceUri`, and `jwtIssuer` to `ShorebirdEnv` (configurable via `AUTH_SERVICE_URL` and `SHOREBIRD_JWT_ISSUER` environment variables, defaulting to `https://auth.shorebird.dev`)
- Route token refresh through the Shorebird auth service's `/token` endpoint for Shorebird-issued credentials, preserving existing Google/Microsoft refresh behavior
- Remove interactive provider selection prompt — provider now defaults to Shorebird and can be overridden with --provider

image

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests


packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
```diff
@@ -0,0 +1,224 @@
...
+  // Validate the access token is a well-formed JWT.
+  final Jwt jwt;
+  try {
+    jwt = Jwt.parse(accessTokenValue);
+  } on FormatException catch (e) {
+    throw ShorebirdAuthException('Invalid access token: ${e.message}');
+  }
+
+  // Validate the issuer matches the expected auth service.
+  final issuer = expectedIssuer.toString();
```
bdero, line 233: The issuer comparison uses expectedIssuer.toString() which may include a trailing slash depending on how the URI was constructed (e.g. https://auth.shorebird.dev vs https://auth.shorebird.dev/). If the JWT issuer doesn't have a trailing slash but the authBaseUrl does (or vice versa), this will fail. Consider normalizing both sides or comparing the Uri objects after stripping trailing slashes.

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
```diff
@@ -0,0 +1,224 @@
...
+/// 4. Waits for the auth service to redirect back with an auth code.
+/// 5. Exchanges the auth code for tokens via the auth service's /token endpoint.
+/// 6. Returns the tokens as [oauth2.AccessCredentials].
+Future<oauth2.AccessCredentials> obtainShorebirdCredentials({
+  required http.Client httpClient,
+  required Uri authBaseUrl,
+  required void Function(String) userPrompt,
+  Duration timeout = const Duration(minutes: 5),
+}) async {
+  final server = await HttpServer.bind(InternetAddress.loopbackIPv4, 0);
```
bdero: The server only binds to InternetAddress.loopbackIPv4. On systems where localhost resolves to ::1 (IPv6 only), the callback would fail. Consider binding to InternetAddress.anyIPv6 or trying IPv6 as a fallback, though this may be an acceptable limitation for now.,

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
```diff
@@ -0,0 +1,226 @@
...
+  // Validate the issuer matches the expected auth service.
+  final issuer = expectedIssuer.toString();
+  if (jwt.payload.iss != issuer) {
+    throw ShorebirdAuthException(
+      'Token issuer mismatch: expected $issuer, '
+      'got ${jwt.payload.iss}',
+    );
+  }
+
+  final expiry = DateTime.now().add(Duration(seconds: expiresIn)).toUtc();
```
bdero: DateTime.now() makes this harder to test deterministically. The existing codebase may already do this, but worth noting — a clock injection would make expiry testing more precise.

packages/shorebird_cli/lib/src/shorebird_env.dart
```diff
@@ -228,6 +228,20 @@ class ShorebirdEnv {
...
+  Uri get authServiceUri => Uri.parse(
+    platform.environment['AUTH_SERVICE_URL'] ?? 'https://auth.shorebird.dev',
+  );
+
+  /// The expected JWT issuer for Shorebird-issued tokens. Can be overridden
+  /// with the `SHOREBIRD_JWT_ISSUER` environment variable. Defaults to
+  /// `https://auth.shorebird.dev`.
+  String get jwtIssuer =>
+      platform.environment['SHOREBIRD_JWT_ISSUER'] ??
+      'https://auth.shorebird.dev';
```
bdero, line 243: authServiceUri and jwtIssuer both default to https://auth.shorebird.dev but are independently overridable via different env vars (AUTH_SERVICE_URL vs SHOREBIRD_JWT_ISSUER). This means they could diverge accidentally. Consider whether jwtIssuer should default to authServiceUri.toString() instead of a hardcoded value, so they stay in sync unless explicitly overridden.

bdero (commented):
Nice. Gave this a quick onceover and looks good! My review bot had a few suggestions that look reasonable.

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
bdero: (This is definitely more of a nit. But seems easy to inject a datetime in this case for covering the logic deterministically.)

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
easymac: We're matching existing behavior right now, but binding a loopbackIPv6 in addition to the IPv4 is trivial, and `localhost` will *almost* certainly resolve to both or the IPv6 address in the case of an IPv6 fallback. Were it a more substantial change, I think my instinct would be to not modify existing behavior here.

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
easymac: I did fall for this in our dev environment. Wasn't sure if I wanted to add a check for it since issuers aren't technically required by spec to be URIs, and hypothetically an issuer string could just end with a trailing slash, but this seems better.

packages/shorebird_cli/lib/src/shorebird_env.dart
easymac: I think I'm okay with this. In the unlikely but possible scenario that we move our login form (e.g., from auth.shorebird.dev to login.shorebird.dev or something), we would be forced to change the issuer to match and that would invalidate all existing tokens. The mismatch is the more realistic failure scenario, but forcing alignment would need to be done in all apps that require auth, and undone if we change our minds.

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -298,6 +337,28 @@ class Auth {
...
+    required http.Client client,
+    required void Function(String) prompt,
+  }) async {
+    if (authProvider == AuthProvider.shorebird) {
+      return _obtainShorebirdCredentials(
+        httpClient: client,
+        authBaseUrl: _authServiceUri,
+        userPrompt: prompt,
+      );
+    } else {
```
eseidel: I don't typically do elses after returns, but I'm not sure what (if any) official style we've written down.

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -413,7 +474,12 @@ class UserNotFoundException implements Exception {
 extension OauthAuthProvider on Jwt {
   /// Get the [AuthProvider] from the JWT issuer.
   AuthProvider get authProvider {
-    if (payload.iss == googleJwtIssuer) {
+    // Strip trailing slashes so "https://auth.shorebird.dev/" and
+    // "https://auth.shorebird.dev" compare as equal.
+    final iss = payload.iss.replaceAll(RegExp(r'/+$'), '');
+    if (iss == shorebirdEnv.jwtIssuer.replaceAll(RegExp(r'/+$'), '')) {
```
eseidel: Why is this only needed for Shorebird?  Is one of our servers configured wrong?  Did we set the global in this CLI wrong?  This seems suspect.

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -478,8 +540,7 @@ extension OauthValues on AuthProvider {
       // Required to get refresh tokens.
       'offline_access',
     ],
-    (AuthProvider.shorebird) => throw UnsupportedError(
-      'Shorebird auth is not yet supported in the CLI',
-    ),
+    // Shorebird auth doesn't use scopes.
+    (AuthProvider.shorebird) => [],
```
eseidel, line 543: I presume this is a "yet", I would expect we will over time?

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
```diff
@@ -0,0 +1,234 @@
...
+Future<oauth2.AccessCredentials> obtainShorebirdCredentials({
+  required http.Client httpClient,
+  required Uri authBaseUrl,
+  required void Function(String) userPrompt,
+  Duration timeout = const Duration(minutes: 5),
+}) async {
+  final server = await HttpServer.bind(
+    InternetAddress.loopbackIPv4,
+    0,
+  ).catchError((_) => HttpServer.bind(InternetAddress.loopbackIPv6, 0));
```
eseidel: Odd that this method is nmaed "obtain Shorebird crededitals" but really is "run local oauth calback server".  I wonder if some of this should be split out?

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
```diff
@@ -0,0 +1,234 @@
...
+  final tokenUrl = authBaseUrl.replace(
+    path: p.url.join(authBaseUrl.path, 'token'),
+  );
+
+  final response = await httpClient.post(
+    tokenUrl,
+    headers: {'Content-Type': 'application/x-www-form-urlencoded'},
+    body:
+        'grant_type=refresh_token'
+        '&refresh_token=${Uri.encodeComponent(refreshToken)}',
```
eseidel: It looks like this is parameter encoded.  I suspetd there is more darty way to do this.

packages/shorebird_cli/lib/src/auth/shorebird_oauth.dart
```diff
@@ -0,0 +1,234 @@
...
+  } on FormatException catch (e) {
+    throw ShorebirdAuthException('Invalid access token: ${e.message}');
+  }
+
+  // Validate the issuer matches the expected auth service.
+  // Strip trailing slashes to normalize URIs for comparison — e.g.
+  // "https://auth.shorebird.dev/" and "https://auth.shorebird.dev" should
+  // be treated as equivalent.
+  final issuer = expectedIssuer.toString().replaceAll(RegExp(r'/+$'), '');
+  final actualIssuer = jwt.payload.iss.replaceAll(RegExp(r'/+$'), '');
```
eseidel: This feels wrong again.  We should fix whatever constant we have wrong.  If this is really needed it needs to be a shared method.

packages/shorebird_cli/lib/src/commands/login_ci_command.dart
```diff
@@ -16,8 +16,8 @@ class LoginCiCommand extends ShorebirdCommand {
       'provider',
       abbr: 'p',
       allowed: api.AuthProvider.values.map((e) => e.name),
-      defaultsTo: api.AuthProvider.google.name,
-      help: 'The authentication provider to use. Defaults to Google.',
+      defaultsTo: api.AuthProvider.shorebird.name,
+      help: 'The authentication provider to use. Defaults to Shorebird.',
```
eseidel: I don't think this is right?  I think this should maybe control which provider we default to when connecting to shorebird auth?

packages/shorebird_cli/lib/src/commands/login_command.dart
```diff
@@ -41,11 +41,7 @@ class LoginCommand extends ShorebirdCommand {
     if (results.wasParsed('provider')) {
       provider = api.AuthProvider.values.byName(results['provider'] as String);
     } else {
-      provider = logger.chooseOne(
-        'Choose an auth provider',
-        choices: api.AuthProvider.values,
-        display: (p) => p.displayName,
-      );
+      provider = api.AuthProvider.shorebird;
```
eseidel: We should either remove the arg, or keep its functionality.  Right now we've kept it and broke it.  If someone was passing --provider microsoft, it no longer does that.

eseidel:
Issues spotted by claude (and edited by me):

1. ShorebirdAuthEndpoints accesses shorebirdEnv at call time — but is never actually used

shorebird_auth_endpoints.dart creates endpoints that reference shorebirdEnv (a scoped dependency), but the doc comment says:

The Shorebird refresh function ignores this parameter, so these values exist only to satisfy the type system.

If they're never used, the shorebirdEnv access is a latent runtime error waiting to happen if someone calls authEndpoints on AuthProvider.shorebird outside a scoped context. Consider making the endpoints throw UnimplementedError instead of silently depending on shorebirdEnv for values that are documented as unused.

2. clientId still throws for AuthProvider.shorebird

auth.dart:524 — AuthProvider.shorebird.clientId throws UnsupportedError('Shorebird auth does not use a client ID'). This is fine today since the _obtainCredentials method branches before hitting it, but it's a footgun. If anyone adds code that calls authProvider.clientId generically, they'll get a runtime exception. A comment above the switch in _obtainCredentials noting this asymmetry would help.

5. hostedUri is now passed but wasn't before

auth.dart now passes hostedUri: shorebirdEnv.hostedUri to _buildCodePushClient in both loginCi and login. This is a behavioral change beyond just adding Shorebird auth — previously the hosted URI wasn't passed, meaning CodePushClient used its own default. Is this intentional? If so, it should be called out separately since it affects Google/Microsoft login paths too.

6. No authServiceUri null check when constructing from credentials path

In AuthenticatedClient.send(), when credentials have expired and the JWT issuer is shorebird, _refreshForProvider is called which requires _authServiceUri to be non-null. But _authServiceUri is optional in the constructor. The StateError thrown at runtime is good, but it would be cleaner to make authServiceUri required when the provider could be Shorebird — or at least add an assertion in the constructor.

Minor / Style

- The catchError IPv6 fallback on line catches any error, not just SocketException. Consider .catchError((Object _) => ...) or a try/catch for more clarity.
- The test JWT token (shorebirdIdToken) is a long base64 string hardcoded as a const. It works, but a comment above it showing the decoded payload would make tests easier to read. (The // cspell:disable-next-line is already there which helps.)

packages/shorebird_cli/lib/src/auth/auth.dart
easymac: Maybe; in the context of a CLI user, if they can run `shorebird patch` there's no reason to prevent them from running `shorebird release`, etc. We already have server-side roles that limit the blast radius. The concern I believe would be the consequences of leaked tokens. A user using `login:ci` might not need a full access key, and instead something scoped to `patch`.

Auth will support tokens in the backend for API keys, but I'm not sure that we'll need to support them in the CLI package.

packages/shorebird_cli/lib/src/commands/login_command.dart
easymac: `shorebird login --provider microsoft` still works for me. This change didn't remove existing providers; just added support for shorebird auth, keeping in line with the changes to the server. I thought that was the plan, but I can remove the existing providers in this change too.

packages/shorebird_cli/lib/src/commands/login_ci_command.dart
easymac: Do you mean redirecting from `auth.shorebird.dev/login` to the Auth service's OAuth flow to the particular provider? We'd have to add support for that in the Auth service, but it's a small ask if that's what you mean.

packages/shorebird_cli/lib/src/commands/login_command.dart
eseidel: Maybe I misunderstood then. :)

packages/shorebird_cli/lib/src/commands/login_ci_command.dart
eseidel: I'm just not sure what this option does then. Previously it would redirect you to the proper oauth flow, correct? And now shorebird's is the only oauth flow and it's just a qeustion of if you bounce through microsoft or Gogole to get your shorebird token?  Does this flag still start you at the right oauth? I guess I'm not sure what "shorebird" for this flag would even mean?

eseidel (approved):

PR #3643, opened 2026-03-10
fix(shorebird_cli): clean stale obfuscation map from supplement directory

bdero:
## Summary
- When a non-obfuscated release is created after a previous obfuscated build, the supplement directory (`build/<platform>/shorebird/`) may still contain an `obfuscation_map.json` from the earlier build
- This stale file gets uploaded as the release's supplement artifact, causing the patch command to incorrectly detect the release as obfuscated and pass `--load-obfuscation-map` to gen_snapshot
- `assembleSupplementDirectory()` now removes any stale obfuscation map when obfuscation is not enabled and returns `null` if the directory is empty, preventing accidental upload

## Test plan
- [x] Existing `aar_releaser_test.dart` tests pass (22/22)
- [x] E2E: non-obfuscated add-to-app release + patch succeeds on Android and iOS


bdero:
Noticed this issue while doing smoketests for the Flutter 3.38.10 release.

easymac (approved):

PR #3645, opened 2026-03-13
feat: use git worktrees for Flutter version management

eseidel:
## Summary

- Switch Flutter revision installs from individual `git clone` to a single shared bare repo (`bin/cache/flutter/.repo/`) with `git worktree add`
- All worktrees share one object store — only one `git fetch` needed for new revisions instead of a full clone
- Old standalone clones (pre-migration) continue to work without migration
- Git query methods (`getVersionForRevision`, `resolveFlutterRevision`, etc.) prefer the bare repo when available
- Add usage tracking via `.last_used/<revision>` timestamp files
- Add `shorebird cache clean --prune` to remove revisions unused for 30+ days
- Add `cloneBare`, `worktreeAdd`, `worktreeRemove` to `Git` class

## Test plan

- [x] All existing tests updated and passing (94 tests across 3 test files)
- [ ] Manual test: delete `bin/cache/flutter/`, run `shorebird release` — should create `.repo/` + worktree
- [ ] Manual test: with existing old-style clones present, run `shorebird release` — old clones still work, new revision uses worktree
- [ ] Manual test: `shorebird cache clean --prune` removes old revisions, preserves current

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -138,6 +138,7 @@ words:
   - unmockable
   - unobfuscated
   - unpadded
+  - unparseable
   - Unpatchable
   - unsets
   - unskipped
@@ -147,6 +148,7 @@ words:
   - VCRUNTIME
   - Verdana
   - vmcode
+  - worktree
   - writeln
   - WRONGPASS
   - xcarchive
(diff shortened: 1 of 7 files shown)


eseidel:
This is not ready for review.  The idea here is to reduce our footprint on disk.  I don't think this is a burning customer issue, but we have gotten a few complaints.

eseidel:
Closing this for now.

PR #3647, opened 2026-03-13
fix: invalidate bootstrap stamp when Flutter version changes

bdero:
## Summary
- `shared.sh` keys the bootstrap stamp on the Shorebird git revision alone
- Changing `bin/internal/flutter.version` without a new commit skips re-bootstrap, so the new Flutter SDK is never downloaded
- The CLI then fails with `dart: No such file or directory` because `DART_PATH` points into the non-existent SDK directory
- Fix: clear the stamp in the entrypoint script (`bin/shorebird`) when the Flutter SDK directory for the requested version doesn't exist yet

## Test plan
- [x] Manually verified: change `flutter.version` to a new hash, run `shorebird --version`, confirm it downloads the new SDK

### DIFF
--- a/bin/shorebird
+++ b/bin/shorebird
@@ -10,6 +10,15 @@ PROG_NAME="$BIN_DIR/$(basename "$BASH_SOURCE")"
 OS="$(uname -s)"
 
 FLUTTER_VERSION=`cat "$BIN_DIR/internal/flutter.version"`
+
+# If the Flutter SDK for the requested version hasn't been downloaded yet,
+# clear the bootstrap stamp so that shared.sh triggers a re-bootstrap.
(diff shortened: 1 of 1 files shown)


bdero:
Tiny fix for some repeat friction I keep running into while testing releases

eseidel (approved):
I assume this is just a developer convenience for when we're working locally. I'm trying to remember a time that I ever changed the flutter version and didn't commit.

easymac:
I believe I ran into this problem just this afternoon. Nice!

easymac (approved):

bdero:
yeah, I tend to switch the the new version and run smoketests and make the commit after tests pass

PR #3653, opened 2026-03-19
feat: Revoke sessions on CLI logout

easymac:
## Description

(Doesn't need to go in changelog)

- `shorebird logout` now revokes the server-side session by POSTing the refresh token to the auth service's `/api/logout` endpoint before clearing local credentials
- Session revocation is best-effort - local credentials are always cleared, even if the server is unreachable or returns an error
- Revocation is skipped when no refresh token is present (e.g., API key auth)

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -357,7 +357,41 @@ class Auth {
   }
 
   /// Logs out the user.
-  void logout() => _clearCredentials();
+  ///
+  /// If a Shorebird refresh token is available, revokes the server-side
+  /// session before clearing local credentials. Local credentials are always
+  /// cleared even if the server call fails.
+  Future<void> logout() async {
+    await _revokeSession();
+    _clearCredentials();
+  }
+
+  /// Sends the current refresh token to the auth service's logout endpoint
+  /// to revoke the server-side session. Failures are logged but swallowed
+  /// so that local logout always succeeds.
+  Future<void> _revokeSession() async {
+    final refreshToken = _credentials?.refreshToken;
+    if (refreshToken == null) return;
+
+    try {
+      final logoutUrl = _authServiceUri.replace(
+        path: p.url.join(_authServiceUri.path, 'api/logout'),
+      );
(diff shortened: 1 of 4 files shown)


eseidel (approved):
What happens if the revoke times out?  I presume it just deletes the local credentials after the timeout?

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -357,7 +357,41 @@ class Auth {
...
+    final refreshToken = _credentials?.refreshToken;
+    if (refreshToken == null) return;
+
+    try {
+      final logoutUrl = _authServiceUri.replace(
+        path: p.url.join(_authServiceUri.path, 'api/logout'),
+      );
+      final response = await _httpClient.post(
+        logoutUrl,
+        headers: {'Authorization': 'Bearer $refreshToken'},
```
bdero, line 382: Here and anywhere else we might be doing this: seems unconventional to send the refresh token as if it's an access token (`Authorization: Bearer ...`). Consider sending `{"refresh_token": "..."}` as part of the body instead, like rfc7009.

packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -357,7 +357,41 @@ class Auth {
...
+      );
+      if (response.statusCode != 200) {
```
bdero: nit: broaden to accept any "success" response. `response.statusCode < 200 || response.statusCode >= 300`

bdero (approved):

packages/shorebird_cli/lib/src/auth/auth.dart
easymac: This seems appropriate and prudent, but will require a larger refactor. Filing an issue against Auth and we'll land it in another PR.

easymac:
Yes, when the server times out (or there are any other problems revoking the session), the credentials are still cleared and the user is logged out & able to login again.

PR #3656, opened 2026-03-21
fix: list orgs and exit gracefully when init is non-interactive

### LINKED ISSUE #3635: CLI: shorebird init unusable in non-interactive environments (no way to discover org ID)
## Problem

When running `shorebird init` in a non-interactive environment (CI, agent, piped stdin), the command fails because it can't prompt for organization selection:

```
StdinException: Error setting terminal echo mode, OS Error: Operation not supported by device, errno = 19
```

The `--organization-id` flag exists, but there's no CLI command to list available organizations and their IDs. The only way to discover your org ID is through the interactive prompt, which defeats the purpose.

## Expected

Either:
1. Add a command like `shorebird organizations list` that prints org IDs, or
2. Have `shorebird init` print available orgs and exit gracefully when stdin is not a TTY

## Reproduction

```bash
# This fails with StdinException
echo "" | shorebird init --force --display-name "Test"

# This requires an org ID, but there's no command to find it
shorebird init --force --display-name "Test" --organization-id ???
```

## Context

Discovered while trying to automate a test build from Claude Code (agent environment without TTY).

eseidel:
## Summary

- When `shorebird init` needs org selection but stdin is not a TTY, list available organizations with their IDs instead of crashing with `StdinException`
- Tells the user to re-run with `--organization-id <id>`
- Exits with `ExitCode.usage`

Closes #3635

## Test plan

- [x] New test: non-interactive with multiple orgs lists them and exits
- [x] All 41 existing init_command tests pass
- [ ] CI green


eseidel:
This should help Claude.

bdero (approved):
lgtm

PR #3657, opened 2026-03-21
feat: warn that preview reinstalls the app on mobile

### LINKED ISSUE #3503: feat: shorebird preview should explain that it clears data
`shorebird preview` does a complete re-install of the app.  This is sometimes confusing.  It should at least ahve a message that that's what it is doing.

eseidel:
## Summary

- Adds a message before installing on Android and iOS: "This will reinstall the app and clear its data on the device."
- Only shown on mobile platforms where preview actually does a reinstall
- Desktop platforms (macOS, Windows, Linux) just run the binary directly, so no message is shown

Closes #3503

## Test plan

- [x] All existing preview_command tests pass
- [ ] CI green


bdero (approved):
lgtm

PR #3658, opened 2026-03-21
feat: suggest cache clean in build failure recommendations

### LINKED ISSUE #3245: feat: clearer logs when a flutter installation is corrupted
In https://github.com/shorebirdtech/shorebird/issues/3244, a user was confronted with an unhelpful error message because their shorebird Flutter checkout was corrupted. We should provide a more helpful error message telling users to run `shorebird cache clean` or similar

eseidel:
## Summary

- Adds a `shorebird cache clean` suggestion to the build failure fix recommendations
- Shown when any `flutter build` command fails, alongside the existing suggestions
- Helps users recover from corrupted Flutter installations without needing to diagnose the root cause

Closes #3245

## Test plan

- [ ] CI green


bdero (approved):
lgtm

PR #3671, opened 2026-03-31
feat: add DD table files to supplement file handling

bdero:
## Summary
- Adds `App.dd.link` and `App.dd_callers.link` to the supplement file list so they get copied between snapshot directories during patch builds.
- These files are produced by the 2-pass release build (in Flutter tooling) and consumed by `aot_tools` during patch linking for DD slot resolution via winner voting.

## Test plan
- [ ] Verify DD table files are copied during patch builds
- [ ] Run existing shorebird CLI tests

### DIFF
--- a/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
@@ -60,15 +60,27 @@ final artifactBuilderRef = create(ArtifactBuilder.new);
 /// The [ArtifactBuilder] instance available in the current zone.
 ArtifactBuilder get artifactBuilder => read(artifactBuilderRef);
 
-extension on String {
-  /// Converts this base64-encoded public key into the `Map<String, String>`:
-  ///   {'SHOREBIRD_PUBLIC_KEY': this}
-  ///
-  /// SHOREBIRD_PUBLIC_KEY is the name expected by the Shorebird's Flutter tool
-  ///
-  /// This allow us to just call var?.toPublicKeyEnv() instead of doing
-  /// a ternary operation to check if the value is null.
-  Map<String, String> toPublicKeyEnv() => {'SHOREBIRD_PUBLIC_KEY': this};
+/// Builds the environment map for Flutter build subprocesses.
+///
+/// Merges the public key (if present) with any additional environment
+/// variables (e.g. DD table configuration).
+///
+/// Environment variables are used (rather than command-line flags) for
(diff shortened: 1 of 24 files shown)


packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
```diff
@@ -60,15 +60,27 @@ final artifactBuilderRef = create(ArtifactBuilder.new);
...
+Map<String, String>? buildEnvironment({
+  String? base64PublicKey,
+  int? ddMaxBytes,
+}) {
+  final env = <String, String>{};
+  if (base64PublicKey != null) {
+    env['SHOREBIRD_PUBLIC_KEY'] = base64PublicKey;
+  }
+  if (ddMaxBytes != null) {
+    env['SHOREBIRD_DD_MAX_BYTES'] = ddMaxBytes.toString();
```
eseidel, line 81: Is this only for debugging, or do you expect that we want to respect this later? I just wonder if you even want to do this plumbing, now that the system has stabilized some?

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -159,6 +159,15 @@ of the iOS app that is using this module. (aar and ios-framework only)''',
         CommonArguments.obfuscateArg.name,
         help: CommonArguments.obfuscateArg.description,
         negatable: false,
+      )
+      ..addOption(
+        'dd-max-bytes',
```
eseidel, line 164: I suspect you want this to be hidden at least. I don't think you want to expose this to users.

eseidel (approved):

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
bdero: _bdero's review bot_: Keeping the plumbing. Even though the cascade limiter has stabilized, the `--dd-max-bytes` knob (now hidden) lets us cut DD off in case of regressions without rebuilding the engine, and lets internal/automated builds pin a specific budget. We've leaned on it heavily over the last week of link-percentage debugging. Would prefer to leave it in place for now and revisit removal once the system has been quiet for longer.

packages/shorebird_cli/lib/src/commands/release/release_command.dart
bdero: _bdero's review bot_: Done in ea744b58 — added `hide: true`.

PR #3672, opened 2026-03-31
feat: include snapshots in patch debug dump

bdero:
## Summary
- Copies the release snapshot and 3 patch compilation stage snapshots (`out.aot`, `out.ct.aot`, `out.optimized.aot`) into a `snapshots/` subdirectory in the debug dump zip.
- Enables offline analysis of linking issues using the actual snapshot binaries.

## Test plan
- [x] Added test verifying all 4 snapshots are copied when present
- [x] Added test verifying missing intermediate snapshots are silently skipped
- [x] All existing `runLinker` tests pass


easymac (approved):
LGTM!

eseidel:
This looks fine, it will make the dump significantly larger.  I'm not actually sure if you need this, though, if we simply have the app_id/release number, we can pull these artifacts from storage.   Or am I misunderstanding?

bdero:
We could pull the release snapshot with some additional metadata, but we don't upload the patch snapshots anywhere AFAICT.

PR #3674, opened 2026-04-02
feat: support module_version for add-to-app AAR releases

eseidel:
## Summary
- When `--release-version` is omitted for AAR releases, default to git commit short hash as both the server-side release version and the module version baked into the AAR
- Pass module version to `flutter build aar` via `SHOREBIRD_MODULE_VERSION` env var; the Flutter tool writes it as `module_version` into the compiled `shorebird.yaml`
- Add `moduleVersion` field to `ShorebirdYaml` schema (so the CLI accepts the field)
- Add `moduleVersion` field to `PatchCheckRequest` protocol
- Require Flutter >= 3.41.4 when `--release-version` is omitted (older Flutter ignores the env var)
- Add `moduleVersion` parameter to `buildAar()` in artifact builder

## Context
Addresses #793 — enables add-to-app (AAR) use cases where the same Flutter module is embedded in multiple host apps with different version numbers.

`release_version` always stays the host app's version (for clean analytics). `module_version` is a separate optional field used only for patch lookup.

**Companion PRs:**
- Flutter tool + engine: https://github.com/shorebirdtech/flutter/pull/121
- Updater: https://github.com/shorebirdtech/updater/pull/329
- Server: https://github.com/shorebirdtech/_shorebird/pull/1974

## Test plan
- [ ] `assertArgsAreValid` tests for Flutter version gating and git hash resolution
- [ ] `getReleaseVersion` tests for explicit and git-hash paths
- [ ] `shorebird_yaml_test` for `module_version` deserialization
- [ ] Verify non-AAR releases are unaffected

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
@@ -72,11 +72,7 @@ class AarReleaser extends Releaser {
 
   @override
   Future<void> assertArgsAreValid() async {
-    if (!argResults.wasParsed('release-version')) {
-      logger.err('Missing required argument: --release-version');
-      throw ProcessExit(ExitCode.usage.code);
-    }
-
+    await resolveModuleReleaseVersionArgs();
     await assertObfuscationIsSupported();
   }
 
@@ -86,12 +82,15 @@ class AarReleaser extends Releaser {
     final buildArgs = [...argResults.forwardedArgs];
     addSplitDebugInfoDefault(buildArgs);
     addObfuscationMapArgs(buildArgs);
+
     await artifactBuilder.buildAar(
       buildNumber: buildNumber,
       targetPlatforms: architectures,
       args: buildArgs,
       base64PublicKey: base64PublicKey,
+      moduleVersion: moduleVersion,
     );
+
     verifyObfuscationMap();
 
     // Copy release AAR to a new directory to avoid overwriting with
@@ -111,7 +110,7 @@ class AarReleaser extends Releaser {
   Future<String> getReleaseVersion({
     required FileSystemEntity releaseArtifactRoot,
   }) async {
-    return argResults['release-version'] as String;
+    return moduleReleaseVersion;
   }
 
   @override
--- a/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
@@ -69,6 +69,25 @@ extension on String {
   /// This allow us to just call var?.toPublicKeyEnv() instead of doing
   /// a ternary operation to check if the value is null.
   Map<String, String> toPublicKeyEnv() => {'SHOREBIRD_PUBLIC_KEY': this};
+
+  /// Returns a map with the SHOREBIRD_MODULE_VERSION environment variable.
+  Map<String, String> toModuleVersionEnv() => {
+    'SHOREBIRD_MODULE_VERSION': this,
+  };
+}
+
+/// Combines public-key and module-version env vars into a single map for
+/// passing through to `flutter build`. Returns `null` when neither is set
+/// so we don't override the parent process env with an empty map.
+Map<String, String>? _buildEnv({
+  String? base64PublicKey,
+  String? moduleVersion,
+}) {
+  if (base64PublicKey == null && moduleVersion == null) return null;
+  return {
+    ...?base64PublicKey?.toPublicKeyEnv(),
+    ...?moduleVersion?.toModuleVersionEnv(),
+  };
 }
 
 /// @{template artifact_builder}
@@ -286,6 +305,7 @@ Reason: Exited with code $exitCode.''',
     Iterable<Arch>? targetPlatforms,
     List<String> args = const [],
     String? base64PublicKey,
+    String? moduleVersion,
   }) async {
     return _runShorebirdBuildCommand(() async {
       const executable = 'flutter';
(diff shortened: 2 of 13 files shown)


packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
```diff
@@ -72,26 +74,70 @@ class AarReleaser extends Releaser {
...
+    // When --release-version is omitted, we use the new env var flow which
+    // requires a Flutter version that supports SHOREBIRD_MODULE_VERSION.
     if (!argResults.wasParsed('release-version')) {
-      logger.err('Missing required argument: --release-version');
-      throw ProcessExit(ExitCode.usage.code);
+      final flutterVersion = await shorebirdFlutter.resolveFlutterVersion(
+        shorebirdEnv.flutterRevision,
+      );
+      if (flutterVersion != null &&
+          flutterVersion < minimumModuleVersionFlutterVersion) {
```
eseidel: I'm not sure when it's possible for Flutter version to be null?  But I suspect in the null case it should show this warning?

packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
```diff
@@ -111,7 +157,13 @@ class AarReleaser extends Releaser {
   Future<String> getReleaseVersion({
     required FileSystemEntity releaseArtifactRoot,
   }) async {
-    return argResults['release-version'] as String;
+    // When --release-version is provided, use it as the server-side release
+    // version (old flow). Otherwise use the module version (git hash).
+    if (argResults.wasParsed('release-version')) {
+      return argResults['release-version'] as String;
+    }
+    _moduleVersion ??= await _resolveModuleVersion();
```
eseidel: It feels like this resolution should happen earlier in some sort of args resolve function, and then we should store the release version and the module version both on this same object and then use them in the various places.

eseidel: My concern is less about where exactly it happens and more that we seem to keep having to reach inside our results and parse it each time, when I feel like we should only have to have done that once. At the same time, we should do the module version.

packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
```diff
@@ -70,11 +72,55 @@ class AarReleaser extends Releaser {
...
+        );
+        throw ProcessExit(ExitCode.software.code);
+      }
+      final gitHash = (result.stdout as String).trim();
+      logger.info(
+        'No --release-version provided, using git hash: '
+        '${lightCyan.wrap(gitHash)}',
+      );
+      _moduleVersion = gitHash;
+      _releaseVersion = gitHash;
```
eseidel: This is wrong.  _releaseVersion should jsut be null here, no?

packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
```diff
@@ -111,7 +160,7 @@ class AarReleaser extends Releaser {
   Future<String> getReleaseVersion({
     required FileSystemEntity releaseArtifactRoot,
   }) async {
-    return argResults['release-version'] as String;
+    return _releaseVersion;
```
eseidel: Shouldn't this just be nullable?

packages/shorebird_cli/lib/src/commands/release/aar_releaser.dart
```diff
@@ -70,11 +73,53 @@ class AarReleaser extends Releaser {
...
+          'require passing --release-version explicitly for AAR releases '
+          '(current: $flutterVersion).',
+        );
+        throw ProcessExit(ExitCode.unavailable.code);
+      }
+
+      final result = await process.run(
+        'git',
+        ['rev-parse', '--short', 'HEAD'],
+      );
```
eseidel: Are we running this in the right directory?

bdero: Discussed "offline". Maybe we should prompt the user to choose a release ID (and of course force it to be unique) when they don't provide one for AAR releases, but fill in the git hash as the default for them if we see they happen to be using git

bdero:
> release_version always stays the host app's version (for clean analytics). module_version is a separate optional field used only for patch lookup.

Can you describe the full workflow you're intending a customer to follow so I can wrap my mind around how this solution fully addresses the problem?

If customer CoolCustomer has two apps NeatApp and AwesomeApp which both embed a shorebird-enabled AAR Flutter module SocialModule, and CoolCustomer wants to release a bugfix in SocialModule across their two apps at the same time... how do they set up the releases ahead of time and what does patch creation look like?

Is the intent to have the user generate separate Shorebird releases for NeatApp and AwesomeApp, but with a shared module version for SocialModule & hopefully identical code (for the purposes of making stuff like MAU maximally useful), thereby allowing the user to create one patch that will distribute to all releases attributed to the SocialModule module version?

eseidel:
My expectation would be that if customer has apps A and B and module M, they would build module M once, and then embed it into A and B as though it were a 3p dependency (already comes prebuilt).  Thus M would end up with the same module_version in our system, and when they send a patch to M, it would appear in both A and B?

eseidel:
So `shorebird release aar --module-version=git` take the output, and put it wherever it goes for your other two apps to consume and then they just get each built with gradel (or XCode) as before.

When you want to patch, you just then run `shorebird patch aar --module-version=latest` (I think I have to add that component).

I think I need to get a full setup working locally (although I'm not entirely sure I can do that before landing at least the server bits) to verify.

bdero:
Triage: Eric is writing a PRD for this at the moment.

bdero:
Now that the doc is written, would you like review on these now @eseidel? Or were there some more changes you were looking to make on these `module_version` patches?

eseidel:
I think once DD is land, this is on the list of things I should hand to you instead of me. :)

PR #3675, opened 2026-04-02
feat(shorebird_cli): generate podspec for iOS framework releases

eseidel:
## Summary

- Generates a `ShorebirdFlutter.podspec` in the `release/` directory alongside the existing xcframeworks during `shorebird release ios-framework`
- Updates post-release instructions to show both CocoaPods and manual Xcode embedding options

This enables users to integrate Shorebird's iOS framework add-to-app via CocoaPods (`pod 'ShorebirdFlutter', :path => 'release'`) instead of manually configuring Xcode framework search paths and embedding.

## Context: iOS add-to-app integration approaches

Flutter documents two main iOS add-to-app integration methods:
- **Option A: CocoaPods** — lower friction, manages framework linking automatically
- **Option B: Manual framework embedding** — what Shorebird supports today

We have customer demand for CocoaPods-based integration. There are two possible approaches:

### Approach A (this PR): Wrap existing xcframeworks in a podspec

The build process is unchanged — we still produce `App.xcframework` and `ShorebirdFlutter.xcframework`. We just additionally generate a podspec that wraps them so users can do `pod install` instead of manual Xcode configuration.

**Pros:**
- Minimal change — just generates a podspec file alongside existing output
- No changes to the build pipeline, patching infrastructure, or artifact uploads
- No project modifications required (user still runs `shorebird release ios-framework` as before)
- The CocoaPods integration step is no different from what they'd do for any framework dependency

**Cons:**
- Not the "full" CocoaPods add-to-app experience — Flutter code doesn't recompile automatically during native Xcode builds
- User still runs `shorebird release ios-framework` separately when Flutter code changes
- Doesn't match Flutter's documented "Option A" workflow exactly (which compiles Flutter from source during `pod install`)

### Approach B (not in this PR): Hook into source-compilation CocoaPods flow

This would be the "real" Option A support where Flutter compiles from source as part of `pod install`. This would require:
- Injecting Shorebird's engine into the CocoaPods build pipeline
- Users modifying their project (Podfile, env vars, or build scripts) to use Shorebird's Flutter fork
- Understanding and hooking into Flutter's iOS build system at the podspec level

This breaks Shorebird's current "no project modifications" advantage and is significantly more complex. Worth exploring only if Approach A doesn't satisfy customer demand.

## Test plan

- [x] Existing `ios_framework_releaser_test.dart` tests pass
- [x] New test verifies podspec is generated with correct content
- [x] Post-release instructions test updated for new two-option format
- [ ] Manual: run `shorebird release ios-framework` and verify `release/ShorebirdFlutter.podspec` is generated
- [ ] Manual: verify `pod install` works in a native iOS app with the generated podspec


packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -148,18 +151,44 @@ class IosFrameworkReleaser extends Releaser {
...
+      releaseDir.path,
+      'ShorebirdFlutter.podspec',
+    );
+    File(podspecPath).writeAsStringSync('''
+Pod::Spec.new do |s|
+  s.name         = 'ShorebirdFlutter'
+  s.version      = '0.0.1'
+  s.summary      = 'Shorebird Flutter framework for add-to-app integration.'
+  s.homepage     = 'https://shorebird.dev'
+  s.license      = { :type => 'Proprietary' }
```
eseidel: Our license is actually the same as Flutter's.

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -148,18 +151,44 @@ class IosFrameworkReleaser extends Releaser {
...
+  /// CocoaPods-based integration as an alternative to manual Xcode embedding.
+  void _writePodspec(Directory releaseDir) {
+    final podspecPath = p.join(
+      releaseDir.path,
+      'ShorebirdFlutter.podspec',
+    );
+    File(podspecPath).writeAsStringSync('''
+Pod::Spec.new do |s|
+  s.name         = 'ShorebirdFlutter'
+  s.version      = '0.0.1'
```
bdero, line 164: _bdero's review bot_:

💡 `s.version = '0.0.1'` is the same string for every release, so `Podfile.lock` and `pod outdated` will show stale info regardless of which Shorebird release the user is actually shipping. With `:path =>` integration the version doesn't drive resolution, so nothing breaks — but plumbing through `argResults['release-version']` (the same value `getReleaseVersion` returns) makes the lockfile actually reflect reality. One catch: Flutter's `release-version` can include `+build` suffixes, and CocoaPods' SemVer parser is stricter than pub's, so we may need to strip or transform.

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -148,18 +151,44 @@ class IosFrameworkReleaser extends Releaser {
...
+  s.summary      = 'Shorebird Flutter framework for add-to-app integration.'
+  s.homepage     = 'https://shorebird.dev'
+  s.license      = { :type => 'BSD-3-Clause' }
+  s.author       = 'Shorebird'
+  s.source       = { :path => '.' }
+  s.platform     = :ios, '12.0'
```
bdero, line 170: _bdero's review bot_:

💡 `s.platform = :ios, '12.0'` is hardcoded, but the actual minimum is whatever `MinimumOSVersion` ends up baked into `App.xcframework` and `ShorebirdFlutter.xcframework` — driven by Flutter's minimum, which has been creeping up. CocoaPods doesn't cross-check podspec platform against framework Info.plist, so a future Flutter min-version bump would silently let users link Shorebird into iOS 12 apps where the binaries actually require 13+. Worth either reading from the xcframework's Info.plist or pulling from a Shorebird-side known-minimum constant.

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -148,18 +151,44 @@ class IosFrameworkReleaser extends Releaser {
...
+      releaseDir.path,
+      'ShorebirdFlutter.podspec',
+    );
+    File(podspecPath).writeAsStringSync('''
+Pod::Spec.new do |s|
+  s.name         = 'ShorebirdFlutter'
+  s.version      = '0.0.1'
+  s.summary      = 'Shorebird Flutter framework for add-to-app integration.'
+  s.homepage     = 'https://shorebird.dev'
+  s.license      = { :type => 'BSD-3-Clause' }
```
bdero, line 167: _bdero's review bot_:

💡 `s.license = { :type => 'BSD-3-Clause' }` — not sure this is accurate for the bundled binaries. The Flutter engine portion is BSD-3-Clause, but Shorebird's modifications and proprietary patching code ship in `ShorebirdFlutter.xcframework` too, and I think those are under different terms. Worth a check with someone who has more licensing context — at minimum, `{ :file => 'LICENSE' }` pointing to a real bundled file is more honest than asserting one license type for the whole thing.

packages/shorebird_cli/test/src/commands/release/ios_framework_releaser_test.dart
```diff
@@ -479,6 +479,24 @@ void main() {
...
+        final content = podspecFile.readAsStringSync();
+        expect(content, contains("s.name         = 'ShorebirdFlutter'"));
+        expect(
+          content,
+          contains(
+            "s.vendored_frameworks = 'App.xcframework', "
+            "'ShorebirdFlutter.xcframework'",
+          ),
+        );
+      });
```
bdero, line 498: _bdero's review bot_:

💡 Test only asserts two `contains` checks. A heredoc typo (missing `end`, mismatched quotes) wouldn't be caught — `pod install` would discover it at runtime in production. A golden-file comparison against the full podspec content would lock down the template; it costs almost nothing and means future template edits show up clearly in diffs. Optional: shelling out to `pod ipc spec` would catch syntax errors but that adds a CI dependency we probably don't want.

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -148,18 +151,44 @@ class IosFrameworkReleaser extends Releaser {
...
+${styleBold.wrap('Option A: CocoaPods')}
+    Add the following to your app's Podfile:
+    ${lightCyan.wrap("pod 'ShorebirdFlutter', :path => '$relativeFrameworkDirectoryPath'")}
+    Then run ${lightCyan.wrap('pod install')}.
 
-Instructions for these steps can be found at https://docs.flutter.dev/add-to-app/ios/project-setup#option-b---embed-frameworks-in-xcode.
+${styleBold.wrap('Option B: Manual Xcode embedding')}
+    1. Add the relative path to the ${lightCyan.wrap(relativeFrameworkDirectoryPath)} directory to your app's Framework Search Paths in your Xcode build settings.
+    2. Embed the App.xcframework and ShorebirdFlutter.xcframework in your Xcode project.
+    Instructions: https://docs.flutter.dev/add-to-app/ios/project-setup#option-b---embed-frameworks-in-xcode
```
bdero, line 191: _bdero's review bot_:

🧹 If CocoaPods is the new recommended path, worth labeling it that way ("Option A: CocoaPods (recommended)") so users don't read Option A, glance at Option B, and feel like they need to pick. Otherwise both options read as equally weighted choices and we get more "which one should I use?" support questions.

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
```diff
@@ -148,18 +151,44 @@ class IosFrameworkReleaser extends Releaser {
...
+  void _writePodspec(Directory releaseDir) {
+    final podspecPath = p.join(
+      releaseDir.path,
+      'ShorebirdFlutter.podspec',
+    );
+    File(podspecPath).writeAsStringSync('''
+Pod::Spec.new do |s|
+  s.name         = 'ShorebirdFlutter'
+  s.version      = '0.0.1'
+  s.summary      = 'Shorebird Flutter framework for add-to-app integration.'
```
bdero, line 165: _bdero's review bot_:

🧹 `s.summary` could mention that the framework is a Shorebird-patched Flutter engine, not stock Flutter — that's the whole point. "Shorebird's patched Flutter engine for add-to-app iOS integration" or similar. Minor.

bdero (commented):
_bdero's review bot_:

## Overall

Pragmatic shape — wrapping existing xcframeworks in a podspec is the right "Approach A" given customer demand and zero changes to the build pipeline. Three things in the podspec template are hardcoded that probably shouldn't be (`s.version`, `s.platform`, `s.license`), and the test only spot-checks two lines so template-level breakage wouldn't be caught. None of this is blocking.

One reminder unrelated to a specific line: both manual checkboxes in the test plan (`shorebird release ios-framework` generates the podspec; `pod install` works in a real iOS app) are unchecked. The unit tests prove we *write* a podspec; only the manual run proves CocoaPods *accepts* it. Want this verified before merge — it's the entire customer-visible value of the PR.

bdero (approved):
This lgtm. Couple of things my review bot suggested that seem to make sense. Plz land at your discretion.

bdero:
Also, IIRC, there isn't an issue we should close when this one lands, right?

bdero:
@eseidel Looks like this is ready to go pending some clanker reviews.

eseidel:
It's ready to go, I'm just not sure if it solves a real customer problem or not.

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
eseidel: Done in 52d264c — `s.version` now carries `--release-version`.

You were right to flag the parser difference, and it is stricter than SemVer-vs-pub: CocoaPods uses RubyGems' `Gem::Version`, which has no concept of build metadata at all, so `1.2.3+4` is not "transformed", it is rejected. `+build` is dropped rather than passed through.

The part worth a second look is the fallback. `--release-version` is free-form for this platform — it is whatever the caller typed, and `getReleaseVersion` returns it verbatim with no validation — so a value like a git short hash is reachable. A podspec CocoaPods cannot parse fails `pod install` outright, which is a worse failure than the stale version this started as, so anything not matching `Gem::Version`'s shape falls back to `0.0.1` instead of being guessed at. Covered by tests for build metadata, a plain version, a pre-release tail, a hash, and empty.

---

packages/shorebird_cli/test/src/commands/release/ios_framework_releaser_test.dart
eseidel: Done in 52d264c — the test now compares the whole podspec rather than two substrings, so a dropped `end` or a mismatched quote fails in CI instead of at a user's `pod install`.

Agreed on skipping `pod ipc spec`; that would put a CocoaPods install on the critical path of every CI run to catch a class of typo the golden already catches.

Worth knowing for anyone reading this thread: the whole group is `testOn: 'mac-os'`, so this only runs in the macOS CI job, not the ubuntu one.

---

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
eseidel: Done in 52d264c:

```ruby
s.summary      = "Shorebird's patched Flutter engine, for add-to-app iOS integration."
```

Switched to double quotes so the apostrophe does not have to be escaped.

---

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
eseidel: Done in 52d264c — `Option A: CocoaPods (recommended)`.

Flagging it as a product call rather than a cleanup, though, since the PR description is more equivocal than the label now is: it lists "not the full CocoaPods add-to-app experience" and "user still runs `shorebird release ios-framework` separately" as real costs of this approach. The label matches Flutter's own docs, where CocoaPods is Option A and embedding is Option B, and it matches "lower friction" in the description — but if the intended answer for add-to-app customers is still manual embedding, drop the parenthetical and I will not argue.

---

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
eseidel: Left as-is: @eseidel already answered this on the thread above, on the same line, back in April — "Our license is actually the same as Flutter's." Flutter's is BSD-3-Clause, so `{ :type => 'BSD-3-Clause' }` is the accurate assertion and the premise that Shorebird's modifications ship under different terms does not hold.

On the `{ :file => 'LICENSE' }` alternative: it is only more honest if the file is actually there, and nothing in this PR copies a LICENSE into the release directory. Pointing a podspec at a file that does not exist is a `pod install` failure, so that swap would need the copy to land with it. Not doing that here, since the type assertion is correct.

---

packages/shorebird_cli/lib/src/commands/release/ios_framework_releaser.dart
eseidel: Real issue, left unfixed deliberately — flagging rather than guessing, because the obvious fixes are each wrong in a way that costs more than the stale constant.

The failure mode is also the opposite of harmless in one direction. Understating the minimum (saying 12.0 when the binaries need 13+) is the silent case you describe. Overstating it breaks `pod install` for users whose deployment target is genuinely fine, so a fix that guesses high is worse than the hardcode.

On the two suggestions:

- **Reading the xcframework's Info.plist.** The top-level `Info.plist` carries `AvailableLibraries` with `SupportedPlatform`/`SupportedArchitectures` and no minimum; `MinimumOSVersion` lives in each slice's inner framework `Info.plist`, as a binary plist. `shorebird_cli` does already depend on `propertylistserialization`, so parsing is available and this is the correct source of truth. What I can't do is verify it — a real `App.xcframework` is a macOS Flutter build, the tests here are `testOn: 'mac-os'`, and I'd be shipping plist-walking code validated only against a synthetic fixture. That's the shape of change that should be written by someone who can run it against a real artifact.

- **A Shorebird-side known-minimum constant.** There isn't one to reference — I checked; `shorebird_cli` has no iOS deployment-target constant today. Adding one just relocates the hardcode and adds a second thing to forget when Flutter's minimum moves.

Worth noting the blast radius is smaller than it first looks: with `:path =>` integration the version doesn't drive resolution (your other comment), and CocoaPods checks the podspec platform against the *host app's* deployment target, never against the binaries — which is exactly why the mismatch is silent either way.

Happy to do the Info.plist read as a follow-up if you want it tracked; say the word and I'll file it.

---

eseidel:
I remain unconvinced this solves a real customer problem. Closing, we can re-open when we have a target customer to land this for.

PR #3678, opened 2026-04-06
feat(shorebird_cli): reject ExportOptions.plist with manageAppVersionAndBuildNumber

eseidel:
## Summary

- Validate user-supplied `--export-options-plist` before invoking `flutter build` in both `shorebird release ios` and `shorebird patch ios`. If the plist sets `manageAppVersionAndBuildNumber` to `true`, error with usage code and explain the failure: Xcode rewrites `CFBundleVersion` in the exported IPA, so the build number that ships to App Store Connect won't match the one Shorebird recorded for the release, and patches will fail to apply.
- Expand the existing `--no-codesign` warning to explain the same failure mode for the manual Distribute App path (where the checkbox isn't observable from shorebird), and print it before the build instead of only in the post-release instructions.

Refs https://docs.shorebird.dev/code-push/troubleshooting/#patch-not-showing-up

## Test plan

- [x] `dart test test/src/platform/apple/plist_test.dart` — new tests for `assertValidExportOptionsPlist` (key absent, key false, key true, malformed plist)
- [x] `dart test test/src/commands/release/ios_releaser_test.dart` — new tests for `assertArgsAreValid` with `--export-options-plist`, plus updated post-release instructions and pre-build warning assertions
- [x] `dart test test/src/commands/patch/ios_patcher_test.dart` — new `assertArgsAreValid` group covering both branches
- [x] `dart analyze lib test` — no new issues


eseidel:
This is still valid it seems @bdero

bdero:
Did this come up recently due to a complaint or issue someone filed?

bdero (approved):

eseidel:
Yeah, I don't have a Slack or Discord link for you though.

PR #3679, opened 2026-04-07
chore: unpin very_good_cli in bootstrap

eseidel:
## Summary
- Stop pinning `very_good_cli` to 0.20.0 in `scripts/bootstrap.sh` so bootstrap picks up the latest release.
- Refresh `pubspec.lock` with updated transitive dependencies (and bump min Dart SDK to 3.11.0).

## Test plan
- [ ] CI green

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -6,5 +6,5 @@
 # export PATH=$HOME/.pub-cache/bin:$PATH
 
 # Fetch dart dependencies
-dart pub global activate very_good_cli 0.20.0
+dart pub global activate very_good_cli
 very_good packages get -r ./packages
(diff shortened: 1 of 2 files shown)


scripts/bootstrap.sh
```diff
@@ -6,5 +6,5 @@
 # export PATH=$HOME/.pub-cache/bin:$PATH
 
 # Fetch dart dependencies
-dart pub global activate very_good_cli 0.20.0
+dart pub global activate very_good_cli
```
bdero, line 9: You sure you want to do this? Seems like we're putting an awful lot of trust in very_good_cli to not surprise break us.

eseidel: Honestly, I think we should just remove the bootstrap script. I don't really need it. It was just convenience. In fact, I haven't run it in years.

bdero (approved):
Whoops, meant to hit approve on this w/ my comment earlier

PR #3692, opened 2026-04-15
feat(cli): capture build trace + privacy-safe summary on release

eseidel:
## Why

Give Shorebird customers build statistics and debugging information about their release and patch builds so we can help them diagnose and optimize build performance.

Two parts:

1. **Raw trace file** — local to the customer, viewable with https://ui.perfetto.dev. Contains file paths and target names, so not currently uploaded.
2. **Summary JSON** — privacy-safe aggregates (counts, sums, percentiles, allow-list of flags). Uploaded as part of existing release/patch metadata.

Companion PRs: shorebirdtech/flutter#116 (flutter `--shorebird-trace` option), shorebirdtech/dart-sdk#787 (aot_tools `--trace` option).

## What it does

Outer release/patch command owns trace lifecycle: `ArtifactBuilder.prepareBuildTrace(platform:)` runs once per platform inside runScoped, populating `BuildTraceSession` with the trace-file path. All downstream producers (flutter, aot_tools, gen_snapshot subprocess, shorebird_cli's own HTTP spans) append to that file. `ArtifactBuilder.writeBuildTraceSummary()` finalizes the summary after all post-build work completes, so the summary uploaded with release/patch metadata reflects the full command.

**Coverage (release + patch, same builders):**

| Build              | `flutter build` → `--shorebird-trace` |
|--------------------|:-:|
| AAB                | ✅ |
| APK                | ✅ |
| AAR                | ✅ |
| Linux              | ✅ |
| macOS              | ✅ |
| iOS (IPA)          | ✅ |
| iOS framework      | ✅ |
| Windows            | ✅ |

Plus:
- `aot_tools` (link, dump_blobs, link_metadata, etc.) — `--trace=<path>` prepended globally so every subcommand writes into the same file.
- `gen_snapshot` invoked directly by `buildElfAotSnapshot` — wrapped in a `shorebirdTracer` span so the duration shows as a subprocess event.

**Outputs:**
- Raw trace: `build/shorebird/debug/build-trace-<platform>.json` (Chrome Trace Event Format).
- Summary: `build/shorebird/debug/build-trace-<platform>-summary.json` — privacy-safe aggregates.

Build environment snapshot (detected CI provider, Gradle/Xcode caching config, Develocity/ccache presence) is included in the summary.

`bin/internal/flutter.version` is pinned to the current `add-build-trace` tip, which carries the `aot_tools --trace` roll.

## Summary schema (v8)

Nested per-topic sections:

- **totals** — command wall clock, flutter umbrella, shorebird overhead (derived: `shorebirdLocalMs = shorebirdOverheadMs − network.ms`), dart vs non-dart, native-compile residual.
- **dart** — kernel snapshot, gen_snapshot, user `build.dart` scripts, assets, codegen, residual.
- **gradle** — per-task histograms (count + sum + p50/p90/max) bucketed by kind (compileKotlin, compileJava, AGP, AIDL, ProGuard, …).
- **xcode** — per-subsection histograms from xcresulttool parsing.
- **network** — Flutter- and Shorebird-side HTTP timings.
- **environment** — CI provider, caching config flags (all bool/int/enum, no paths).

The most useful single signal is **`nativeCompileMs`** (native subprocess − Σ Flutter assemble time): the wall time a Flutter-aware native build cache could eliminate.

## Real data (plugin-heavy app, cold builds)

| Platform    | flutterBuildMs | dartMs | nativeCompileMs | % native |
|-------------|---------------:|-------:|----------------:|---------:|
| Android AAB |      1,003,915 | 17,111 |         981,527 | **97.8%** |
| iOS ipa     |         52,822 | 14,196 |          26,528 |     50.2% |

Both platforms show native-compile dominance at cold-build scale.

## Test plan

- [x] `dart test` in `packages/shorebird_cli` — 1769 passing
- [x] `dart analyze --fatal-warnings lib test` — clean
- [x] Ran `shorebird release android` (AAB + `--artifact=apk`) and `shorebird release ios --no-codesign` on a plugin-heavy scaffold — summary written, values match hand-computed sums from the raw trace
- [x] Opened the raw trace in https://ui.perfetto.dev — flutter + aot_tools + gen_snapshot spans all appear in the merged view

## Follow-ups (not in this PR)

- Server-side: BigQuery consumer for the uploaded `buildTraceSummary` blob + release-side pub/sub publication to fix the release/patch write-path asymmetry (release GCS-writes inline, patch pub/subs).
- Opt-in upload of the full raw trace file to save/display to the user.


packages/shorebird_cli/lib/src/artifact_builder/shorebird_tracer.dart
```diff
@@ -0,0 +1,114 @@
...
+    }
+    _tracer
+      ..addProcessNameMetadata(pid: _pid, name: 'shorebird_cli')
+      ..addThreadNameMetadata(pid: _pid, tid: _networkTid, name: 'network')
+      ..addThreadNameMetadata(
+        pid: _pid,
+        tid: _shorebirdTid,
+        name: 'shorebird_cli',
+      );
+    _tracer.writeToFile(traceFile);
```
bdero: So it would seem that `mergeInfo` reads the entire trace file to validate the JSON array, and then `_tracer.writeToFile()` reads it again to merge. Not sure how huge these traces can be, but maybe offer a pre-parsed param to avoid the second disk round trip.

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
```diff
@@ -504,6 +532,145 @@ Reason: Exited with code $exitCode.''',
...
+    buildTraceSession.traceFile = traceFile;
+  }
+
+  /// Emits a flow-start event (`ph: "s"`) on the shorebird_cli tracer
+  /// tied to the spawned flutter process's real pid. When flutter builds
+  /// with `--shorebird-trace`, it records a flow-end with its own pid as
+  /// the flow id — Perfetto draws an arrow from our spawn point into
+  /// flutter's first span.
+  void _emitFlutterSpawnFlow(Process flutter) {
+    shorebirdTracer.addSpawnFlowStart(id: flutter.pid, at: DateTime.now());
```
bdero: Using `DateTime.now()` here is inconsistent with other stuff in this patch. Consider using the `clock` package (like we do in `writeBuildTraceSummary`) for better testability instead.

bdero (approved):
lgtm. My comments are for minor issues that can be fixed during follow-up work.

packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
eseidel: Addressed in 112fdcfb — switched to `clock.now()`.

packages/shorebird_cli/lib/src/artifact_builder/shorebird_tracer.dart
eseidel: Addressed in d735f2c2 — `BuildTracer.writeToFile` now accepts an optional `existingEvents` param and `mergeInto` forwards its already-parsed events so the file is only read once.

PR #3694, opened 2026-04-17
fix: fetch latest refs before resolving --flutter-version

### LINKED ISSUE #3693: Flutter release branch refs may be stale when resolving --flutter-version
When a user runs `shorebird release --flutter-version=3.38.5`, the CLI resolves the version string to a commit hash via `git rev-parse refs/remotes/origin/flutter_release/3.38.5` in the local Flutter clone (`shorebird_flutter.dart:267`). However, the local clone's remote refs are only updated during CLI bootstrap (`shared.sh:14`, inside `update_flutter`), which is gated on stamp invalidation — meaning it only runs after `shorebird upgrade`.

If a `flutter_release/*` branch is updated on the remote (e.g., to include an engine fix), users who haven't upgraded Shorebird will resolve the old commit hash, build against the stale Flutter, and never pick up the fix.

## Expected behavior

When building a release, the CLI should fetch the latest remote refs for the Flutter clone before resolving `--flutter-version` to a commit hash, ensuring users always get the latest commit for a given release branch.

## Suggested fix

Add a `git fetch` call in `ShorebirdFlutter.getRevisionForVersion()` (or at the start of the release flow before version resolution) targeting the default Flutter clone at `_workingDirectory()`. This ensures `refs/remotes/origin/flutter_release/*` are up to date before `rev-parse` runs.

## Affected code

- `packages/shorebird_cli/lib/src/shorebird_flutter.dart` — `getRevisionForVersion()` (line 265)
- `third_party/flutter/bin/internal/shared.sh` — `update_flutter()` (line 12) — currently the only place `git fetch` runs

bdero:
## Summary

- Adds a `git fetch` in `getRevisionForVersion()` before resolving a Flutter version string to a commit hash, ensuring `refs/remotes/origin/flutter_release/*` are up to date
- Logs a warning if the fetch fails (e.g., offline) so users know they may be building against stale data
- Previously, release branch refs were only updated during CLI bootstrap (`shared.sh`), which is gated on stamp invalidation — meaning users had to `shorebird upgrade` to pick up updated release branches

Closes #3693

## Test plan

- [x] Added test verifying `git.fetch` is called before `rev-parse`
- [x] Added test verifying warning is logged and resolution still works when fetch fails
- [x] All existing `shorebird_flutter_test.dart` tests pass (42/42)

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_command.dart
@@ -420,6 +420,10 @@ For more information see: ${supportedFlutterVersionsUrl.toLink()}''');
   Future<String> resolveTargetFlutterRevision() async {
     if (flutterVersionArg == 'latest') return shorebirdEnv.flutterRevision;
 
+    // Fetch the latest remote refs so that release branch pointers
+    // (e.g. flutter_release/3.38.5) are up to date.
+    await shorebirdFlutter.fetchRemoteRefs();
+
     final String? revision;
     try {
       revision = await shorebirdFlutter.resolveFlutterRevision(
--- a/packages/shorebird_cli/lib/src/shorebird_flutter.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_flutter.dart
@@ -268,6 +268,19 @@ class ShorebirdFlutter {
     }
   }
 
+  /// Fetches the latest remote refs for the Flutter clone so that
+  /// release branch pointers (e.g. `flutter_release/3.38.5`) are up to date.
+  Future<void> fetchRemoteRefs() async {
+    try {
+      await git.fetch(directory: _workingDirectory());
+    } on Exception {
+      logger.warn(
+        'Failed to fetch latest Flutter versions. '
+        'Resolving with potentially stale data.',
+      );
+    }
+  }
+
   /// Returns the git revision for the provided [version].
   /// e.g. 3.16.3 -> b9b23902966504a9778f4c07e3a3487fa84dcb2a
   Future<String?> getRevisionForVersion(String version) async {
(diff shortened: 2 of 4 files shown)


eseidel:
My only concern would be if we're running this repeatedly within a `release` command.  If we're only running this once, and only on release, I think we're good.

eseidel:
_(Comment written by Claude at @eseidel's request.)_

On the "only once per process, only on release" question — traced the call graph:

- `getRevisionForVersion` has one caller in lib/: `resolveFlutterRevision` (`shorebird_flutter.dart:230`).
- `resolveFlutterRevision` has one caller in lib/: `resolveTargetFlutterRevision` in `release_command.dart:425`, which runs once per `shorebird release`.
- It short-circuits on `--flutter-version latest` before calling `resolveFlutterRevision`, and the raw-hash branch of `resolveFlutterRevision` goes through `revParse` directly, so `getRevisionForVersion` is only hit on the semver path.
- `patch` and other commands don't call it. `resolveFlutterVersion` (the reverse lookup) calls `getVersionForRevision`, not this one, so the other call sites in `release_command.dart:399` / `releaser.dart:209` are unaffected.

So "at most once per release invocation, zero times for patch" holds today — no cache needed.

On placement: one suggestion worth considering is hoisting the fetch up to `resolveTargetFlutterRevision` (release_command.dart:420) instead of burying it inside `getRevisionForVersion`. The accessor name reads like a pure lookup, and a future caller of `ShorebirdFlutter.getRevisionForVersion` would silently pick up a network side-effect. The fetch only actually helps the semver branch of `resolveFlutterRevision` anyway — raw hashes don't benefit. Putting it at the release-flow entry point keeps the side-effect visible and scoped to the "resolve user-supplied version string" intent.

Alternative worth a thought: fix the root cause in `shared.sh` (shorter stamp TTL, or a separate stamp just for `refs/remotes/origin/flutter_release/*`) so no per-command fetch is needed at all.

bdero:
Yes, it's only called once, on release. Moved the call to `resolveTargetFlutterRevision` as suggested.

eseidel (approved):

PR #3696, opened 2026-04-19
feat(shorebird_code_push_protocol): generate models + messages from OpenAPI spec

eseidel:
## Summary

Replace the hand-written model and message DTO files in
`shorebird_code_push_protocol` with output generated by
[space_gen](https://github.com/eseidel/space_gen) from the public
OpenAPI spec served at
<https://api.shorebird.dev/openapi.json>. The spec itself landed
separately via shorebirdtech/_shorebird#2048 (now merged);
`shorebird_code_push_client` keeps its hand-written `CodePushClient`
but consumes the generated model types.

All tests still pass:
- `shorebird_code_push_protocol`: 216/216 (65 hand-written under
  `test/src/` + 151 auto-generated round-trip tests under
  `test/generated/`).
- `shorebird_code_push_client`: 82/82.

## Latent bug fixed along the way

The hand-written `CreateReleaseArtifactRequest` carried
`@JsonSerializable(includeIfNull: false)`, so its `toJson()`
silently dropped null fields. The client's multipart stringifier
did `.toJson().map((k, v) => MapEntry(k, '$v'))` — safe only
because nulls never made it into the map.

The generated `toJson()` correctly returns proper JSON types
(`null` stays `null`), so the same stringifier would turn a null
`podfile_lock_hash` into the four-character string `"null"` on the
wire. The fix (in the second commit on this branch) filters nulls
at the multipart boundary. The two multipart tests are rewritten
to assert against a hard-coded `Map<String, String>` payload —
the old tests round-tripped through `toJson()` on both sides and
so carried the bug both ways, passing while the wire was wrong.

Almost certainly harmless in production (the server probably
rejected or ignored `podfile_lock_hash: "null"` the same way as
absent), but objectively incorrect.

## What's generated

All 14 core models + 26 message DTOs:

| | |
|---|---|
| Models | `App`, `AppMetadata`, `Channel`, `Organization`, `OrganizationMembership`, `OrganizationUser`, `PublicUser`, `PrivateUser`, `Patch`, `PatchArtifact`, `Release`, `ReleaseArtifact`, `ReleasePatch`, `ErrorResponse`, plus enums `ReleasePlatform`, `ReleaseStatus`, `Role`, `AppCollaboratorRole`, `OrganizationType` |
| Messages | `Create/Update/Get{App,Channel,Patch,Release,…}Request/Response`, `PatchCheck{Request,Response,Metadata}`, `CreatePatchResponse`, etc. |

Field types match the hand-written package one-for-one, including
the enum-keyed `Map<ReleasePlatform, ReleaseStatus>` on
`Release.platformStatuses` — expressed via the standard JSON
Schema `propertyNames` keyword.

## What stays hand-written

Three files under `lib/extensions/` cover concerns the OpenAPI spec
cannot express:

- `ReleasePlatformExtensions.{displayName, supportsFlavors}` —
  irregular UI casing (`iOS`, `macOS`) plus a platform predicate.
- `AuthProvider` — enum classifying JWT issuers server-side. Never
  crosses the wire, so correctly excluded from the spec.
- Top-level test helpers `organizationForTest`,
  `privateUserForTest`, `publicUserFromPrivateUser`. Extension
  methods can't add statics to a class, so the former
  `Organization.forTest` / `PublicUser.fromPrivateUser` factories
  pivot to free functions.

`shorebird_code_push_client` stays fully hand-written; it just
consumes the generated types.

## Quality

Ways the generated protocol **exceeds** hand-written:
- No `json_serializable` / `.g.dart` parts → no `build_runner`, no
  `part of` files.
- Drops five dependencies: `json_annotation`, `equatable`,
  `decimal`, `build_runner`, `json_serializable`. Adds one:
  `collection` (used by the generated `model_helpers.dart`).
- Adds `maybeFromJson` on every model.
- Round-trip tests auto-generated for every model, asserting
  `Type.fromJson(instance.toJson()) == instance` and that malformed
  input throws `FormatException`. Enum tests additionally iterate
  every variant and assert `toString() == toJson()`.
- `toJson()` returns proper JSON types rather than a pre-stringified
  multipart hack (which caught the null-field bug above).

Ways it **matches** hand-written: snake-case JSON keys, typed
enums, per-field dartdoc, per-enum-case dartdoc, `DateTime` / `Uri`
from `format` hints, nullability semantics.

Ways it **differs** (all wire-equivalent):
- `@JsonKey(includeIfNull: false)` on `hash_signature` → generated
  emits `null` explicitly. JSON consumers treat absent and null
  identically in the body. Existing golden test updated.
- `ReleaseArtifact.toString()` JSON-rendering dropped. Callers
  should use `artifact.toJson().toString()`. One test deleted.
- `CreatePatchResponse` is now narrower (`{id, number}`) — the
  server never populates `notes` at creation. Matches actual
  runtime behavior.

## Coverage gaps closed (post-codecov)

Two paths surfaced by codecov that the auto-generated round-trip
suite can't reach on its own:

- `ReleasePlatformExtensions.displayName` — hand-written extension,
  no spec representation. Covered by per-value asserts in
  `test/src/models/release_platform_test.dart`.
- `PatchCheckResponse.fromJson`'s
  `(json['rolled_back_patch_numbers'] as List?)?.cast<int>()` —
  the round-trip uses the minimal required-only constructor, so the
  optional list stayed null and the cast stayed cold. Covered by a
  fromJson-side assertion with the field populated.

The third gap — the generated `toString()` override on every enum
— was closed upstream in eseidel/space_gen#96 and pulled in via the
pin bump to `58b4410`; regenerating gave each enum test
`toString matches toJson for every value` plus
`fromJson round-trips every value`.

## Commits

1. `feat(shorebird_code_push_protocol): generate models from OpenAPI spec` — the migration itself, plus the `tool/gen.dart` entrypoint and pubspec cleanup.
2. `fix(shorebird_code_push_client): drop null fields from createReleaseArtifact multipart` — client adapt + the bug fix described above.
3. `chore(shorebird_cli): adapt to regenerated protocol types` — `forTest` → helper sweeps and `.g.dart` regen for the new `ReleasePlatform` shape.
4. `test(shorebird_code_push_protocol): cover displayName + rolledBackPatchNumbers parsing` — hand-written tests for the two non-spec/non-default-round-trip codecov gaps.
5. `chore(shorebird_code_push_protocol): bump space_gen pin + regen enum tests` — picks up eseidel/space_gen#96 so the generated enum tests cover `toString()` and every variant.

## Test plan

- [x] `cd packages/shorebird_code_push_protocol && dart test` — 216/216
- [x] `cd packages/shorebird_code_push_client && dart test` — 82/82
- [x] `dart analyze` in both packages — zero errors/warnings
- [ ] Exercise against a live dev/staging CodePush server before merge — unit tests cover round-trips but not wire compatibility.


nickshorebird (approved):
LGTM - Thanks for this.

PR #3697, opened 2026-04-19
docs(claude): add clean-worktrees skill

eseidel:
## Summary
Adds a Claude Code skill for finding and safely cleaning up stale git worktrees. Lives at `.claude/skills/clean-worktrees/SKILL.md`.

Establishes `.claude/skills/` as the place for shared Claude Code skills in this repo (mirrors the per-user `~/.claude/skills/` layout, so team members can symlink or vendor it directly). Happy to change the location (`skills/` top-level, separate `claude-skills` repo, etc.) if we'd rather.

## Why
Worktrees pile up — especially with agent-based tools that create isolated worktrees per task. The naive cleanup path ("anything with no commits ahead of main is safe") misses three real failure modes we hit this week:

- Squash/rebase merges rewrite SHAs, so merged commits look unmerged to \`git log\`.
- Branches silently back open PRs you haven't seen — deleting them closes someone's in-flight review.
- \`.git/worktrees/<name>/\` mtimes get bumped by read-only commands, so "recently modified" is misleading.

The skill encodes the right checks (subject-match against base, \`gh pr list --search head:\`, direct \`stat\` on the worktree dir) plus the gotchas that took real cleanup accidents to learn.

## Test plan
- [ ] Drop the skill into your \`~/.claude/skills/clean-worktrees/\` (or symlink from this repo), invoke \`/clean-worktrees\`, and verify it finds and categorizes your own stale worktrees before touching anything.

### DIFF
--- a/.claude/skills/clean-worktrees/SKILL.md
+++ b/.claude/skills/clean-worktrees/SKILL.md
@@ -1,0 +1,177 @@
+---
+name: clean-worktrees
+description: |
+  Find and clean up stale git worktrees. Categorizes each as safe-to-remove
+  (clean, zero commits ahead, commits already merged under a different SHA, or
+  branch already backing an open PR) vs needs-review (real unmerged work).
+  Handles common gotchas: misleading admin-dir mtimes, branches silently backing
+  PRs, squash-merged commits, noise in the dirty-file list.
+  Use when asked to "clean up worktrees", "find orphan worktrees", "prune
+  worktrees", or when stale worktree directories pile up.
+---
+
+# clean-worktrees
+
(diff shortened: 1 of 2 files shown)


eseidel:
I'm not sure what we want our convention to be here. If we want these in this repository or our internal monorepo or what? But it seems like we should have a skills directory somewhere.

eseidel:
In any case, this skill successfully cleaned up 100GB+ off my machine tonight :)

tomarra:
Maybe this is just my own mental blocking but I've never found worktree support to be useful as it just gets into this unmaintained mess. I'm typically just having the robot work on branches (either locally or in a cloud environment). I'm fine with whatever you guys choose here given that I wouldn't be using it.

eseidel:
Looks like Nick landed his internally. I'm going to close this for now.

Issue #2270, opened 2024-06-19
Unexpected native changes warning when patching an Android release


bryanoltman:
When running `shorebird patch android`, Discord user `michaelgolfi` is reporting seeing an unexpected native changes warning ([thread](https://discord.com/channels/1030243211995791380/1250070546008379423)):

```
[WARN] Your app contains native changes, which cannot be applied with a patch.
    Changed files:
        base/dex/classes2.dex
```

Their pubspec.yaml file (organized for readability):

```yaml
name: mobile_app
publish_to: none
version: 0.0.1+1

environment:
  sdk: ">=2.18.0 <4.0.0"

dependencies:
  add_2_calendar: 3.0.1
  animated_toggle_switch: 0.8.1
  async: 2.11.0
  back_button_interceptor: 7.0.3
  badges: 3.1.2
  battery_plus: 6.0.1
  bubble_box: 0.5.3
  built_collection: 5.1.1
  built_value: 8.7.0
  calendar_date_picker2: 0.5.3
  collection: 1.18.0
  components:
    path: ../lib/components
  connectivity_plus: 6.0.3
  convert: 3.1.1
  cookie_jar: 4.0.8
  cr_file_saver:
    git:
      ref: 1db62fd53adf9b34d4b769788d6c4be4fa416817
      url: https://github.com/crit-iq/cr_file_saver.git
  crypto: 3.0.3
  device_info_plus: 9.0.2
  easy_rich_text: 2.1.0
  email_validator: 2.1.17
  extended_image: 8.2.0
  facebook_app_events: 0.19.2
  ferry: 0.15.0+1
  ferry_flutter: 0.9.0
  ferry_hive_store: 0.5.2
  firebase_analytics: 11.0.0
  firebase_core: 3.0.0
  firebase_crashlytics: 4.0.0
  firebase_messaging: 15.0.0
  firebase_performance: 0.10.0
  flutter:
    sdk: flutter
  flutter_animation_progress_bar: 2.3.1
  flutter_branch_sdk: 8.0.2
  flutter_facebook_auth: 6.2.0
  flutter_image_compress: 2.3.0
  flutter_localizations:
    sdk: flutter
  flutter_markdown: 0.7.1
  flutter_multi_formatter: 2.12.3
  flutter_svg: 2.0.9
  get_it: 7.7.0
  golden_toolkit: 0.15.0
  google_fonts: 6.1.0
  google_sign_in: 6.2.1
[...]

PR #3698, opened 2026-04-19
fix(cli): detect --obfuscate and --split-debug-info after --

### LINKED ISSUE #3695: link_failure: vm_data_hash mismatch on every patch attempt (iOS, same machine/CLI/Flutter)
## Description

`shorebird patch ios` consistently fails with `link_failure: base and patch snapshots have differing VM sections` immediately after a successful `shorebird release ios` on the same machine, same CLI version, same Flutter version, and same source code.

The `vm_data_hash` values are deterministic but always different between the stored baseline and the locally-built patch. Clean builds (`flutter clean`) do not help.

## Steps to reproduce

1. `shorebird release ios -- --build-number=45 --dart-define=SERVER_URL=... --dart-define=WS_URL=... --dart-define=DISABLE_AUTH=false`
   - Release succeeds, uploads to TestFlight.
2. Make a small Dart-only change (e.g. adjust a widget padding value).
3. `shorebird patch ios --release-version=0.1.6+45 --allow-asset-diffs -- --dart-define=SERVER_URL=... --dart-define=WS_URL=... --dart-define=DISABLE_AUTH=false`
   - Fails with `link_failure`.

Also tried:
- `flutter clean` before patch — same failure.
- With and without `--obfuscate` — same failure. (Shorebird reports the release was built without obfuscation, even though `--obfuscate` was passed to the release command.)
- Patching with zero code changes (identical source to release) — same failure.
- Multiple releases (build 43, 45) — all produce the same `base` hash (`4272422645`), patch always produces `2308514119`.

## Error output

```json
{
  "type": "link_failure",
  "reason": "base and patch snapshots have differing VM sections",
  "details": {
    "dart_version": "3.11.4 (stable) (Thu Mar 26 17:35:43 2026 -0700) on \"macos_simarm64\"",
    "snapshot_version": "7d8fb82a698d78184f7d1e3bbc00540a",
    "vm_data_length": {"base": 39296, "patch": 39296},
    "vm_instructions_length": {"base": 65280, "patch": 65280},
    "vm_data_hash": {"base": 4272422645, "patch": 2308514119},
    "vm_instructions_hash": {"base": 1550369841, "patch": 1550369841}
  }
}
```

Note: `vm_instructions_hash` matches, `vm_data_length` matches, only `vm_data_hash` differs.

## Environment

```
Shorebird 1.6.92
Flutter 3.41.6 • revision 76ca3dff01cc7c5e6978e6b51e442777c6930419
Engine • revision a8578c0205c4f52bc8a0d07c5d75f2a19e208187
macOS 25.2.0 (Darwin), Apple Silicon
```

## Additional context

- The release is run via fastlane with `yes | shorebird release ios -- ...` to auto-confirm prompts.
- App ID: `46f4571f-3393-4e4b-a0e3-6d122946a08c`
- Release ID from logs: `569254`
- The release successfully built and uploaded an IPA to TestFlight + registered with Shorebird's cloud. The patch downloads the xcarchive from Shorebird's storage successfully but fails at the link step.
- `--obfuscate` was passed to the release command, but Shorebird reports "the release was not built with obfuscation" when attempting a patch with `--obfuscate`, suggesting the flag was silently dropped during release.

eseidel:
## Summary

Fixes #3695.

When users pass `--obfuscate` after the `--` separator (as fastlane and other wrappers do when forwarding flags to `flutter build`), the releaser's `useObfuscation` getter only checked `argResults['obfuscate']` and missed it. That caused `shorebird release` to:

- skip `--save-obfuscation-map` / `--strip` in gen_snapshot extra options
- skip uploading the obfuscation-map supplement artifact
- record the release as unobfuscated on the server

…while Flutter still saw `--obfuscate` in the forwarded args and obfuscated the shipped IPA. Every subsequent `shorebird patch` then failed with:

```
link_failure: base and patch snapshots have differing VM sections
vm_data_hash { base: …, patch: … }   // mismatch
vm_instructions_hash { base: …, patch: … }  // matches
```

Classic obfuscation-only mismatch: obfuscation changes string/symbol data (VM data section) but not instruction bytes.

## Fix

Adds two helpers to the existing \`OptionFinder\` extension on \`ArgResults\`, both of which consult parsed args AND \`rest\` (post-\`--\`):

- \`flagPresent(name)\` — boolean flags (\`--obfuscate\`)
- \`optionPresent(name)\` — value-bearing options (\`--split-debug-info\`), matching both \`--name=value\` and \`--name value\` forms, rejecting similar-prefix names like \`--name-extra\`

Applied at:
- \`releaser.useObfuscation\` — the actual bug fix
- \`patch_command\` \`userPassedObfuscate\` / \`hasObfuscate\` — collapses two existing inline checks that already did the same thing
- \`patch_command\` \`hasSplitDebugInfo\` — same class of bug for the companion option

## Test plan

- [x] 11 new unit tests cover \`flagPresent\` (parsed true/false/absent, in-rest, parsed+rest, \`--no-\` in rest) and \`optionPresent\` (parsed, absent, \`=\`-form in rest, space-form in rest, similar-prefix non-match)
- [x] \`dart test\` on the affected suites (\`arg_results_test\`, \`patch_command_test\`, \`ios_releaser_test\`, \`android_releaser_test\`): 127 passing
- [x] \`dart analyze\` clean on all edited files

## Follow-up (not in this PR)

While working on this I noticed \`patcher.dart:288\` uses \`argResults.options.containsAnyOf([buildName, buildNumber])\` which, in production, always matches because \`build-number\` has \`defaultsTo: '1.0'\` — so the \`buildNameAndNumberArgsFromReleaseVersion\` early-return always fires and the build-name/number injection intended by #2270 never runs. Existing tests stub \`options\` to \`[]\` and don't catch it. Worth a separate PR; didn't want to bundle a behavior change with this bugfix.

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -432,11 +432,7 @@ Building with Flutter $flutterVersionString to determine the release version...
     // If the user explicitly passed --obfuscate but the release has no
     // obfuscation map, the patch would be obfuscated against a non-obfuscated
     // release, producing a broken patch.
-    // Also check rest for `-- --obfuscate`, which bypasses the parser but
-    // still flows through forwardedArgs to the Flutter build command.
-    final userPassedObfuscate =
-        (results.wasParsed('obfuscate') && results['obfuscate'] == true) ||
-        results.rest.any((a) => a == '--obfuscate');
+    final userPassedObfuscate = results.flagPresent('obfuscate');
     if (userPassedObfuscate && obfuscationMapFile == null) {
       logger.err(
         '--obfuscate was passed, but the release was not built with '
@@ -475,9 +471,9 @@ Building with Flutter $flutterVersionString to determine the release version...
     // if --obfuscate will be in the build args (from the user or from
     // the obfuscation map injection above) but --split-debug-info is not.
     final hasObfuscate =
-        (results.wasParsed('obfuscate') && results['obfuscate'] == true) ||
+        results.flagPresent('obfuscate') ||
         extraBuildArgs.contains('--obfuscate');
-    final hasSplitDebugInfo = results.wasParsed('split-debug-info');
+    final hasSplitDebugInfo = results.optionPresent('split-debug-info');
     if (hasObfuscate && !hasSplitDebugInfo) {
       extraBuildArgs.add(
         '--split-debug-info=${p.join('build', 'shorebird', 'symbols')}',
--- a/packages/shorebird_cli/lib/src/commands/release/releaser.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/releaser.dart
@@ -96,8 +96,10 @@ abstract class Releaser {
   /// Returns null if no public key is configured.
   Future<String?> getEncodedPublicKey() => argResults.getEncodedPublicKey();
 
-  /// Whether the user is building with obfuscation.
-  bool get useObfuscation => argResults['obfuscate'] == true;
+  /// Whether the user is building with obfuscation. Checks for `--obfuscate`
+  /// both as a parsed flag and after the `--` separator, since it can be
+  /// passed either to Shorebird directly or forwarded to Flutter.
+  bool get useObfuscation => argResults.flagPresent('obfuscate');
 
   /// Path where the obfuscation map is saved during obfuscated builds.
   String get obfuscationMapPath => p.join(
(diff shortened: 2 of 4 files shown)


bdero:
Maybe rather than silently taking over stuff after `--`, we should just fail loudly and tell the user they will explicitly break things by using `--obfuscate` after `--`.

If we get into this business of interpreting stuff after `--`, then we should just commit to it all the way and get rid of `--` entirely.

eseidel:
That's fine.  Although we do already detect at least one flag after --obfuscate.  But I think you're right we could/shoud? just get rid of it.  It likely means adding a lot of flags to our release, since we'd need to support all flags across all flutter versions?

bdero:
Was nervous about the precedent that selectively ignoring `--` sets, but I don't have a better short term alternative for UX.

bdero (approved):

Issue #3700, opened 2026-04-19
Record --dart-define set at release time; warn on patch if it changed


eseidel:
## Background

When `shorebird patch` fails with `link_failure: base and patch snapshots have differing VM sections` and the signature is "VM data hash differs, VM instructions hash matches", the cause is almost always that the patch was built with a different `--dart-define` set than the release. `--dart-define` values become `String.fromEnvironment()` compile-time constants that live in the VM data section of the AOT snapshot, so any difference produces exactly this mismatch.

See #3695 for a concrete incident, and #3699 for the short-term CLI-side hint we just merged, which now surfaces that `--dart-define` / `--obfuscate` mismatch is the likely cause whenever the signature matches.

## Proposal

Record the set of `--dart-define` and `--dart-define-from-file` values used at release time, server-side (in the release metadata or the supplement artifact). At `shorebird patch` time:

1. Read the stored defines from the release.
2. Compare to the current invocation's defines.
3. If they don't match, warn **before** we even run the linker, listing specifically which keys differ (and where values diverge, redacting values if they look sensitive).

That turns a cryptic late-stage link failure into an up-front, actionable diagnostic.

## Open questions

- Storage: inline in the release metadata, or as a supplement artifact like the obfuscation map?
- Privacy: `--dart-define` values sometimes contain secrets. Store hashes + keys rather than raw values? Or raw values but only server-side?
- Scope: just `--dart-define(-from-file)`, or also record `--obfuscate`, `--split-debug-info`, and other VM-affecting flags? (#3698 fixed the obfuscate detection bug but the mismatch is still possible across CLI versions.)

## Related

- #3695 — incident report
- #3698 — fix for `--obfuscate` passed after `--` being silently dropped at release
- #3699 — short-term user-facing hint on link failure

PR #3701, opened 2026-04-19
docs: design notes for build-flag capture at release, verify at patch

eseidel:
From Claude:

**Draft — seeking design review before I write the implementation.**

Design doc at `NOTES_ON_FLAG_CAPTURE.md`.

## What this proposes

Record the set of flags used at \`shorebird release\` time that affect the AOT snapshot, so \`shorebird patch\` can:

1. Auto-apply flags whose values we can safely store (e.g. \`--obfuscate\`, \`--flavor\`, \`--dart-define\`).
2. Require flags whose values we can't store (e.g. \`--split-debug-info\`, \`--target\` — paths).
3. Fail with a specific diff before linking when something's off (turn today's generic VM-section hash mismatch into "your \`--dart-define=SERVER_URL\` value differs").

Follows on from #3695 (incident), #3698 (obfuscate-after-\`--\` drop), #3699 (CLI-side hint), #3700 (this issue).

## What I want review on before implementing

1. **Flag classification.** The doc has a table bucketing every shorebird-recognized flag into \`record value\` / \`record presence only\` / \`don't record\`. Please shout if anything is misclassified. Unknown flags forwarded via \`rest\` default to presence-only.
2. **Dart-define-from-file handling.** Recommending we expand-and-record k/v pairs from the file (consistent with the privacy rule: values are already in the binary). Conservative alternative of keys-only is flagged as an open question — reviewer's call.
3. **Storage.** Explicitly deferred in the doc — three options listed (supplement artifact / release metadata field / dedicated endpoint) with tradeoffs. My lean is the supplement artifact for this PR, but I'd like a cloud-team opinion before writing storage code.
4. **Privacy rule.** "If the value is baked into the shipped binary, recording it is not a new disclosure." Explicit and worth signing off on.

## Plan after review

- Land this design doc.
- Implement the release-side capture and patch-side verifier against a stub storage interface, in a follow-up PR.
- Wire up the chosen storage once (3) above is resolved.
- End-to-end tests only after storage lands.

## Test plan

- [x] CSpell clean on the new doc.
- [x] No code changes in this PR.

### DIFF
--- a/NOTES_ON_FLAG_CAPTURE.md
+++ b/NOTES_ON_FLAG_CAPTURE.md
@@ -61,13 +61,18 @@ only presence.
 
 - Values of `--dart-define=KEY=VALUE` are constant-folded into the AOT snapshot
   (`String.fromEnvironment('KEY')` returns the value as a compile-time const).
-  Any shipped binary already contains every define value. Recording them
-  server-side is not a net new disclosure.
-- Dart-define *keys* usually do NOT survive into the binary — `String.
-  fromEnvironment` is evaluated by the front-end compiler, which substitutes
-  the value and drops the key. So recording keys is a mild new disclosure, but
-  the customer passed the key on the command line at release time knowing it
-  would drive the build — the expectation is low.
+  Any shipped binary contains every define value. Recording them server-side
+  is not a net new disclosure.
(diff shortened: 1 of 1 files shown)


eseidel:
This should turn into a PRD first.  Closing for now.

PR #3709, opened 2026-04-21
fix(shorebird_cli): gate build tracing on resolved version; drop the runtime flag probe

eseidel:
Three commits, stacked. Rebased onto current main (2026-09-11).

## 1. `fix(cli): gate build tracing off for unknown Flutter versions`

`prepareBuildTrace` used `flutterVersion ?? minVersion` to handle revisions that aren't on any `flutter_release/*` branch (dev pins). That fallback is wrong for a feature-enable check: a Shorebird-fork dev revision predating the tracing PR would be treated as new enough, get `--shorebird-trace` forwarded to flutter, and fail argparse hard.

Flip `FlutterSupportConstraint.isSatisfiedBy` to accept `Version?` and require either a real version ≥ the floor **or** an allowlisted revision to satisfy it. Unknown + non-allowlisted now leaves the trace session off.

## 2. `Revert "fix: only pass --shorebird-trace to build commands that support it (#3708)"`

#3708 probed each `flutter build <cmd> -h -v` at runtime to decide whether to pass `--shorebird-trace`. That was a belt-and-suspenders guard around two real bugs:

1. The gate mis-classified unresolved versions (fixed by commit 1).
2. `flutter build aar` / `ios-framework` / `linux` / `macos` / `windows` didn't register the option (fixed by shorebirdtech/flutter#136, **merged** as `e137e182`).

With both fixes in place, every subcommand this CLI invokes accepts the flag when the gate admits tracing, and every subcommand refuses it when the gate rejects tracing — so the per-subcommand runtime probe (one extra flutter-tool startup per release/patch, plus its 8 per-command call-site detours and help-output cache) is dead weight.

## 3. `docs(cli): describe the build-trace gate as it stands today`

The `buildTraceSupportConstraint` comment still described the pre-3.41.7 world where every pin needed allowlisting. The current pin (`e16cf749`, tip of `flutter_release/3.47.2`) satisfies the floor through its resolved version, so only dev revisions need entries. Drops the superseded `c2c56ab2` entry.

## Landing order — still blocked on a pin bump

**Do not merge until `bin/internal/flutter.version` points at a revision that includes shorebirdtech/flutter#136 (`e137e182` or later).** The current pin `e16cf749` predates it, so with the probe removed, `shorebird release macos` / `ios-framework` / `aar` / `linux` / `windows` would crash with `Could not find an option named "--shorebird-trace"` (reproduced in the original test plan, Scenario C).

Two ways the bump can satisfy the gate:
- a new `flutter_release/3.47.x` branch cut at or after `e137e182` — resolves to a version ≥ 3.41.7, nothing else to do; or
- a dev-hash pin — add it to `allowedRevisions`, otherwise tracing is silently off for everyone on that pin.

## Test plan

- [x] `dart test packages/shorebird_cli/test/src/artifact_builder/artifact_builder_test.dart` — 93 pass after rebase (conflicts were all probe-only code that main had grown since April: the `-h -v` fix in #3708's follow-up and its tests).
- [x] `dart analyze` clean on touched files (one pre-existing `comment_references` info in `artifact_builder.dart`, untouched here).

Original end-to-end verification (April, on the then-current pin) is in the edit history; Scenarios A–C still describe the behavior.

https://claude.ai/code/session_01X2aY3kQvT2tgM1mYbN9c7W

### DIFF
--- a/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
@@ -582,10 +582,13 @@ Reason: Exited with code $exitCode.''',
     final flutterVersion = await shorebirdFlutter.resolveFlutterVersion(
       revision,
     );
-    // Treat an unknown version (e.g. a pinned dev revision) as new enough,
-    // matching the pattern used for other version-gated features.
+    // When resolveFlutterVersion returns null (dev revision not on any
+    // flutter_release/* branch), the gate falls through to the allowlist.
+    // That's intentional: a dev pin predating the tracing PR would otherwise
+    // get `--shorebird-trace` passed to a flutter build that doesn't know
+    // the flag, and fail hard.
     final supportsTrace = buildTraceSupportConstraint.isSatisfiedBy(
-      version: flutterVersion ?? buildTraceSupportConstraint.minVersion,
+      version: flutterVersion,
       revision: revision,
     );
     if (!supportsTrace) {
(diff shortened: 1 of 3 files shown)


eseidel:
Will discuss with @bdero  and either close or land.

PR #3713, opened 2026-04-22
fix: download arm64 patch binary on Apple Silicon

### LINKED ISSUE #3410: fix: `patch` is wrong CPU type on Mac ARM64
```
➜  shorebird_cli git:(main) ✗ file ~/.shorebird/bin/cache/artifacts/patch/patch
/Users/eseidel/.shorebird/bin/cache/artifacts/patch/patch: Mach-O 64-bit executable x86_64
➜  shorebird_cli git:(main) ✗ uname -a                                         
Darwin mac.lan 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6020 arm64
```
I suspect that Shorebird simply depends on you having enabled rosetta2.  Which used to be required for Flutter, but may no longer be.

bdero:
## Summary

The \`patch\` binary was always downloaded as \`patch-darwin-x64.zip\` on macOS, requiring Rosetta 2 to be installed. Apple Silicon users without Rosetta got \`ProcessException: Bad CPU type in executable\` when creating patches.

Now uses \`dart:ffi\` \`Abi.current()\` to detect the host architecture and downloads \`patch-darwin-arm64.zip\` on Apple Silicon hosts.

### Changes

- **\`cache.dart\`**: \`PatchArtifact.storageUrl\` checks \`Abi.macosArm64\` to select between \`darwin-arm64.zip\` and \`darwin-x64.zip\`
- **\`cache_test.dart\`**: Added \`storageUrl\` tests for macOS (arm64 + x64), Linux, and Windows

### Dependencies

- shorebirdtech/updater#337 (build arm64 binary in CI) — merged
- shorebirdtech/flutter#129 (upload arm64 asset to GCS) — merged

Fixes #3410

## Test plan

- [x] All 28 cache tests pass
- [x] New tests verify correct artifact URL for each platform/architecture combination


packages/shorebird_cli/lib/src/cache.dart
```diff
@@ -342,7 +344,8 @@ class PatchArtifact extends CachedArtifact {
   String get storageUrl {
     var artifactName = 'patch-';
     if (platform.isMacOS) {
-      artifactName += 'darwin-x64.zip';
+      final macArch = abi.current == Abi.macosArm64 ? 'arm64' : 'x64';
+      artifactName += 'darwin-$macArch.zip';
```
eseidel: Will this work even with old versions of Flutter?   Or is this patch version independent and we'll just grab the latest patch even for old flutters?

eseidel (approved):

packages/shorebird_cli/lib/src/cache.dart
bdero: uhh you're right this is going to break for old versions of Flutter. Just below here, we're grabbing the binary from the engine artifacts (we pull in the latest arm64 binary for new engine builds).

bdero: Maybe simplest/safest to just to fall back to x64 on 404.

eseidel: We also could just gate on flutter version, we have that mechanism in place. 🤷

eseidel:
I think this solution is fine, but choosing a "dynamic" behavior vs. a static one.  I think one downside of this approach is we won't know when to remove the dynamic behavior.

packages/shorebird_cli/lib/src/flutter_version_constraints.dart
```diff
@@ -65,6 +65,27 @@ class FlutterSupportConstraint {
...
+/// oldest supported Shorebird CLI ships with a pin from this set, the
+/// constant — and the `useArm64` branch in `PatchArtifact.storageUrl`
+/// — can be removed.
+const arm64PatchSupportingFlutterRevisions = <String>{
+  // 3.41.8 — current pin (bin/internal/flutter.version).
+  '8e0003bdaef662b7f8f953a018c1dd38b7e2737a',
+  // 3.41.7 — prior pin; users who downgrade with `--flutter-version 3.41.7`
+  // resolve to this SHA.
+  'c2c56ab2d5483bdf86152725342f55ca6faed946',
+};
```
eseidel: Some of the other checks (in this file) I believe are allowed set + min version.  I think you're going to want that?

eseidel (approved):

packages/shorebird_cli/lib/src/flutter_version_constraints.dart
bdero: yup, done.

PR #3717, opened 2026-04-27
feat(shorebird_cli): Add --json output for agentic CLI consumption    

nickshorebird:
## Summary / Reasoning

AI agents overwhelmingly prefer CLI over API, but structured output is a critical gap as agents can't reliably parse human-formatted terminal output. This PR adds `--json `support so agents can programmatically consume CLI results or users can pipe results to `jq` for readability. In this mode data is returned in a UNIX pattern of STDERR only gets shell bootstrap output (flutter --version, pub get) when it runs w/ the final json being returned in STDOUT.

  - Adds a global `--json` flag with a structured envelope (status, data/error, meta) that commands opt into via emitJsonSuccess/emitJsonError helpers
  - Wires `--json` into doctor and flutter versions list as these are the two existing commands that return queryable diagnostic data
  - Redirects shell bootstrap output to stderr when `--json` is passed so piping to `jq` works cleanly
  - Uses enums (JsonStatus, JsonErrorCode) for status and error codes to keep the contract maintainable as more commands adopt `--json`

This is the foundation for the agentic CLI experience (plan: docs/plan_agentic_cli.md). Subsequent PRs will add TTY detection, `--json` on new commands (releases list, patches list), and llms.txt generation.

## Testing
- bin/shorebird --json doctor 2>/dev/null | jq . — valid JSON, all fields present
- bin/shorebird --json flutter versions list 2>/dev/null | jq . — valid JSON, versions array
- bin/shorebird --json --verbose doctor 2>/dev/null | jq . — includes speed_test field
- bin/shorebird --json nonexistent 2>/dev/null | jq . — error envelope with usage_error
- bin/shorebird doctor — unchanged behavior
- bin/shorebird --verbose doctor — unchanged behavior

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/doctor_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/doctor_command.dart
@@ -196,21 +196,21 @@ Android Toolchain
 
     Map<String, dynamic>? speedTest;
     if (results['verbose'] == true) {
-      double? uploadMbps;
-      double? downloadMbps;
+      double? uploadMbPerSec;
+      double? downloadMbPerSec;
       try {
-        uploadMbps = await networkChecker.performGCPUploadSpeedTest();
+        uploadMbPerSec = await networkChecker.performGCPUploadSpeedTest();
       } on Exception {
         // Report as null on failure.
       }
       try {
-        downloadMbps = await networkChecker.performGCPDownloadSpeedTest();
+        downloadMbPerSec = await networkChecker.performGCPDownloadSpeedTest();
       } on Exception {
         // Report as null on failure.
       }
       speedTest = {
-        'upload_mbps': uploadMbps,
-        'download_mbps': downloadMbps,
+        'upload_megabytes_per_sec': uploadMbPerSec,
+        'download_megabytes_per_sec': downloadMbPerSec,
       };
     }
 
--- a/packages/shorebird_cli/lib/src/commands/doctor_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/doctor_command.dart
@@ -5,11 +5,13 @@ import 'package:shorebird_cli/src/android_sdk.dart';
 import 'package:shorebird_cli/src/android_studio.dart';
 import 'package:shorebird_cli/src/doctor.dart';
 import 'package:shorebird_cli/src/executables/executables.dart';
+import 'package:shorebird_cli/src/http_client/http_client.dart';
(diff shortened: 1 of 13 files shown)


packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
easymac, line 328: I think this is still logged with human-readable text when expecting `--json` output? Not sure if this is detrimental in practice, but it might be a good idea to skip the update prompt when commands are ran with `--json`?

easymac, line 227: Same thing here; agents will definitely check the app version. Unsure if this has material harm? But it might be wise to emit a JSON envelope here too

```json
{
    "shorebird_version": "...",
    "flutter_version": "...",
    "engine_revision": "...",
}
```

packages/shorebird_cli/lib/src/commands/doctor_command.dart
```diff
@@ -138,6 +142,98 @@ Android Toolchain
...
+      } on Exception {
+        // Report as null on failure.
+      }
+      try {
+        downloadMbps = await networkChecker.performGCPDownloadSpeedTest();
+      } on Exception {
+        // Report as null on failure.
+      }
+      speedTest = {
+        'upload_mbps': uploadMbps,
```
easymac: Human readable output (line 111) reports these as MB/s. Calling the field "mbps" might cause the agent to erroneously believe the value is in *megabits per second* instead of *megabytes*.

Didn't verify whether the actual speed test returns values in Mbps or MB/s, but this seems out of alignment

easymac (approved):
👍  Looks like a good change. Left a few nits to consider at your discretion.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
nickshorebird: Thanks for the suggestion. Ran tests w/ all flags for the CLI commands and added fixes for each based on your comments.

nickshorebird: Thanks for the suggestion `--version` now outputs json.

packages/shorebird_cli/lib/src/commands/doctor_command.dart
nickshorebird: Thanks, missed this. Changed to be verbose in what it is rather than shorthand.

PR #3719, opened 2026-04-28
fix(cli): xcodeproj_flutter_override_validator should scan .ios/ for Flutter modules

bdero:
## Summary

`XcodeprojFlutterOverrideValidator` (added in #3406) blocks `shorebird release ios-framework` and `shorebird patch ios-framework` on Flutter modules with the message:

```
The /path/to/module/ios/Runner.xcodeproj directory does not exist.

The command you are running must be run within a Flutter app project that supports the iOS platform.
```

Flutter modules (created with `flutter create -t module`, used in add-to-app integrations) **do** have an Xcode project — Flutter generates and maintains it at `.ios/Runner.xcodeproj/project.pbxproj`, not `ios/Runner.xcodeproj`. The validator was hard-coded to look in `ios/`, so it always concluded "no Xcode project here, fail."

This change resolves the right path based on whether the project is a Flutter module:

| Project type | Marker | Path validated |
|---|---|---|
| Standard Flutter app | (no `flutter > module` in pubspec.yaml) | `ios/Runner.xcodeproj/project.pbxproj` |
| Flutter module (add-to-app) | `flutter > module` in pubspec.yaml | `.ios/Runner.xcodeproj/project.pbxproj` |

When neither path exists (e.g. a freshly cloned module before `flutter pub get`, or a project without the iOS platform at all), `validate()` returns no issues. The previous `canRunInCurrentContext` override is removed so `validatePreconditions` no longer hard-fails on missing Xcode projects.

## Repro

In any Flutter module:

```bash
flutter create -t module my_module
cd my_module
shorebird init
shorebird release ios-framework
```

Before this change: hard-fails on the missing `ios/Runner.xcodeproj` validator. After: validator silently scans `.ios/Runner.xcodeproj/project.pbxproj` (or no-ops if it does not yet exist) and the build proceeds.

## Notes

- Module detection follows the existing pattern in `shorebird_env.dart:252` (`pubspec.flutter['module']`).
- Error message dynamically references `ios/` or `.ios/` as appropriate so the validator points at the file the user actually needs to edit.
- A module that also has a leftover `ios/Runner.xcodeproj` from a previous `flutter create` is intentionally **not** scanned in `ios/` — the module's `.ios/` is what `flutter build ios-framework` actually uses.

## Test plan

- [x] Updated `xcodeproj_flutter_override_validator_test.dart`: removed the obsolete `canRunInCurrentContext` group, replaced "returns error if project.pbxproj does not exist" with no-issue cases for missing artifacts, and added a new `Flutter module (.ios/)` group covering:
  - silent skip when `.ios/Runner.xcodeproj/project.pbxproj` is missing
  - clean module pbxproj produces no issues
  - FLUTTER_ROOT in `.ios/Runner.xcodeproj/project.pbxproj` is flagged with the `.ios/` path in the error message
  - leftover `ios/Runner.xcodeproj` in a module project is not scanned (only `.ios/` is)
- [x] `dart test test/src/validators/xcodeproj_flutter_override_validator_test.dart` → 21 / 21 passing.
- [x] `dart test test/src/commands/doctor_command_test.dart test/src/commands/init_command_test.dart test/src/validators/` → 73 / 73 passing.
- [ ] Manual smoke: re-run `shorebird release ios-framework` against a Flutter module fixture and confirm it succeeds.


bdero:
I think the right fix for this would be to run the validator against the xcode project in the correct location for modules/add-to-app. `.ios/Runner.xcodeproj/project.pbxproj`

eseidel (approved):
Thanks.  Sorry for the trouble.

PR #3720, opened 2026-04-28
feat(shorebird_cli): non-interactive (TTY) detection and prompt failure

nickshorebird:
## Summary / Reasoning

#3717 added `--json` so agents can parse CLI output. But several commands still block on interactive prompts, render spinners with carriage returns, and emit ANSI color codes when stdout isn't a real terminal — so an agent invoking them either hangs or gets garbled output. This PR closes those gaps so the CLI is genuinely usable from a non-TTY context.

The trigger is `stdout.hasTerminal`, augmented with two explicit overrides (`--json` and a new `--no-input` flag) so callers can opt into non-interactive behavior even on a real TTY (useful in CI, scripts, and automated tooling).

  - Adds a `--no-input` global flag and an `isInteractive` predicate as the canonical TTY/flags check. `ShorebirdEnv.canAcceptUserInput` is extended to honor both new flags so existing prompt gates (init, release, patch, patch diff checker, ios/macos patchers) automatically respect them
  - Makes `confirm` / `chooseOne` / `prompt` / `promptAny` fail fast when not interactive instead of blocking on stdin. Each call site provides a per-site hint naming the flag that would supply the input (e.g. `--release-version=<version>`, `--allow-native-diffs`, `--no-confirm`). Errors flow through a JSON envelope under `--json` (new `JsonErrorCode.interactivePromptRequired`) or to stderr otherwise; exit code 64
  - Replaces mason_logger's animated spinner with a `_StaticProgress` ("Starting X… / Done X / Failed X") when not interactive. Routed to stderr under `--json` so it doesn't corrupt the envelope
  - Forces ANSI escapes off via `overrideAnsiOutput(false, …)` under `--json` / `--no-input` even on a TTY. The plain non-TTY case is already handled by `package:io`
  - Closes a parse-time gap from #3717: unknown global flags (`shorebird --json --bogus`) and unrecognized commands now emit the JSON error envelope. Also fixes a `Run: shorebird shorebird --help` hint when no subcommand was recognized
  - Uses enums (extends `JsonErrorCode`) and reuses existing primitives (`canAcceptUserInput`, scoped-ref pattern) rather than introducing parallel concepts

## Commits

Squash-friendly, but each is independently reviewable in 10–15 minutes:

  1. `feat: add --no-input flag and isInteractive predicate` — foundation, no behavior change
  2. `feat: fail interactive prompts in non-interactive contexts` — logger overrides + per-site hint migration on 11 call sites; mock matchers updated across 10 test files
  3. `feat: static progress and ANSI suppression in non-interactive mode` — `_StaticProgress` + `overrideAnsiOutput` wiring
  4. `fix: emit JSON error envelopes for parse-time failures` — pre-scan argv for `--json` so the outer `try` block can route errors through the JSON path; also fixes the duplicated-`shorebird` hint
  5. `test: mark untestable lines with coverage:ignore` — diff coverage to 100%; every ignore has a "why" comment naming the test that covers the equivalent logic or why the branch can't be exercised
  6. `fix: correct misleading hint in set-track prompt` — original hint suggested piping `yes`, which doesn't work on this branch (the override throws before reading stdin)

## Testing

- bin/shorebird --version --json 2>/dev/null | jq . — valid JSON, all fields present
- bin/shorebird --json fly_to_the_moon 2>/dev/null | jq . — error envelope, hint `Run: shorebird --help`, exit 64
- bin/shorebird --json --bogus 2>/dev/null | jq . — error envelope (the parse-time fix), exit 64
- bin/shorebird --json patches notreal 2>/dev/null | jq . — error envelope, hint `Run: shorebird patches --help`, exit 64
- bin/shorebird --json doctor 2>/dev/null | jq . — valid JSON, no ANSI escapes on stdout
- bin/shorebird --no-input doctor — runs to completion, no ANSI on stdout
- bin/shorebird doctor | cat — non-TTY: spinners replaced with `Starting X… / Done X` lines, no `\r` or `\x1b`
- bin/shorebird --json --verbose --version 2>/dev/null | jq . — verbose suppressed from stdout, JSON intact
- Every subcommand `--help` (cache, create, doctor, flutter, init, login, login:ci, logout, patch, patches, preview, release, releases, upgrade) — exits 0 with usage text
- dart test — 1896 tests pass (~30 new, covering the `isInteractive` matrix, prompt-failure for all four methods, runner translation, `_StaticProgress` states, ANSI override, and parse-time JSON errors)
- Diff coverage on `lib/` — 100%

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -158,7 +158,12 @@ This may indicate that the patch contains native changes, which cannot be applie
           throw UnpatchableChangeException();
         }
 
-        if (!logger.confirm('Continue anyway?')) {
+        if (!logger.confirm(
+          'Continue anyway?',
+          hint:
+              'Pass --allow-native-diffs to acknowledge native changes '
+              'without prompting.',
+        )) {
           throw UserCancelledException();
         }
       }
--- a/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
@@ -160,9 +160,7 @@ This may indicate that the patch contains native changes, which cannot be applie
 
         if (!logger.confirm(
           'Continue anyway?',
-          hint:
-              'Pass --allow-native-diffs to acknowledge native changes '
-              'without prompting.',
+          hint: allowNativeDiffsHint,
         )) {
           throw UserCancelledException();
         }
--- a/packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
@@ -116,9 +116,7 @@ This may indicate that the patch contains native changes, which cannot be applie
 
         if (!logger.confirm(
           'Continue anyway?',
-          hint:
-              'Pass --allow-native-diffs to acknowledge native changes '
-              'without prompting.',
+          hint: allowNativeDiffsHint,
         )) {
           throw UserCancelledException();
         }
--- a/packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
@@ -114,7 +114,12 @@ This may indicate that the patch contains native changes, which cannot be applie
           throw UnpatchableChangeException();
         }
 
-        if (!logger.confirm('Continue anyway?')) {
+        if (!logger.confirm(
+          'Continue anyway?',
+          hint:
+              'Pass --allow-native-diffs to acknowledge native changes '
+              'without prompting.',
+        )) {
           throw UserCancelledException();
         }
       }
--- a/packages/shorebird_cli/lib/src/interactive_mode.dart
+++ b/packages/shorebird_cli/lib/src/interactive_mode.dart
@@ -7,7 +7,11 @@ import 'package:shorebird_cli/src/json_output.dart';
 final isNoInputModeRef = create(() => false);
 
 /// Whether non-interactive mode (`--no-input`) is active in the current zone.
-bool get isNoInputMode => read(isNoInputModeRef);
+///
+/// Defaults to `false` outside of any scope that has overridden the ref --
+/// callers that read this from non-runner contexts should not be forced to
+/// set up the scoped value.
+bool get isNoInputMode => read(isNoInputModeRef, orElse: () => false);
 
 /// Whether the CLI is running in an interactive context.
 ///
@@ -21,3 +25,39 @@ bool get isNoInputMode => read(isNoInputModeRef);
 ///   * `--no-input` was passed (explicit non-interactive opt-in).
 bool get isInteractive =>
     io.stdout.hasTerminal && !isJsonMode && !isNoInputMode;
+
+/// The default hint emitted when an interactive prompt is reached in a
+/// non-interactive context and no per-site hint was provided.
+const String defaultInteractivePromptHint =
+    'Re-run with a TTY-attached stdout, or provide the required input via a '
+    'command-line flag. Pass --no-input to make this a hard error in scripts.';
+
+/// {@template interactive_prompt_required_exception}
+/// Thrown when a `confirm`/`chooseOne`/`prompt`/`promptAny` call is reached
+/// while the CLI is in a non-interactive context (no TTY, `--json`, or
+/// `--no-input`).
+///
+/// The runner catches this exception and emits either a JSON error envelope
+/// (under `--json`) or a verbose human-readable error to stderr.
+/// {@endtemplate}
+class InteractivePromptRequiredException implements Exception {
+  /// {@macro interactive_prompt_required_exception}
+  const InteractivePromptRequiredException({
+    required this.promptText,
+    required this.hint,
+  });
+
+  /// The text of the prompt that would have been shown to the user.
+  final String promptText;
+
+  /// An actionable recovery hint -- typically the flag the caller could pass
+  /// to provide the required input non-interactively.
+  final String hint;
+
+  @override
+  String toString() =>
+      'Interactive input was required but the CLI is running in a '
+      'non-interactive context.\n'
+      'Prompt: $promptText\n'
+      'Hint: $hint';
+}
--- a/packages/shorebird_cli/lib/src/interactive_mode.dart
+++ b/packages/shorebird_cli/lib/src/interactive_mode.dart
@@ -4,14 +4,13 @@ import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/json_output.dart';
 
 /// A reference to whether non-interactive mode (`--no-input`) is active.
-final isNoInputModeRef = create(() => false);
+final isNoInputModeRef = create<bool>(
+  () => throw StateError('isNoInputModeRef accessed outside a runner scope'),
+);
 
 /// Whether non-interactive mode (`--no-input`) is active in the current zone.
-///
-/// Defaults to `false` outside of any scope that has overridden the ref --
-/// callers that read this from non-runner contexts should not be forced to
-/// set up the scoped value.
-bool get isNoInputMode => read(isNoInputModeRef, orElse: () => false);
+/// Note: `--json` implies `--no-input`.
+bool get isNoInputMode => read(isNoInputModeRef);
 
 /// Whether the CLI is running in an interactive context.
 ///
--- a/packages/shorebird_cli/lib/src/interactive_mode.dart
+++ b/packages/shorebird_cli/lib/src/interactive_mode.dart
@@ -1,16 +1,6 @@
 import 'dart:io' as io;
 
-import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/json_output.dart';
-
-/// A reference to whether non-interactive mode (`--no-input`) is active.
-final isNoInputModeRef = create<bool>(
-  () => throw StateError('isNoInputModeRef accessed outside a runner scope'),
-);
-
-/// Whether non-interactive mode (`--no-input`) is active in the current zone.
-/// Note: `--json` implies `--no-input`.
-bool get isNoInputMode => read(isNoInputModeRef);
 
 /// Whether the CLI is running in an interactive context.
 ///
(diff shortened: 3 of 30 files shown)


packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
```diff
@@ -158,7 +158,12 @@ This may indicate that the patch contains native changes, which cannot be applie
           throw UnpatchableChangeException();
         }
 
-        if (!logger.confirm('Continue anyway?')) {
+        if (!logger.confirm(
+          'Continue anyway?',
+          hint:
+              'Pass --allow-native-diffs to acknowledge native changes '
+              'without prompting.',
```
eseidel: I'd be cautious about encouraging people to bypass these. Anytime they're allowing native diffs, they may be causing a crash in their app.

packages/shorebird_cli/lib/src/interactive_mode.dart
```diff
@@ -0,0 +1,63 @@
...
+/// Whether the CLI is running in an interactive context.
+///
+/// This is the canonical predicate for behaviors that require a TTY:
+/// progress spinners, ANSI color output, and interactive prompts
+/// (`chooseOne`/`confirm`/`prompt`).
+///
+/// Returns `false` when any of the following is true:
+///   * stdout is not connected to a terminal (`!stdout.hasTerminal`).
+///   * `--json` was passed (machine-readable output expected).
+///   * `--no-input` was passed (explicit non-interactive opt-in).
```
eseidel: Why both --json and --no-input?  When do the agents use one vs. the other?  Should --no-input simply imply --json?

packages/shorebird_cli/lib/src/json_output.dart
```diff
@@ -9,7 +9,11 @@ import 'package:shorebird_cli/src/version.dart';
 final isJsonModeRef = create(() => false);
 
 /// Whether JSON output mode is active in the current zone.
-bool get isJsonMode => read(isJsonModeRef);
+///
+/// Defaults to `false` outside of any scope that has overridden the ref --
+/// callers that read this from non-runner contexts (e.g. unit tests for
+/// `ShorebirdEnv`) should not be forced to set up the scoped value.
+bool get isJsonMode => read(isJsonModeRef, orElse: () => false);
```
eseidel: Having a default here seems a bit dangerous.  Seems like it would be very easy to accidentally not plumb this down where we expect it to be and then have it use the default?

packages/shorebird_cli/lib/src/patch_diff_checker.dart
```diff
@@ -117,7 +117,12 @@ If you don't know why you're seeing this error, visit our troubleshooting page a
           throw UnpatchableChangeException();
         }
 
-        if (!logger.confirm('Continue anyway?')) {
+        if (!logger.confirm(
+          'Continue anyway?',
+          hint:
+              'Pass --allow-native-diffs to acknowledge native changes '
+              'without prompting.',
```
eseidel: Given we have the same string in multiple places, maybe we should try to share it.

packages/shorebird_cli/lib/src/patch_diff_checker.dart
```diff
@@ -156,7 +161,12 @@ If you don't know why you're seeing this error, visit our troubleshooting page a
           throw UnpatchableChangeException();
         }
 
-        if (!logger.confirm('Continue anyway?')) {
+        if (!logger.confirm(
+          'Continue anyway?',
+          hint:
+              'Pass --allow-asset-diffs to acknowledge asset changes '
+              'without prompting.',
```
eseidel: Same thing here.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
```diff
@@ -239,10 +283,18 @@ Engine • revision ${shorebirdEnv.shorebirdEngineRevision}''');
         }
       } on UsageException catch (e) {
         if (isJsonMode) {
+          // Avoid "Run: shorebird shorebird --help" when no subcommand was
```
eseidel: `shorebird shorebird`?

packages/shorebird_cli/lib/src/commands/patch/ios_patcher.dart
nickshorebird: Changed wording to be prefaced with a Warning message before the command to skip. I can remove the hint altogether if wanted and just return the warning.

packages/shorebird_cli/lib/src/interactive_mode.dart
nickshorebird: In my mind I viewed `jq` usage as a different flow, but you are piping anyways. Made both of these uniform.

packages/shorebird_cli/lib/src/json_output.dart
nickshorebird: Thanks for catching this - Changed the default to throw rather than defaulting to false.

packages/shorebird_cli/lib/src/patch_diff_checker.dart
nickshorebird: Switched to var string usage to reduce boilerplate.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
nickshorebird: Fixed comment and the possibility of the hint to show shorebird shorebird.

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
```diff
@@ -114,7 +114,12 @@ This may indicate that the patch contains native changes, which cannot be applie
           throw UnpatchableChangeException();
         }
 
-        if (!logger.confirm('Continue anyway?')) {
+        if (!logger.confirm(
+          'Continue anyway?',
+          hint:
+              'Pass --allow-native-diffs to acknowledge native changes '
+              'without prompting.',
```
eseidel: allowNativeDiffsHint?

packages/shorebird_cli/lib/src/json_output.dart
```diff
@@ -6,22 +6,26 @@ import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/version.dart';
 
 /// A reference to whether JSON output mode is active.
-final isJsonModeRef = create(() => false);
+final isJsonModeRef = create<bool>(
+  () => throw StateError('isJsonModeRef accessed outside a runner scope'),
```
eseidel: I don't think this is needed.  I think it will automatically do this if the ref isn't made available in scope?

packages/shorebird_cli/lib/src/release_chooser.dart
```diff
@@ -98,5 +101,8 @@ Release chooseRelease({
     prompt,
     choices: sorted,
     display: display,
+    hint:
+        'Pass --release-version=<version> to $action a specific release '
+        'without prompting.',
```
eseidel, line 108: I would ask your claude to evaluate this change for DRY, and tell it to make all of these shared strings contstants like you already have started.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
```diff
@@ -138,28 +154,57 @@ class ShorebirdCliCommandRunner extends CompletionCommandRunner<int> {
...
+      // FormatException from `parse(args)` is rare in practice; the JSON
+      // branch is hard to trigger from real argv.
+      // coverage:ignore-start
+      if (jsonModeFromArgs) {
+        JsonResult.error(
+          code: JsonErrorCode.usageError,
+          message: e.message,
+          hint: 'Run: shorebird --help',
+          command: executableName,
+        ).write();
```
eseidel, line 176: I'm confused why this needs special handling?  Is the idea that you want the error itself to be json?  If so, should we be handling that genericly?

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
```diff
@@ -253,6 +302,25 @@ Engine • revision ${shorebirdEnv.shorebirdEngineRevision}''');
...
         // sure" message, so we do an early return here.
         return ExitCode.usage.code;
+      } on InteractivePromptRequiredException catch (e) {
+        if (isJsonMode) {
+          JsonResult.error(
+            code: JsonErrorCode.interactivePromptRequired,
+            message: e.promptText,
+            hint: e.hint,
+            command: commandName,
+          ).write();
```
eseidel, line 303: This feels like a helper.

packages/shorebird_cli/lib/src/commands/patch/macos_patcher.dart
nickshorebird: Will switch to shared constants allowNativeDiffsHint and allowAssetDiffsHint defined at the top of patch_diff_checker.dart.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
nickshorebird: The FormatException path is only reachable if `parse(args)` throws before we've set up the scoped refs so `isJsonMode` isn't available yet. That is why jsonModeFromArgs is used instead. It's a narrow pre-scope edge case, hence the coverage:ignore hard to trigger from real argv. Having the error print out to STDERR would break agent that pipe `2>/dev/null | ` due to silent errors.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
nickshorebird: Fair - I opted to not have it as a helper to not extract for the sake of abstraction. It is only called here, is readable and hit by tests for both paths.

packages/shorebird_cli/lib/src/release_chooser.dart
nickshorebird: Will extract to releaseVersionHint constant at the top of release_chooser.dart after an audit by claude.

packages/shorebird_cli/lib/src/json_output.dart
nickshorebird: Confirmed it does thus making this redundant. Will remove `orElse => false` as no default is needed.

packages/shorebird_cli/lib/src/interactive_mode.dart
nickshorebird: Dropped `--no-input` entirely. Terraform and others do exactly this: terraform apply --json implies --input=false

eseidel (approved):
I think this is a good start. Get CI to pass and address the open concerns, and you're good to land.

PR #3721, opened 2026-04-28
feat: add shorebird_ci package

eseidel:
## Summary

Adds the `shorebird_ci` package — open-source CI tooling for Dart
and Flutter monorepos on GitHub Actions. Scans a repo, generates a
dynamic two-stage workflow, and verifies CI coverage stays in sync
with the Dart dependency graph. Designed to be used by both humans
and AI agents (Claude can run the CLI and read its output).

### Commands

| Command | Purpose |
|---|---|
| `generate` | Emit `.github/workflows/shorebird_ci.yaml` — a dynamic two-stage workflow (setup job computes affected packages via the Dart dep graph, matrix fans out per package). |
| `verify` | Fail if any package is missing CI coverage (use `--ignore <name>,…` to skip packages that intentionally have no CI). |
| `affected_packages` | List packages affected by a PR diff, including transitive dependents via `path:` deps and Dart workspaces. Supports `--sdk dart\|flutter`. |
| `flutter_version` | Resolve a Flutter version from `environment.flutter` constraints (workaround for `subosito/flutter-action` requiring exact versions). |
| `update_actions` | Bump `uses:` action versions in workflow files to the latest major; also bootstraps `.github/dependabot.yml` if missing. |

The generated workflow auto-detects Dart vs. Flutter, Dart workspaces,
codecov, cspell, nested subpackages, `bloc_lint`, integration tests,
and pinned Flutter versions. Adding or removing packages requires no
workflow changes — the setup job discovers them at runtime.

`generate --style static` is also available for repos that already
use a `dorny/paths-filter`-based workflow and just want help keeping
filter paths in sync with the dep graph.

## Test plan

- [x] `dart analyze --fatal-warnings --fatal-infos lib bin test` passes
- [x] `dart format` clean
- [x] `dart test` — 136 tests pass
- [x] `shorebird_ci generate --repo-root . --dry-run` produces a valid workflow YAML
- [x] `shorebird_ci affected_packages --sdk dart` correctly identifies affected packages with transitive deps
- [x] `shorebird_ci verify` correctly detects stale path filters

### DIFF
--- a/packages/shorebird_ci/lib/src/affected_packages.dart
+++ b/packages/shorebird_ci/lib/src/affected_packages.dart
@@ -1,0 +1,68 @@
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:shorebird_ci/src/flutter_version_resolver.dart';
+import 'package:shorebird_ci/src/repository_analyzer.dart';
+
+/// Computes the list of package metadata maps consumed by the dynamic
+/// CI workflow's `matrix.include`.
+///
+/// Each entry has the shape documented on `AffectedPackagesCommand`.
+/// [sdkFilter] may be `'dart'`, `'flutter'`, or `null` (no filter).
+List<Map<String, Object?>> affectedPackagesMetadata({
+  required Directory repoRoot,
+  String? baseRef,
+  String? headRef,
+  String? sdkFilter,
+  bool all = false,
+  RepositoryAnalyzer? analyzer,
+}) {
+  final a = analyzer ?? RepositoryAnalyzer();
+  final repository = a.analyze(repositoryRoot: repoRoot);
+  if (repository.packages.isEmpty) return const [];
+
+  var packages = all
+      ? repository.packages
+      : a
+            .affectedPackages(
+              repository: repository,
+              baseRef: baseRef ?? 'origin/main',
+              headRef: headRef ?? 'HEAD',
+            )
+            .toList();
+
+  if (sdkFilter != null) {
+    packages = packages.where((pkg) {
+      final isFlutter = RepositoryAnalyzer.dependsOnFlutter(root: pkg.root);
+      return sdkFilter == 'flutter' ? isFlutter : !isFlutter;
+    }).toList();
+  }
+
+  final sorted = packages.toList()..sort((a, b) => a.name.compareTo(b.name));
+
+  return sorted.map((pkg) {
+    final isFlutter = RepositoryAnalyzer.dependsOnFlutter(root: pkg.root);
+    final subpackages =
+        RepositoryAnalyzer.subpackages(
+            package: pkg,
+          ).map((sub) => p.relative(sub.rootPath, from: pkg.rootPath)).toList()
+          ..sort();
+
+    return {
+      'name': pkg.name,
+      'path': p.relative(pkg.rootPath, from: repoRoot.path),
+      'sdk': isFlutter ? 'flutter' : 'dart',
+      'flutter_version': isFlutter
+          ? (resolveFlutterVersion(packagePath: pkg.rootPath) ?? '')
+          : '',
+      'has_bloc_lint': RepositoryAnalyzer.dependsOnBlocLint(root: pkg.root),
+      'has_integration_tests': RepositoryAnalyzer.hasIntegrationTests(
+        root: pkg.root,
+      ),
+      // Space-separated so the matrix job can iterate with a shell for
+      // loop. Safe because RepositoryAnalyzer rejects any package path
+      // containing shell metacharacters; see _requireSafePath there.
+      'subpackages': subpackages.join(' '),
+    };
+  }).toList();
+}
--- a/packages/shorebird_ci/lib/src/affected_packages.dart
+++ b/packages/shorebird_ci/lib/src/affected_packages.dart
@@ -1,6 +1,5 @@
 import 'dart:io';
 
-import 'package:path/path.dart' as p;
 import 'package:shorebird_ci/src/flutter_version_resolver.dart';
 import 'package:shorebird_ci/src/repository_analyzer.dart';
 
@@ -43,14 +42,14 @@ List<Map<String, Object?>> affectedPackagesMetadata({
   return sorted.map((pkg) {
     final isFlutter = RepositoryAnalyzer.dependsOnFlutter(root: pkg.root);
     final subpackages =
-        RepositoryAnalyzer.subpackages(
-            package: pkg,
-          ).map((sub) => p.relative(sub.rootPath, from: pkg.rootPath)).toList()
+        RepositoryAnalyzer.subpackages(package: pkg)
+            .map((sub) => posixRelative(sub.rootPath, from: pkg.rootPath))
+            .toList()
           ..sort();
 
     return {
       'name': pkg.name,
-      'path': p.relative(pkg.rootPath, from: repoRoot.path),
+      'path': posixRelative(pkg.rootPath, from: repoRoot.path),
       'sdk': isFlutter ? 'flutter' : 'dart',
       'flutter_version': isFlutter
           ? (resolveFlutterVersion(packagePath: pkg.rootPath) ?? '')
--- a/packages/shorebird_ci/lib/src/commands/generate_command.dart
+++ b/packages/shorebird_ci/lib/src/commands/generate_command.dart
@@ -1,0 +1,685 @@
+import 'dart:io';
+
+import 'package:args/command_runner.dart';
+import 'package:path/path.dart' as p;
+import 'package:shorebird_ci/src/commands/repo_root_option.dart';
+import 'package:shorebird_ci/src/dependency_resolver.dart';
+import 'package:shorebird_ci/src/flutter_version_resolver.dart';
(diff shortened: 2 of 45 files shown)


bdero (commented):
_bdero's bot review_ — re-review of #2041's revised version. Most of the prior critical issues are resolved (sentinel marker for verify, separator-correct submodule pruning, dep cycle guard, name-collision check, `_requireSafePath` mitigating shell injection, `replaceAllMapped` for action rewrites, `updateActionVersions` removed from `generate`). Two remaining critical items in inline comments, plus a suggestion and a nit.

Side note unrelated to a specific line: the PR description says `verify` supports `--apply` to fix dorny filters, but I don't see the flag in `verify_command.dart` (only `--ignore`). Either the description is stale or the flag was dropped during revision — worth aligning before merge.

packages/shorebird_ci/lib/src/commands/generate_command.dart
```diff
@@ -0,0 +1,685 @@
...
+      );
+    }
+
+    buffer.write('''
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: dart-lang/setup-dart@v1
+      - run: dart pub global activate shorebird_ci
```
bdero, line 518: _bdero's bot review:_

🚨 **Critical: workflow installs from pub.dev but the package is `publish_to: none`.**

Generated workflows include `dart pub global activate shorebird_ci`, but `pubspec.yaml:7` is `publish_to: none`. So every generated workflow is broken at the install step the moment a user runs `generate` and pushes — the package can't be fetched from pub.dev because it's not there. Same line at `generate_command.dart:209` (static main).

Two consistent options:
1. Drop `publish_to: none`, publish 0.1.0, and pin (`activate shorebird_ci 0.1.0`) so consumers aren't tracking a moving target.
2. Switch the activation to `dart pub global activate --source git https://github.com/shorebirdtech/shorebird.git --git-path packages/shorebird_ci --git-ref <sha>`. Slower but works without publishing.

Whichever you pick, this PR shouldn't merge with the generated workflow pointing at an install path that doesn't resolve.

packages/shorebird_ci/lib/src/repository_analyzer.dart
```diff
@@ -0,0 +1,399 @@
...
+  /// Allowed characters in a package or subpackage path: letters,
+  /// digits, `_`, `-`, `.`, and `/`. Any other character (whitespace,
+  /// shell metacharacters, Unicode) is rejected.
+  ///
+  /// The generated workflow word-splits matrix entries via
+  /// `for sub in ${{ matrix.subpackages }}; do ... done` — a path with
+  /// shell metacharacters would otherwise be injected directly into
+  /// the runner's shell. Restricting paths to a portable POSIX subset
+  /// makes the simple loop safe by construction.
+  static final _safePathRegex = RegExp(r'^[A-Za-z0-9._/-]+$');
```
bdero, line 132: _bdero's bot review:_

🚨 **Critical: Windows portability — `_requireSafePath` rejects every path on Windows.**

`_safePathRegex = RegExp(r'^[A-Za-z0-9._/-]+$')` only allows forward slashes. The paths fed in are produced by `p.relative` (e.g. `analyze()` line 78, `subpackages()` line 117, `affected_packages.dart:53` and `:42`), which uses platform-native separators — so on Windows the inputs are `packages\foo` and the regex throws `FormatException` for every package.

Even if the regex passed, the `p.relative` output would also be wrong downstream: it lands in YAML as `working-directory: ${{ matrix.path }}` and `package_path: ${{ inputs.package_path }}`, which run on Linux runners where `\` separators don't resolve.

Fix once, in two places:
- At each `p.relative` call site, normalize: `p.relative(...).replaceAll(r'\', '/')` (or use `p.posix.joinAll(p.split(...))`).
- Then the regex stays as-is and validates the post-normalization path.

If the intent is "Linux/macOS only," fail-fast in `bin/shorebird_ci.dart` with a clear "Windows not supported" message — don't dribble out platform-specific FormatExceptions per package.

packages/shorebird_ci/lib/src/affected_packages.dart
```diff
@@ -0,0 +1,68 @@
...
+          ? (resolveFlutterVersion(packagePath: pkg.rootPath) ?? '')
+          : '',
+      'has_bloc_lint': RepositoryAnalyzer.dependsOnBlocLint(root: pkg.root),
+      'has_integration_tests': RepositoryAnalyzer.hasIntegrationTests(
+        root: pkg.root,
+      ),
+      // Space-separated so the matrix job can iterate with a shell for
+      // loop. Safe because RepositoryAnalyzer rejects any package path
+      // containing shell metacharacters; see _requireSafePath there.
+      'subpackages': subpackages.join(' '),
```
bdero, line 64: _bdero's bot review:_

💡 **Suggestion: switch `subpackages` to a JSON array.**

The comment correctly notes that `_requireSafePath` makes the space-join + `for sub in $X` pattern safe today. That's input validation as defense-in-depth, which is good — but the contract between this file and `generate_command.dart:1242` (and `:351`/`:428`) is implicit: weakening the regex, adding a new path source that bypasses the analyzer, or someone refactoring `_requireSafePath` to be less strict would silently re-open the injection class.

Emitting a `List<String>` here, threading it through the matrix as JSON, and consuming via `fromJSON()` into a sub-matrix (or even a `for sub in ${subpackages_array[@]}` with `jq -r '.[]'`) closes the issue structurally — it stops being a question of input validation and starts being a question of data shape. Not a blocker, but worth doing while the design is still mutable.

packages/shorebird_ci/lib/src/repository_analyzer.dart
```diff
@@ -0,0 +1,399 @@
...
+    final packageDescriptions = packages
+        .map((pubspec) {
+          final name = packageName(root: pubspec.parent);
+          if (name == null) return null;
+          final relative = p.relative(
+            pubspec.parent.path,
+            from: repositoryRoot.path,
+          );
+          _requireSafePath(relative);
+          return PackageDescription(
```
bdero, line 82: _bdero's bot review:_

🧹 **Nit: package `name:` from pubspec isn't validated for YAML/shell safety.**

`_requireSafePath` rightly defends path values, but the package's `name:` field flows into the generator unchanged and ends up as a YAML key (`generate_command.dart:209`, `:264`) and a job ID. Pub's own conventions enforce `[a-z][a-z0-9_]*`, but `pubspec.yaml` is just a YAML file — a malformed `name: not valid$` would slip through.

Cheap consistency fix: add a `_requireSafePackageName` modeled on `_requireSafePath` (e.g. `^[a-z][a-z0-9_]*$`) and call it next to the duplicate check at line 90. Keeps the analyzer as the single boundary that validates anything before it leaves the function.

packages/shorebird_ci/lib/src/commands/generate_command.dart
eseidel: Fixed in e2b001c5 by dropping `publish_to: none`. The package will be published to pub.dev separately; this PR just unblocks that step.

packages/shorebird_ci/lib/src/repository_analyzer.dart
eseidel: Fixed in e2b001c5. Added a `posixRelative` helper and applied it at every emission boundary (analyzer, affected_packages, generate_command, verify_command). The validation regex stays POSIX-only — paths are now normalized before they hit it.

packages/shorebird_ci/lib/src/repository_analyzer.dart
eseidel: Fixed in e2b001c5. Added `_requireSafePackageName` (`^[a-z][a-z0-9_]*$` — pub's own convention) called from `analyze()` next to the duplicate-name check, with a corresponding test.

packages/shorebird_ci/lib/src/affected_packages.dart
eseidel: Discussed this on the previous round of review and decided to stay with input validation rather than restructure the matrix data shape. The reasoning:

- The threat model for a Dart/Flutter monorepo CI tool is narrow. GitHub's "first-time contributor approval" gate is on by default for public repos, and the runner's `GITHUB_TOKEN` on `pull_request` from forks is read-only. So the realistic attack window is a returning contributor adding a path with shell metacharacters — which `_requireSafePath` rejects loudly at `shorebird_ci verify` time, before any matrix step runs.
- The JSON-array shape pushes complexity into every consumer site: env-var indirection, `jq` parsing, escaped `"$sub"` in every install step. We considered it and it ends up being noticeably uglier YAML for everyone using the tool, in exchange for closing a window the validation already closes.
- The contract you're worried about (someone weakens the regex or adds a new path source that bypasses the analyzer) is centralized in one file (`repository_analyzer.dart`). A future change there is reviewable as a single diff.

Acknowledging the tradeoff — defense-in-depth via shape vs. defense via boundary validation. We picked boundary validation. Happy to revisit if a concrete bypass appears.

nickshorebird (approved):
LGTM - Thanks for this.

PR #3724, opened 2026-04-29
feat: add cli_io package and migrate from mason_logger

eseidel:
## Summary

- Adds `packages/cli_io/`, a implementation of the subset of `mason_logger`'s API that Shorebird uses. Modeled after mason_logger so existing call sites only change their import line.
- Migrates 112 `.dart` files (`shorebird_cli`, `flutter_version_resolver`, plus integration tests) from `package:mason_logger/mason_logger.dart` to `package:cli_io/cli_io.dart`.
- Drops `mason_logger: ^0.3.5` from `shorebird_cli` and `flutter_version_resolver`; adds `cli_io` as a path dependency.

## Notable behavior change

`chooseOne` is now a numbered prompt (`Enter selection (1-3): _`) instead of arrow-key navigation. No FFI, works over SSH and in CI. Three production call sites (`release_chooser.dart`, `init_command.dart`, `preview_command.dart`); existing tests pass without modification.

cli_io also omits APIs we don't use: `chooseAny`, `promptAny`, `alert`, `delayed`, `flush`, `LogTheme` configuration, hidden-input prompts, and the raw-mode keyboard primitives. `ExitCode` and ANSI constants are inlined rather than re-exported from `package:io`.

## Test plan

- [x] `dart pub get` resolves the workspace cleanly
- [x] `dart analyze --fatal-warnings` on `cli_io`, `shorebird_cli`, `flutter_version_resolver` — 0 errors / 0 warnings (44 info lints, down from baseline 51 — `dart fix --apply --code=directives_ordering` cleaned up imports)
- [x] `dart test` in `packages/cli_io` — 49/49 pass
- [x] `dart test` in `packages/shorebird_cli` — 1863 pass (1 pre-existing skip)
- [x] `dart test` in `packages/flutter_version_resolver` — 8/8 pass
- [ ] CI runs green


eseidel:
Wrote this for you @nickshorebird, will require some manual testing to make sure it still feels good, but then we can improve it to our liking from here.

nickshorebird:
Thanks @eseidel  - Will give this a review + merge after #3720 is merged to not require a re-base and redo of logic.

eseidel:
Removing the fancy chooseOne support also fixes things like https://github.com/felangel/mason/issues/1356

felangel:
PRs to mason logger are always welcome fwiw

felangel:
> Removing the fancy chooseOne support also fixes things like [felangel/mason#1356](https://github.com/felangel/mason/issues/1356)

Sad to see this approach taken instead of contributing bug fixes upstream to mason_logger 😢

eseidel:
This stemmed out of confusion by an engineer about how to work around
issues with our logger. Specifically disabling progress display when in
json/no-input mode. It was easiest for them to add code at our layer rather
than in the logger. Making me realize that it would be simpler for all to
remove the external dependency for something smaller and fit to our purpose
to do logging.  Logging isn’t an area we Shorebird is trying to
innovate, so just trying to do something simple.

I have no plans for us to distribute this package to others, just us moving
off of an external dependency for simplicity within our codebase.  It just
so happens that the bulk of our code is open source so others see this
small internal refactoring rather than it being silent. 🤷🏼‍♂️

felangel:
> This stemmed out of confusion by an engineer about how to work around
issues with our logger. Specifically disabling progress display when in
json/no-input mode.

mason_logger should do that already and if didn’t in your case an issue would’ve been great. Either way it’s your call. It’s a bit of a shame because this feels like it was an opportunity to collaborate rather than to diverge 🤷

packages/cli_io/lib/src/logger.dart
felangel, line 1: This looks identical to the mason_logger implementation. If you copied the code and tweaked it slightly then you should maintain the original license and this should probably go into third_party. A simpler approach imo would’ve been a fork that removed the interactive elements 🤷

packages/cli_io/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Lineage
+
+The public API is modeled after [`mason_logger`][mason_logger], which
+Shorebird depended on historically. When we replaced that dependency with our
+own implementation, we kept method names and signatures intentionally
+compatible so existing call sites only needed to change their import from
+`package:mason_logger/mason_logger.dart` to `package:cli_io/cli_io.dart`.
+
+This is a clean-room implementation, not a fork — none of mason_logger's
```
felangel, line 24: This looks like a copy/paste of mason_logger's source code with slight modifications (I'm assuming AI generated). I'm very disappointed to see this 😕

packages/cli_io/lib/src/link.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:cli_io/src/ansi.dart';
```
felangel, line 1: As one example, this file is virtually identical to mason_logger's implementation 😕 

Image

packages/cli_io/README.md
spydon: In my opinion, the minimum thing to do in an open source ecosystem like this would be to at least give proper attribution...

packages/cli_io/lib/src/link.dart
spydon: It looks like someone has told an LLM to look at the mason_logger source and do a "clean room implementation" of it (which it of course by definition then isn't) :sweat_smile:

eseidel:
I've caused way more controversy than intended with this.  I still think the right decision for Shorebird is to remove this dependency, but that's something Nick can consider later.

felangel:
> I've caused way more controversy than intended with this. I still think the right decision for Shorebird is to remove this dependency, but that's something Nick can consider later.

If you want to remove the dependency and write your own CLI logger from scratch go for it. I have no problem with that although I imagine that's not the best use of anyone's time especially when you can contribute features/fixes upstream.

In case it wasn't clear, the problem I do have is with how this was framed as Shorebird is rolling its own logger even though it was just a copy/paste of `mason_logger` (with small edits most likely generated by AI which was also trained on `mason_logger`).

This is not setting a good example for the open source community and is very disappointing to see.
The lack of acknowledgement/accountability is even more concerning 😢

PR #3726, opened 2026-04-30
chore(shorebird_code_push_protocol): bump space_gen pin + regen

eseidel:
## Summary

Bumps `space_gen` from a git pin to the released `^1.2.2` on pub.dev and regenerates `lib/src/` from the public spec at [api.shorebird.dev/openapi.json](https://api.shorebird.dev/openapi.json) (per the README's regen instructions).

## What landed in space_gen 1.2.2

This package's regen surfaced two regressions in 1.2.1's long-line lint suppression pass, both fixed in 1.2.2 ([eseidel/space_gen#209](https://github.com/eseidel/space_gen/pull/209) + [#210](https://github.com/eseidel/space_gen/pull/210)):

1. **Wrong scope.** The post-walk pass mutated any `.dart` file in the output directory, including this package's hand-written `shorebird_code_push_protocol.dart` barrel — which our `tool/gen.dart` deliberately preserves by overriding `renderClient` / `renderPublicApi` as no-ops. Now scoped to files space_gen actually emits, with URI imports/exports excluded so files whose only over-80 lines are imports don't trip `unnecessary_ignore`.
2. **Style mismatch + orphan comments.** The fix in #209 fired against pre-format content; `dart fix --apply` then stripped the now-unnecessary directive but left a 3-line orphan justification on 61 files. #210 formats in-process via `DartFormatter` at write time so the directive decision matches what the analyzer sees, with format settings resolved from this package's own `pubspec.lock` / `pubspec.yaml` (language version) and `analysis_options.yaml` (`formatter: page_width` and `trailing_commas`, with relative `include:` paths followed so the workspace's `formatter:` block at the workspace root applies). `comment_references` resolution also now looks at same-file field declarations — fixes the same orphan-justification problem on `/// The signature of the [hash].` (referring to `final String hash;`).

## Verification

- [x] `dart analyze`: No issues found.
- [x] `dart test`: all 216 tests pass.
- [x] Hand-written `shorebird_code_push_protocol.dart` barrel **untouched** by space_gen.
- [x] **0 orphan-comment files** (long-line or comment_references).
- [x] Trailing commas preserved per `analysis_options.yaml`'s `trailing_commas: preserve`.

## Side note

`https://api.shorebird.dev/openapi.yaml` returns `{"code":"upgrade_required", "details":"Missing x-version header."}` even though the README links to it for human review. The JSON URL works without any header; the YAML URL is going through the version-required middleware. Worth a small server-side fix in a separate PR.

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/models/private_user.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/models/private_user.dart
@@ -28,7 +28,8 @@ class PrivateUser {
         id: json['id'] as int,
         email: json['email'] as String,
         displayName: json['display_name'] as String?,
-        hasActiveSubscription: json['has_active_subscription'] as bool?,
+        hasActiveSubscription:
+            json['has_active_subscription'] as bool? ?? false,
         stripeCustomerId: json['stripe_customer_id'] as String?,
         jwtIssuer: json['jwt_issuer'] as String,
         patchOverageLimit: json['patch_overage_limit'] as int?,
--- a/packages/shorebird_code_push_protocol/lib/model_helpers.dart
+++ b/packages/shorebird_code_push_protocol/lib/model_helpers.dart
@@ -13,6 +13,8 @@ T parseFromJson<T>(
 ) {
   try {
     return build();
+    // Rewrapping TypeError as FormatException is the whole point.
+    // ignore: avoid_catching_errors
   } on TypeError catch (error) {
     throw FormatException('Failed to parse $className from JSON: $error', json);
   }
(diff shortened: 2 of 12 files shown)


packages/shorebird_code_push_protocol/lib/src/models/private_user.dart
```diff
@@ -28,7 +28,8 @@ class PrivateUser {
         id: json['id'] as int,
         email: json['email'] as String,
         displayName: json['display_name'] as String?,
-        hasActiveSubscription: json['has_active_subscription'] as bool?,
+        hasActiveSubscription:
+            json['has_active_subscription'] as bool? ?? false,
```
nickshorebird, line 32: Nit - `hasActiveSubscription` is `bool?` but is always `bool` here.

nickshorebird (approved):
LGTM - Minor nit. Thanks for this.

packages/shorebird_code_push_protocol/lib/src/models/private_user.dart
eseidel: It's bool? because it's not required.  We should probably mark it x-internal or readonly so it doesn't show up in any "write" version.  But it's bool? in the class because you could imagine wanting to set it to null in the request.

Issue #2273, opened 2024-06-19
tests: write a unit test to prevent the cache clear regression


felangel:
**Description**

This is tracking the work to add a unit test to prevent a regression of https://github.com/shorebirdtech/shorebird/pull/2235.
An e2e test was added but it was flaky and reverted since we felt we could prevent this regression via a unit test instead.

PR #3733, opened 2026-05-03
fix: skip bootstrap when running 'shorebird cache clean' without a snapshot

### LINKED ISSUE #2234: feat: cache clean should not download artifacts if they are not present.
**Description**

Right now, if we run `shorebird cache clean` on an installation that doesn't have anything cached, the command will download the artifacts, just to delete them as part of the command logic.

We shouldn't update the cache when we are running a cache clean command, since that is kinda of a waste.

**Requirements**

- [ ] Don't try to update the cache on cache clean.

**Additional Context**

This is easy to reproduce by running `shorebird cache clean` twice.

bdero:
Fixes #2234, addresses #2273.

After `shorebird cache clean` removes `bin/cache/`, the next `shorebird` invocation sees the snapshot is gone and re-bootstraps from scratch: clones the Flutter fork, downloads the Flutter SDK, runs `flutter --version` (which precaches engine artifacts), runs `pub get`, and compiles the snapshot. If that next invocation is another `cache clean`, all of that work gets immediately deleted — exactly the "run `shorebird cache clean` twice" repro from #2234.

The fix short-circuits in `bin/shorebird`: when args are exactly `cache clean` or `cache clear` and the snapshot is missing, just `rm -rf` the cache directory and exit. No bootstrap, no downloads. Anything else (extra args, global flags, snapshot already present) falls through to the existing Dart command.

### Changes

- **`bin/shorebird`**: fast-path for `cache clean` / `cache clear` when the snapshot is absent. Mirrors the Dart command's exit codes and Windows superuser hint.
- **`clean_cache_command_test.dart`**: regression assertion that `cache.updateAll()` is never called from the Dart command. This is the unit-test follow-up tracked in #2273.

Note: there is some intentional duplication between the shell shortcut and the Dart command (success/failure messages, exit codes, the Windows hint). The shell can't call into the Dart command without the snapshot, which is exactly what we are avoiding, so they have to be expressed twice. Happy to consolidate by removing the Dart command entirely if reviewers prefer that direction.

## Test plan

- [x] `bash -n bin/shorebird` passes
- [x] Manual smoke tests: snapshot missing × cache dir present/absent, `clear` alias, extra-args fall-through, `cache foo` fall-through
- [x] `dart test test/src/commands/cache/clean_cache_command_test.dart` — 5/5 pass (4 pre-existing + 1 new)
- [x] `dart analyze` clean
- [x] `dart format` clean


eseidel:
I'm not sure if we should bother with this.  Or if we do, we should do shorebird.ps1 as well?

bdero:
yeah I figured it was a trivial fix, so decided to clank instead of close. We don't really need this

PR #3730, opened 2026-05-01
fix(shorebird_cli): improve help text and add input validation for agentic use

nickshorebird:
## Summary / Reasoning

Follow-up to #3720 and #3717. Agents reading `--help` output need accurate, complete descriptions to invoke commands correctly. This PR audits every command's help output from an agent's perspective and fixes what's vague, missing, or incorrect.

- Improves `--display-name` help text in `shorebird init` to describe what the field controls and its constraints; adds length validation (1–128 chars) with a clear error message
- Improves `--release-version` / `--release` help text across `patches promote`, `patches set-track`, and `releases get-apks` to show the version format (`"1.0.0+1"`) and clarify the patch ownership relationship; `releases get-apks` also notes it prompts for selection if omitted
- Improves `--track` help text in `patches set-track` to name the built-in tracks and clarify any custom name is accepted; adds length validation (1–128 chars, matching the `app_channel_.name` column)
- Removes the `allowed` list on `patch --min-link-percentage` (which produced an unhelpful parse error for out-of-range values) and replaces it with explicit runtime validation with a clear error message; extracts `minLinkPercentageMin/Max` constants
- Fixes grammar: trailing periods and a "to to" typo across several command descriptions
- Extracts `CommonArguments.patchReleaseVersionDescription` to avoid duplicating the same help string across `patches promote` and `patches set-track`

## Testing

- Ran `shorebird <cmd> --help` for every command and subcommand against the dev build
- `dart test` — 110 tests pass across the 3 affected test files

PR #3617, opened 2026-02-15
feat: add releases list command (#3616)

jorg1piano:
## Description
Creates a command that lets you list out releases in Shorebird.

`shorebird release list --platform ios --flavor ci --plain`

# --plain
only return a clean list. easier to work with in CI

# --platform ios|android|linux|windows
Filter result based on platform

# --flavor ci
Pick a flavor

## Type of Change
- [x ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 🛠️ Bug fix (non-breaking change which fixes an issue)
- [ ] ❌ Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 🧹 Code refactor
- [ ] ✅ Build configuration change
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ x] 🧪 Tests

PR #3736, opened 2026-05-04
feat(shorebird_cli): add `releases list` and `releases info` commands

nickshorebird:
## Summary / Reasoning

Follow-up to #3717 and #3730. Part of making Shorebird first-class in agentic environments. Agents need to enumerate releases and inspect release state without scraping unstructured output — these commands provide that. Supersedes #3617 (closes it — thanks @jorg1piano for the prior work on `releases list`; this extends it with `releases info`, `--json` support, and the stderr fix).

- Adds `shorebird releases list` — lists all releases for an app (ID, version, platform statuses, Flutter version); supports `--platform` filter, `--app-id`, `--flavor`, and global `--json`
- Adds `shorebird releases info --release-version <v>` — shows full details for a single release (ID, version, Flutter version, revision, created/updated dates, optional notes, per-platform statuses); supports `--app-id`, `--flavor`, and global `--json`
- Both commands emit a structured JSON error envelope on fetch failure under `--json`, making them safe for scripted consumers that need to distinguish "release not found" from other failures
- Fixes a pre-existing bug where progress spinner output leaked to stdout; progress is now unconditionally routed to stderr in both static and animated paths

## Testing Done

- `dart test packages/shorebird_cli/test/src/commands/releases/` — all tests pass
- `dart test packages/shorebird_cli/test/src/logging/shorebird_logger_test.dart` — stderr routing tests pass
- `dart analyze --fatal-warnings` and `dart format --set-exit-if-changed` — clean
- Diff coverage 100% on changed lines
- Ran both commands against a real app in interactive and `--json` modes; confirmed JSON output pipes cleanly to `jq`, progress stays on stderr, and a bad release version returns exit 70 with the error envelope

### DIFF
--- a/packages/shorebird_cli/lib/src/logging/shorebird_logger.dart
+++ b/packages/shorebird_cli/lib/src/logging/shorebird_logger.dart
@@ -6,7 +6,6 @@ import 'package:mason_logger/mason_logger.dart';
 import 'package:path/path.dart' as p;
 import 'package:scoped_deps/scoped_deps.dart';
 import 'package:shorebird_cli/src/interactive_mode.dart';
-import 'package:shorebird_cli/src/json_output.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 
 /// A reference to a [Logger] instance.
@@ -118,7 +117,7 @@ class ShorebirdLogger extends Logger {
   ///     `stderr` — progress is diagnostic, never content.
   @override
   Progress progress(String message, {ProgressOptions? options}) {
-    if (isInteractive && !isJsonMode) {
+    if (isInteractive) {
       return super.progress(message, options: options); // coverage:ignore-line
     }
     return _StaticProgress(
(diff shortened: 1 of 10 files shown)


packages/shorebird_cli/lib/src/logging/shorebird_logger.dart
```diff
@@ -114,15 +114,16 @@ class ShorebirdLogger extends Logger {
...
-  ///     under `--json` so it doesn't corrupt the JSON envelope, and to
-  ///     `stdout` otherwise.
+  ///     "Failed X" line on `complete`/`fail`. Output is always routed to
+  ///     `stderr` — progress is diagnostic, never content.
   @override
   Progress progress(String message, {ProgressOptions? options}) {
-    if (isInteractive) return super.progress(message, options: options);
+    if (isInteractive && !isJsonMode) {
+      return super.progress(message, options: options); // coverage:ignore-line
+    }
```
easymac: Minor: `isInteractive` is already defined as:
```dart
bool get isInteractive => io.stdout.hasTerminal && !isJsonMode;
```

The additional `&& !isJsonMode` check is dead today, but might be worth keeping if we're being defensive against future changes.

easymac (approved):
Looks good!

packages/shorebird_cli/lib/src/logging/shorebird_logger.dart
nickshorebird: Thanks for noticing this. Fixed on recent commit.

easymac (approved):

PR #3738, opened 2026-05-04
feat(cli): unified cross-platform patch numbering

easymac:
## Summary

Make `POST /v1/apps/{appId}/patches` idempotent on a user-supplied correlation key so cross-platform builds resolve to one patch row instead of one per platform.

Customer outcome: a developer who shipped "patch 3" sees the same number on iOS and Android — in their analytics, in-app, and in their release notes. The dominant CI flow (iOS bot publishes, Android bot publishes minutes later, no shared state between them) collapses to one patch as long as both bots pass the same `--patch-id`.

## Protocol — `shorebird_code_push_protocol`

- `CreatePatchRequest` gains optional `clientPatchId` (wire field `client_patch_id`). `toJson` always emits the key (null when unset) so server parsing doesn't have to special-case absence.
- `CreatePatchResponse` (new field) echoes `clientPatchId` back so callers can verify whether an idempotent re-use occurred.
- `ReleasePatch` carries `clientPatchId` so existing GET endpoints surface the correlation key for admin/console rendering.

## Client — `shorebird_code_push_client`

- `CodePushClient.createPatch` accepts and threads `clientPatchId`, and now returns `CreatePatchResponse` instead of `Patch` so callers can read the echoed key.
- Empty-string `clientPatchId` is normalized to null at the client boundary — a caller that passes an unexpanded template variable or empty flag shouldn't land on the idempotent path keyed on `''` and inherit a stranger's patch. Done at the client rather than the protocol so the protocol layer remains a faithful round-trip.

## CLI — `shorebird_cli`

- New `--patch-id=<string>` flag on `shorebird patch`. Common usage: `--patch-id=${{ github.sha }}` or any stable token (`hotfix-login`, a CI run ID, etc.).
- Multi-platform invocations (`--platforms=ios,android`) auto-generate a UUID and share it across every per-platform fan-out — single-invocation is just cross-invocation-with-an-internal-ID.
- Append-after-promotion confirmation: when an idempotent hit lands on a patch already promoted to stable, the new platform's artifacts go live to that platform's stable users immediately. In interactive mode the CLI prompts before continuing; in CI it proceeds silently (the iOS-publishes-then-Android-completes flow is the canonical use case and prompting would deadlock).
- Suppresses the prompt when the same run promoted the patch (e.g. the first platform of a single `--platforms ios,android` invocation — the user already opted in by passing both platforms together).
- Success line: `✅ Published Patch 3 (iOS, Android)` instead of one log per platform; degenerate cases where platforms land on different numbers each get their own line.

## Backwards compatibility

- Missing `client_patch_id` deserializes to null. Old clients that don't set the field fall through the existing per-call create path on the server.
- Old servers that don't recognize the field ignore it on deserialization; the CLI flag becomes a no-op until the server side ships, with no error surface.

This is the protocol + CLI half of unified patch numbering. The server-side migration and idempotent `createPatch` live in the private parent repo and are gated behind the same `client_patch_id` correlation key.

## Test plan

- [ ] `dart test packages/shorebird_code_push_protocol`
- [ ] `dart test packages/shorebird_code_push_client`
- [ ] `dart test packages/shorebird_cli`
- [ ] Manual: `shorebird patch android --patch-id=test-sha` against a server with the matching server-side change deployed
- [ ] Manual: `shorebird patch --platforms=ios,android` end-to-end (single invocation; auto-generated UUID)
- [ ] Manual: simulate cross-invocation by running two `shorebird patch` calls minutes apart with the same `--patch-id`
- [ ] Manual: append-after-promotion in interactive TTY (expect prompt) and in CI/non-TTY (expect silent proceed)

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -94,10 +94,10 @@ class CodePushClientWrapper {
   /// The underlying code push client.
   final CodePushClient codePushClient;
 
-  /// Patch ids this wrapper has promoted during the current command run.
-  /// Used to suppress the append-after-promotion warning for the second
-  /// platform of a single `--platforms ios,android` invocation, where the
-  /// "already on stable" state was created moments earlier by this same run.
+  /// Patch ids promoted by this wrapper instance — command-scoped, so a
+  /// fresh `shorebird patch` gets an empty set and cross-invocation reuse
+  /// still prompts. Suppresses the append-after-promotion warning for the
+  /// second platform of a single `--platforms ios,android` invocation.
   final Set<int> _patchesPromotedThisRun = {};
 
   /// Create an app with the given [organizationId] and [appName].
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -981,8 +981,6 @@ aar artifact already exists, continuing...''');
     // responsible for making sure the developer knows.
     if (clientPatchId != null) {
       await _confirmAppendToPromotedPatch(
-        appId: appId,
-        releaseId: releaseId,
         patch: patch,
         platform: platform,
       );
@@ -1017,9 +1015,12 @@ aar artifact already exists, continuing...''');
   /// Skipped entirely when this run is the one that promoted the patch
   /// (e.g. the second platform of a `--platforms ios,android` invocation),
   /// since the user already opted in by passing both platforms together.
+  ///
+  /// The patch's current channel comes from [CreatePatchResponse.channel],
+  /// which the server populates on the idempotent return path. Older servers
+  /// that don't echo `channel` leave it null and the prompt simply doesn't
+  /// fire — append-after-promotion stays allowed per the design.
   Future<void> _confirmAppendToPromotedPatch({
-    required String appId,
-    required int releaseId,
     required CreatePatchResponse patch,
     required ReleasePlatform platform,
   }) async {
@@ -1028,22 +1029,7 @@ aar artifact already exists, continuing...''');
     // doing and the prompt would be spurious.
     if (_patchesPromotedThisRun.contains(patch.id)) return;
 
-    final List<ReleasePatch> existing;
-    try {
-      existing = await codePushClient.getPatches(
-        appId: appId,
-        releaseId: releaseId,
-      );
-    } on Exception {
-      // If we can't determine the patch's promotion state, fall back to the
-      // server's permissive behavior — the design treats append-after-
-      // promotion as allowed; the prompt is a courtesy.
-      return;
-    }
-
-    final priorState = existing.firstWhereOrNull((p) => p.id == patch.id);
-    final isOnStable = priorState?.channel == DeploymentTrack.stable.channel;
-    if (!isOnStable) return;
+    if (patch.channel != DeploymentTrack.stable.channel) return;
 
     final platformName = platform.displayName;
     final message =
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -94,6 +94,12 @@ class CodePushClientWrapper {
   /// The underlying code push client.
   final CodePushClient codePushClient;
 
+  /// Patch ids this wrapper has promoted during the current command run.
+  /// Used to suppress the append-after-promotion warning for the second
+  /// platform of a single `--platforms ios,android` invocation, where the
+  /// "already on stable" state was created moments earlier by this same run.
+  final Set<int> _patchesPromotedThisRun = {};
+
   /// Create an app with the given [organizationId] and [appName].
   Future<App> createApp({required int organizationId, String? appName}) async {
     late final String displayName;
@@ -872,11 +878,16 @@ aar artifact already exists, continuing...''');
   }
 
   /// Creates a patch for the given [appId], [releaseId], and [metadata].
+  ///
+  /// When [clientPatchId] is supplied, the server treats the call as
+  /// idempotent: if a patch on this release already has that id, the
+  /// existing patch is returned instead of creating a new one.
   @visibleForTesting
-  Future<Patch> createPatch({
+  Future<CreatePatchResponse> createPatch({
     required String appId,
     required int releaseId,
     required Json metadata,
+    String? clientPatchId,
   }) async {
     final createPatchProgress = logger.progress('Creating patch');
     try {
@@ -884,6 +895,7 @@ aar artifact already exists, continuing...''');
         appId: appId,
         releaseId: releaseId,
         metadata: metadata,
+        clientPatchId: clientPatchId,
       );
       createPatchProgress.complete();
       return patch;
@@ -896,7 +908,7 @@ aar artifact already exists, continuing...''');
   @visibleForTesting
   Future<void> createPatchArtifacts({
     required String appId,
-    required Patch patch,
+    required CreatePatchResponse patch,
     required ReleasePlatform platform,
     required Map<Arch, PatchArtifactBundle> patchArtifactBundles,
   }) async {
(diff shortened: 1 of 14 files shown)


packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -970,8 +1000,67 @@ aar artifact already exists, continuing...''');
...
+    required ReleasePlatform platform,
+  }) async {
+    // If this run promoted the patch itself (e.g. the first platform of
+    // `--platforms ios,android`), the "already on stable" state is our own
+    // doing and the prompt would be spurious.
+    if (_patchesPromotedThisRun.contains(patch.id)) return;
+
+    final List<ReleasePatch> existing;
+    try {
+      existing = await codePushClient.getPatches(
```
bdero: _bdero's review bot_:

💡 The dominant CI flow is "every run has a unique SHA" — so every `publishPatch` is a genuinely new patch row, but we still fetch the full patch list on every one of them just to confirm "not already on stable." Two server round-trips per platform instead of one, scaling with release-patch-count on the read path.

The cleanest fix lives server-side: have `CreatePatchResponse` return a `created: bool` (or echo back the patch's current channel) so the CLI can skip the `getPatches` lookup when the response says we created a new row. Worth raising with the server PR while it's still in flight.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -970,8 +1000,67 @@ aar artifact already exists, continuing...''');
...
+        releaseId: releaseId,
+      );
+    } on Exception {
+      // If we can't determine the patch's promotion state, fall back to the
+      // server's permissive behavior — the design treats append-after-
+      // promotion as allowed; the prompt is a courtesy.
+      return;
+    }
+
+    final priorState = existing.firstWhereOrNull((p) => p.id == patch.id);
```
bdero: _bdero's review bot_:

💡 `firstWhereOrNull((p) => p.id == patch.id)` over the full list — if `getPatches` ever returns a windowed/recent slice (or starts to), an older patch's promotion state silently reads as "not on stable" and the courtesy prompt vanishes. Not sure if `getPatches` paginates today — worth confirming, and if there's a "fetch one patch by id" endpoint we should prefer that.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -215,6 +230,72 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
...
+  /// preserving the legacy behavior where the server allocates a fresh patch
+  /// number per call.
+  late final String? clientPatchId = _resolveClientPatchId();
+
+  String? _resolveClientPatchId() {
+    final explicit = results['patch-id'] as String?;
+    if (explicit != null && explicit.isNotEmpty) return explicit;
+    if (results.releaseTypes.length > 1) return const Uuid().v4();
+    return null;
+  }
```
bdero, line 257: _bdero's review bot_:

💡 `if (explicit != null && explicit.isNotEmpty) return explicit;` treats `--patch-id=""` and `--patch-id` (omitted) the same. The failure mode: someone writes `--patch-id=${{ env.PATCH_SHA }}` in their CI template, the env var is undefined, the flag arrives as `""`, and on a multi-platform invocation we silently generate a UUID — the iOS bot and the Android bot never converge because each generates its own. Loud failure when the flag is *present but empty* would surface this on the first broken CI run instead of after someone notices the patches aren't unifying.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -273,14 +273,27 @@ class CodePushClient {
...
+  /// verify the correlation.
+  Future<CreatePatchResponse> createPatch({
     required String appId,
     required int releaseId,
     required Json metadata,
+    String? clientPatchId,
   }) async {
+    // Coerce empty to null. A caller that passes an unexpanded template
+    // variable or empty flag would otherwise land on the idempotent path
+    // keyed on `''` and inherit a stranger's patch.
```
bdero, line 290: _bdero's review bot_:

💡 The PR description argues normalization belongs at the client boundary (so the protocol layer stays a faithful round-trip), which I agree with — but then `_resolveClientPatchId` in `patch_command.dart` also normalizes empty-to-null at the command layer. Pick one. If the "error on `--patch-id=""`" suggestion lands, the command-layer normalization goes away naturally and this stays the only normalizer.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -239,9 +322,37 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
...
+    final byPatchNumber = <int, List<ReleasePlatform>>{};
+    for (final entry in platformPatches.entries) {
+      byPatchNumber.putIfAbsent(entry.value.number, () => []).add(entry.key);
+    }
+
+    for (final number in byPatchNumber.keys.toList()..sort()) {
+      final platforms =
+          byPatchNumber[number]!.map((p) => p.displayName).toList()..sort();
+      logger.success(
+        '\n✅ Published Patch $number (${platforms.join(', ')})',
```
bdero: _bdero's review bot_:

🧹 Lost the `!` — was `Published Patch X!`, now `Published Patch X (Android, iOS)`. The `!` matches the celebratory tone of other `logger.success` lines in the CLI. `Published Patch 3 (Android, iOS)!` keeps it consistent.

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -94,6 +94,12 @@ class CodePushClientWrapper {
   /// The underlying code push client.
   final CodePushClient codePushClient;
 
+  /// Patch ids this wrapper has promoted during the current command run.
+  /// Used to suppress the append-after-promotion warning for the second
+  /// platform of a single `--platforms ios,android` invocation, where the
+  /// "already on stable" state was created moments earlier by this same run.
+  final Set<int> _patchesPromotedThisRun = {};
```
bdero, line 101: _bdero's review bot_:

🧹 The name reads as if it tracks global state — it actually tracks "promotions made by this wrapper instance," which has command-scoped lifetime via the scoped ref. A field comment noting the scope (and that cross-invocation reuse correctly prompts because the set starts empty) would save the next reader the trace.

packages/shorebird_cli/test/src/commands/patch/patch_command_test.dart
```diff
@@ -382,6 +388,382 @@ void main() {
...
+        test(
+          '''generates a UUID for multi-platform invocations so platforms share a patch''',
+          () async {
+            when(() => argResults['platforms']).thenReturn(['ios', 'android']);
+            await runWithOverrides(() => command.createPatch(patcher));
+            expect(capturedClientPatchId, isNotNull);
+            // RFC4122 v4: 8-4-4-4-12 hex with version nibble == 4. Asserted
+            // as a single raw string to keep the regex semantics obvious;
+            // splitting it earned a "use raw string" / "adjacent strings"
+            // ping-pong from the linter.
```
bdero: _bdero's review bot_:

🧹 The "linter ping-pong" comment is more interesting than the regex it explains. Either trust the raw string and drop the comment, or hoist the regex to a top-level `final RegExp _v4 = RegExp(r'...');` with a one-line "RFC 4122 v4" note — both quieter than the current four-line explanation of why-the-lint-fought-us.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -215,6 +230,72 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
...
+  /// The correlation key used to make patch creation idempotent across
+  /// platforms within this invocation. When the user passes `--patch-id`,
+  /// that value is used as-is. When the user omits it on a multi-platform
+  /// invocation, a fresh UUID is generated so the platforms in this single
+  /// command end up sharing one patch number on the server.
+  ///
+  /// Returns null only for single-platform invocations with no `--patch-id`,
+  /// preserving the legacy behavior where the server allocates a fresh patch
+  /// number per call.
+  late final String? clientPatchId = _resolveClientPatchId();
```
bdero, line 247: _bdero's review bot_:

🧹 `late final String? clientPatchId = _resolveClientPatchId();` works, but `_resolveClientPatchId()` reads `results[...]` and calls `Uuid().v4()`, which means the order of first access decides which `Uuid` instance is materialized. Tests have to go through `runWithOverrides` for argResults to be wired before the first access — they do — but a getter (or a method returning a memoized value) reads more obviously than `late final`-with-side-effect.

bdero (approved):

bdero:
Overall makes sense & approach seems good to me. Bot has some good looking suggestions. My human comment keeps getting dropped when submitting these for some reason 🤷

packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
```diff
@@ -970,8 +998,55 @@ aar artifact already exists, continuing...''');
...
+    final message =
+        'Patch ${patch.number} is already promoted to the stable track. '
+        'Uploading $platformName artifacts will go live to $platformName '
+        'stable users immediately.';
+
+    if (shorebirdEnv.canAcceptUserInput) {
+      logger.warn(message);
+      if (!logger.confirm('Continue?')) {
+        logger.info('Aborting.');
+        throw ProcessExit(ExitCode.success.code);
```
nickshorebird, line 1044: Nit - `0` would cause further `$?` checks to succeed when the logic was aborted. There is not unix standard for this that I know of, but it is typically a `1` return for general failure.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -239,9 +342,37 @@ NOTE: this is ${styleBold.wrap('not')} recommended. Asset changes cannot be incl
       await patcherFuture;
     }
 
+    logUnifiedSuccess();
```
nickshorebird, line 345: Consider moving this to a finally or log inside each patch success. There is an edge case where `patcherFutures[0]` succeeds and `patcherFutures[1]` fails and no log is fired. Fine w/ a retry, but should notify the user w/ the UUID or patchID to retry or else this will be a friction point / force the user to call follow up commands to identify the patchID.

nickshorebird (commented):
LGTM overall. One friction point worth surfacing before merge.

Auto genned UUIDs from the multi-platform path are never echoed back. They go to the server in the request, come back on the response, and stay in memory until the process exits. If the fan out partial fails, say iOS publishes and android errors, the user has no way to recover the correlation key and rerun android against the same patch row.

Easiest fix: echo the resolved patchid in the success line whenever it was auto genned, or include it in a final failure message so the resume command is visible. Even better would be to add a hint to tell the user to rerun w/ command for easier agentic usage + quick copy paste. Otherwise the cross-invocation recovery flow only works for the explicit --patch-id path.

PR #3740, opened 2026-05-05
feat(shorebird_cli): add patches list/info; --json for set-track

nickshorebird:
## Summary / Reasoning

Follow-up to #3736. Same agentic-CLI motivation: agents need to enumerate patches and inspect patch state through structured output, and the new `set-track --json` lets agents change deployment tracks without an interactive prompt.

- Adds `shorebird patches list --release-version <v>`, lists all patches for a release with id, number, track, and rolled-back state. Supports `--app-id`, `--flavor`, and global `--json`.
- Adds `shorebird patches info --release-version <v> --patch-number <n>`, shows full details for a specific patch including artifacts. Same flag surface as list.
- `patches set-track` gains `--json` mode and an `--app-id` flag so it can run without `shorebird init` first.
- `patches promote` is deprecated. Under `--json` it now refuses with a structured `usage_error` envelope pointing to `set-track`, instead of leaking the deprecation warning to stdout. Deprecated commands shouldn't grow new output surface.
- Extracts `ShorebirdCommand.resolveAppId` and `ShorebirdCommand.jsonHint` so `--app-id` resolution and the "Pass --json…" help suffix aren't duplicated. Retrofits `releases list`/`info` to use them too.
- Help text mirrors the releases pattern: short summary, example output block, then a "Pass --json …" hint. Skimmable for humans, example-as-schema for agents.
- Renders `[no track]` for patches with a null channel instead of the literal `null`. Covers both human and JSON paths.
- Reuses the existing `JsonErrorCode` envelopes: `fetch_failed` for API errors, `software_error` for unexpected failures, `usage_error` for bad input, `interactive_prompt_required` when `set-track` would need a TTY prompt.

Note on legacy: `set-track` keeps `--release` and `--patch` as flag names. The new commands use `--release-version` and `--patch-number`. Renaming would break scripts already in use, so it needs its own deprecation cycle.

Note on size: the diff is large but lib code is +422 / -98. The remaining +1169 / -68 is unit tests covering the new commands and `--json` paths. Scope stays within recently-merged sibling work, mostly the agentic-CLI series #3717, #3720, #3730, #3736.

## Testing

- Unit tests cover human output, `--json` success and error envelopes, network failures via `ProcessExit`, null-channel `[no track]` rendering, rolled-back rendering, app-id resolution from flag, yaml, and flavor, and empty list. Tests added across `patches list`, `patches info`, `patches set-track`, and `patches promote`.
- Human-readable output is asserted line-by-line on exact strings, not substrings, so renaming a label like `Track:` to `Channel:` or regressing the size formatter from `1.20 MB` back to raw bytes would fail the test.
- 123 of 123 patches and releases tests pass. `dart format` and `dart analyze --fatal-warnings` clean on every edited file.
- Smoke-tested end-to-end against api-dev with a real app: `releases list/info`, `patches list/info`, plus bad-input error paths. Confirmed `--json` stdout is pure JSON so `… --json | jq` works without redirection.


packages/shorebird_cli/lib/src/commands/patches/set_track_command.dart
```diff
@@ -43,64 +38,118 @@ class SetTrackCommand extends ShorebirdCommand {
...
   @override
   String get name => 'set-track';
 
   @override
-  String get description => 'Sets the track of a patch.';
+  String get description =>
+      'Sets the track of a patch.\n\n'
+      'Example output:\n'
+      '  Patch 1 on release 1.0.0+1 is now in channel stable!\n\n'
+      '${ShorebirdCommand.jsonHint('shorebird patches set-track --release 1.0.0+1 --patch 1 --track stable --app-id <id> --json')}';
```
easymac, line 60: Should this hint use --release-version and --patch number?

easymac (approved):

packages/shorebird_cli/lib/src/commands/patches/set_track_command.dart
nickshorebird: Relates to the legacy-flags note in the PR description. The dartdoc above, line 16, was inadvertently using `--release-version` and `--patch-number`, which do not exist on this command. Reverted in 5ba9. The `jsonHint` here matches the actual flags `--release` and `--patch`, so I am going to leave it as is.

PR #3742, opened 2026-05-06
feat(shorebird_cli): add `account` command (whoami + orgs)

nickshorebird:
## Why

`shorebird account` was a dead end. Anyone scripting against the CLI (or just running `--help`) had no way to ask "who am I logged in as?" or "what orgs do I belong to?" without going to the web console. This fills that gap w/ the two read-only commands that map directly to existing client methods, so it's a small, low-risk addition that closes an obvious discoverability hole.

## Summary

- Adds `account whoami` (email, display name, plan, overage limit) and `account orgs` (id, name, type, role).
- Both support the global `--json` envelope and follow the same shape as `releases list/info` (#3736) and `patches list/info` (#3740), reusing `emitJsonSuccess` / `emitJsonError` w/ `JsonErrorCode.fetchFailed`.
- No protocol or server changes. Reuses existing `CodePushClient.getCurrentUser()` and `getOrganizationMemberships()`.
- Remaining console-equivalent surface (org CRUD, member mgmt, API keys, usage, audit logs) deferred. Those need new client+protocol work and will land separately.

## Testing

- `dart analyze --fatal-warnings lib test`, clean
- `dart format`, clean
- `dart test test/src/commands/account/ test/src/code_push_client_wrapper_test.dart`, 108 pass
- `dart test test/src/shorebird_cli_command_runner_test.dart`, pass
- Smoke-tested human and `--json` output for `whoami`, `orgs`, and the forced-unauth error envelope locally

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -116,6 +116,23 @@ class CodePushClientWrapper {
     );
   }
 
+  /// Fetches the currently authenticated user.
+  Future<PrivateUser> getCurrentUser() async {
+    final progress = logger.progress('Fetching account');
+    final PrivateUser? user;
+    try {
+      user = await codePushClient.getCurrentUser();
+      progress.complete();
+    } catch (error) {
+      _handleErrorAndExit(error, progress: progress);
+    }
+    if (user == null) {
+      logger.err('Could not find current user.');
+      throw ProcessExit(ExitCode.software.code);
+    }
+    return user;
+  }
+
   /// Fetches the organization memberships for the current user.
   Future<List<OrganizationMembership>> getOrganizationMemberships() async {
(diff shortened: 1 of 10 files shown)


eseidel (approved):
It does feel a bit odd to duplicate the pathway js vs logger.  I wonder if we should build ourselves pathways that only produce json, and then optional human interfaces that turn that json into logged text, with some automatic formatting for simple stuff?  Not in this PR, just over time.

nickshorebird:
> It does feel a bit odd to duplicate the pathway js vs logger. I wonder if we should build ourselves pathways that only produce json, and then optional human interfaces that turn that json into logged text, with some automatic formatting for simple stuff? Not in this PR, just over time.

Agreed it is worth doing after all the commands are released. Adds some overhead for new commands now, but allows them to go out + decide on human first, like gh, or json first, like gcloud or aws cli. That is to say, refactor this to remove the repeated `if` antipattern and move to a single return object w/ a printer && formatter layer overtop.

PR #3743, opened 2026-05-06
feat(shorebird_cli): add account apps subcommand

nickshorebird:
## Summary

- Adds `shorebird account apps`, which lists apps the current user has access to. Mirrors the `account orgs` shape so the `account` namespace stays parallel (whoami, orgs, apps).
- Pure CLI add. Reuses the existing `codePushClientWrapper.getApps()` already used internally by release/patch flows. No new wrapper, protocol, or server work.
- Human output: space-separated `app_id  display_name  latest_release_version  latest_patch_number`, w/ `-` for unreleased apps. `--json` returns `{ apps: [{ app_id, display_name, latest_release_version, latest_patch_number }] }`.
- Help-text column labels match JSON keys verbatim, so an agent reading `--help` and then parsing `--json` sees identical field names. Caught while comparing the three account subcommands' help output side-by-side.
- Closes the last CLI-only gap in the `account` surface from the agentic smoke test. Remaining `account` capabilities (org create, member mgmt, API keys, usage/plan, audit logs, transfer-app-to-org) all need backend work and will land separately.

## Why not a plain `app` command?

I originally planned to have `app`, eg `shorebird app`, but each of the app commands are either semantically covered already, eg `create` w/ `release`, or logically fit better into account or other parents.

## Testing

- 10 new tests in `apps_command_test.dart`: precondition failure, auth required, populated and empty-state human output, API-error path in human and JSON modes, JSON success shape w/ flat fields, projection check (no `created_at`/`updated_at` leak), empty-array case
- `dart test packages/shorebird_cli`: 2028 passed, 1 skipped (pre-existing), 0 failures
- `dart analyze --fatal-warnings lib test`: clean on new files
- `dart format`: clean
- Manual smoke against an authenticated dev environment: `shorebird account apps` and `shorebird account apps --json` both render correctly; forced-unauthenticated path returns the expected JSON error envelope

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/account/apps_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/account/apps_command.dart
@@ -1,0 +1,85 @@
+import 'package:mason_logger/mason_logger.dart';
+import 'package:shorebird_cli/src/code_push_client_wrapper.dart';
+import 'package:shorebird_cli/src/json_output.dart';
+import 'package:shorebird_cli/src/logging/logging.dart';
+import 'package:shorebird_cli/src/shorebird_command.dart';
+import 'package:shorebird_cli/src/shorebird_validator.dart';
+import 'package:shorebird_cli/src/third_party/flutter_tools/lib/src/base/process.dart';
+import 'package:shorebird_code_push_client/shorebird_code_push_client.dart';
+
+/// {@template apps_command}
+/// `shorebird account apps`
+/// List the apps the current user has access to.
+/// {@endtemplate}
+class AppsCommand extends ShorebirdCommand {
+  /// {@macro apps_command}
+  AppsCommand();
+
+  @override
+  String get name => 'apps';
+
+  @override
+  String get description =>
+      'List the apps you have access to.\n\n'
+      'Example output (space-separated: app_id  display_name  '
+      'latest_release_version  latest_patch_number):\n'
+      '  01H...  Acme Mobile  1.2.3  4\n'
+      '  01J...  Acme Internal  -  -\n\n'
+      '"-" indicates no release or patch has been published yet.\n\n'
+      '${ShorebirdCommand.jsonHint('shorebird account apps --json')}';
+
+  @override
+  Future<int> run() async {
+    try {
+      await shorebirdValidator.validatePreconditions(
+        checkUserIsAuthenticated: true,
+      );
+    } on PreconditionFailedException catch (error) {
+      return error.exitCode.code;
+    }
+
+    final List<AppMetadata> apps;
+    try {
+      apps = await codePushClientWrapper.getApps();
+    } on ProcessExit catch (e) {
+      if (isJsonMode) {
+        emitJsonError(
+          code: JsonErrorCode.fetchFailed,
+          message: 'Failed to fetch apps.',
+        );
+        return e.exitCode;
+      }
+      rethrow;
+    }
+
+    if (isJsonMode) {
+      emitJsonSuccess({
+        'apps': [
+          for (final app in apps)
+            {
+              'app_id': app.appId,
+              'display_name': app.displayName,
+              'latest_release_version': app.latestReleaseVersion,
+              'latest_patch_number': app.latestPatchNumber,
+            },
+        ],
+      });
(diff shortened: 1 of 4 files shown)


packages/shorebird_cli/lib/src/commands/account/apps_command.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  @override
+  String get description =>
+      'List the apps you have access to.\n\n'
+      'Example output (space-separated: app_id  display_name  '
+      'latest_release_version  latest_patch_number):\n'
+      '  01H...  Acme Mobile  1.2.3  4\n'
+      '  01J...  Acme Internal  -  -\n\n'
+      '"-" indicates no release or patch has been published yet.\n\n'
+      '${ShorebirdCommand.jsonHint('shorebird account apps --json')}';
```
bdero, line 29: _bdero's review bot_:

💡 help text describes the output as `space-separated`, but `display_name` can contain spaces (the example literally uses `Acme Mobile`). A naive `awk '{print $2}'` won't survive that. `--json` is the right answer for parsers and is already advertised, but the "space-separated" phrasing implies parseability it doesn't actually have. Could soften to something like "columns separated by two spaces (use `--json` for parsing)". Same wording exists in `orgs_command.dart`, so this is a parallel issue, and fine as a follow-up across both.

packages/shorebird_cli/lib/src/commands/account/apps_command.dart
```diff
@@ -0,0 +1,85 @@
...
+    if (apps.isEmpty) {
+      logger.info('No apps found.');
+      return ExitCode.success.code;
+    }
+
+    for (final app in apps) {
+      logger.info(
+        '${app.appId}  ${lightCyan.wrap(app.displayName)}  '
+        '${app.latestReleaseVersion ?? '-'}  '
+        '${app.latestPatchNumber ?? '-'}',
```
bdero, line 79: _bdero's review bot_:

🧹 not sure if this can happen in practice, but `latestPatchNumber: 0` would render as the literal `0` while the help text reserves `-` for "no release or patch published yet". If the server only ever returns `null` (never `0`) for a release-with-no-patches state then this is moot, but worth a sanity check.

packages/shorebird_cli/test/src/commands/account/apps_command_test.dart
```diff
@@ -0,0 +1,253 @@
...
+
+    setUp(() {
+      argResults = MockArgResults();
+      codePushClientWrapper = MockCodePushClientWrapper();
+      logger = MockShorebirdLogger();
+      progress = MockProgress();
+      shorebirdValidator = MockShorebirdValidator();
+      command = runWithOverrides(AppsCommand.new)..testArgResults = argResults;
+
+      when(() => logger.progress(any())).thenReturn(progress);
```
bdero, line 63: _bdero's review bot_:

🧹 `progress` and the `logger.progress(any())` stub are dead weight here: `codePushClientWrapper.getApps()` is mocked at the wrapper boundary, so the progress code inside the wrapper never runs. Same in `orgs_command_test.dart`, so I get why it's there. Flagging only for awareness, no need to fix in this PR.

bdero (approved):

bdero (approved):
lgtm, very minor bot comments

packages/shorebird_cli/lib/src/commands/account/apps_command.dart
nickshorebird: Thanks for the comment. Going to roll this into the follow up single object and writer return change that is meant to be a follow up.

packages/shorebird_cli/lib/src/commands/account/apps_command.dart
nickshorebird: Dropping as confirm contract is always null for no patches. Will revisit if this ever changes.

packages/shorebird_cli/test/src/commands/account/apps_command_test.dart
nickshorebird: Thanks. Will roll this into a follow up debt change rather than leave this open.

Issue #3728, opened 2026-05-01
bug: checkForUpdate reports upToDate after patch-to-release rollback (regression of #3206)


eseidel:
## Summary

`shorebird_code_push.checkForUpdate()` returns `UpdateStatus.upToDate` when
the running patch is rolled back to the base release version. This is a
follow-up to #3206 — the fix shipped in shorebird_code_push 2.0.6
(shorebirdtech/updater#312) addresses some rollback cases but not the
patch-to-release case described here.

## Reproduction

1. Install patch 1 on a device, let the app boot successfully.
2. Roll back patch 1 from the dashboard (no replacement patch).
3. From the running app, call `await updater.checkForUpdate()`.

**Expected:** `UpdateStatus.restartRequired`
**Actual:** `UpdateStatus.upToDate`

The app does revert to the release version on next launch, but the running
session has no signal to prompt a restart.

Patch-to-patch rollback (e.g. rolling back patch 2 while patch 1 is still
available) is detected correctly, because the server response includes a
downloadable patch and `checkForUpdate` short-circuits to `outdated`.

## Versions

- shorebird_code_push: 2.0.6
- Flutter: 3.41.7
- Reported by a customer migrating to Shorebird.

## Root cause

The fix in shorebirdtech/updater#312 relies on Dart seeing
`current.number != null && next.number == null` after a rollback. That
holds in the test (which mocks `currentPatchNumber` to return 1), but not
in the real FFI path:

1. After successful boot, `record_boot_success` clears
   `currently_booting_patch` and sets `last_booted_patch = patch_1`.
2. `checkForUpdate` calls `check_for_downloadable_update`, which calls
   `roll_back_patches_if_needed([1])` →
   `try_fall_back_from_patch(1)` (`library/src/cache/patch_manager.rs`).
3. Because patch 1 is both `last_booted_patch` and `next_boot_patch`,
   the "both bad" branch clears **both** fields.
4. `current_boot_patch()` is defined as
   `currently_booting_patch.or(last_successfully_booted_patch)`
   (`library/src/cache/updater_state.rs`), so it now returns `None`.
5. `shorebird_current_boot_patch_number()` returns 0,
   `shorebird_next_boot_patch_number()` returns 0,
   Dart compares `null != null` → false → `upToDate`.

The original issue's note — *"After the rollback has been applied, both
currentPatch and nextPatch report null, so that might be the real issue"*
— was correct; the Dart-only fix did not address that.

## Proposed fix

[...]

PR #3750, opened 2026-05-08
feat(shorebird_cli): add patches rollback / rollforward

bdero:
## Summary

Rollback and rollforward have existed as HTTP endpoints for a while but the
dashboard is the only way to reach them, so reproducing #3728 against a real
device meant scripting raw curl calls in a side repo, and a CI kill-switch has
to hand-roll an HTTP client and token handling. Review turned up two defects in
the first cut and both are fixed here rather than deferred: the commands
derived their result from a pre-flight read, so a patch flipped between that
read and the POST reported a change that never happened, and a non-integer
`--patch-number` threw a `FormatException` that surfaced as exit 70 with a
"file an issue" footer. The result now comes from the server, which answers 304
when the patch is already in the requested state. The bad-argument handling is
a mixin shared with `patches info`, which had the same crash.

- `patches rollback` / `patches rollforward`, taking `--release-version` and
  `--patch-number` alongside `--app-id`, `--flavor`, and `--json`
- A no-op exits 0 and prints the patch, since the requested end state holds.
  `--require-change` gets a non-zero exit when the server made no change
- Both print the `patches info` detail block on success, so a follow-up
  `patches info` call is unnecessary. `--json` carries `changed` and `patch`

## Testing

- Smoke-tested end to end against `api.shorebird.dev` while reproducing #3728.
  Rollback returns 204, the device picks up the signal on its next poll, then
  rollforward returns 204 and the auto-update thread reinstalls the patch
- Diff coverage is 100% across the three touched packages, 181 changed lines


packages/shorebird_cli/lib/src/commands/patches/rollback_command.dart
```diff
@@ -0,0 +1,155 @@
...
+          'Available patches: ${patches.map((p) => p.number).join(', ')}',
+        );
+      return ExitCode.usage.code;
+    }
+
+    if (patch.isRolledBack) {
+      if (isJsonMode) {
+        emitJsonError(
+          code: JsonErrorCode.usageError,
+          message: 'Patch $patchNumber is already rolled back.',
```
nickshorebird, line 113: Take it or leave it - End state is the same is the same on success or hitting this route. Returning a non zero exit suggests that something went wrong even though the end state is correct. Consider adding some flag IE `--require-change`, I am awful at naming to please feel free to change it, if you want the behavior to be there on default or non default. Changing to this would also match other CLI tool behaviors. This is in line w/ current set-track behavior hence take it or leave it.

packages/shorebird_cli/lib/src/commands/patches/rollback_command.dart
```diff
@@ -0,0 +1,155 @@
...
+        return e.exitCode;
+      }
+      rethrow;
+    }
+
+    if (isJsonMode) {
+      emitJsonSuccess({
+        'release_version': releaseVersion,
+        'patch_number': patchNumber,
+        'action': 'rollback',
```
nickshorebird, line 145: Take it or leave it - Consider printing the patch object similar to patches info on rollback and rollforward. Would reduce the need to do follow up patches info call.

nickshorebird (approved):
LGTM overall - Two nits for behavior. Thanks for putting this together.

nickshorebird:
Picked this up to close nits and two bug fixes. Thanks for the initial changes.

PR #3753, opened 2026-05-09
chore(shorebird_ci): make package cleanly publishable

eseidel:
## Summary

- Scores 160/160 on `pana` (was failing the executable, example, and format checks).
- `pubspec.yaml` declares `shorebird_ci` as an executable so `dart pub global activate shorebird_ci` installs the CLI on PATH. Adds `very_good_analysis` as a dev_dep so the lint include resolves in pana's isolated copy.
- New `example/` (README + sample generated workflow) satisfies pana's "package has an example" check.
- `analysis_options.yaml` is now self-contained — no `include: ../../analysis_options.yaml`. The parent isn't shipped on publish, so the include silently failed and dart_style fell back to its defaults, meaning the worktree's enforced format diverged from what pana / pub.dev see. Inlining keeps the formatter config honest in every environment.
- Reformatted 8 source files to match the now-consistently-applied formatter config, and switched two `final foo = {...}` set literals to `final foo = <String>{...}` so `specify_nonobvious_property_types` passes without a rule disable.

The other published packages in the workspace (`jwt`, `scoped_deps`, `shorebird_redis_client`, `shorebird_code_push_protocol`) still have the parent include and the same latent format-divergence bug. Worth a follow-up; not in scope here.

## Test plan

- [x] `pana` reports 160/160
- [x] `dart format --output=none --set-exit-if-changed lib bin` passes
- [x] `dart analyze --fatal-warnings --fatal-infos lib test` clean
- [x] `dart test` — all 136 tests pass
- [x] `shorebird_ci generate --dry-run` against a small fixture matches the checked-in `example/shorebird_ci.yaml`


nickshorebird (approved):
LGTM - Thanks for this

Issue #3619, opened 2026-02-20
fix: Shorebird API/CDN not reachable from Saudi Arabia (patches fail without VPN)


Dev-Salem:
## Description

Shorebird CLI and runtime patch downloads do not work without a VPN from Saudi Arabia.

Observed behavior:

- `shorebird install` fails without VPN
- `shorebird --version` hangs without VPN
- `shorebird patch` fails without VPN
- Shorebird website does not load without VPN
- Runtime patch downloads inside the app do not work without VPN

With VPN enabled, everything works immediately.

This suggests Shorebird API/CDN endpoints may be blocked or partially blocked by ISPs in Saudi Arabia.

This is critical because end users will not use VPN, meaning OTA patches cannot be delivered.

---

## Environment

- Country: Saudi Arabia
- ISP: <STC / Mobily / Zain / Other>
- Network type tested:
  - [x] Home WiFi
  - [x] Mobile data
- macOS version: 26.1 (25B78)
- Flutter version: Flutter 3.38.4 • channel stable
- Shorebird version: Shorebird 1.6.82 (Flutter 3.41.1 )
- Device tested for runtime patch:
  - Android

---

## CLI Error Output

When running:

```
shorebird --version
```

Without VPN, it  just hangs

```
% shorebird doctor
Shorebird 1.6.82 • git@github.com:shorebirdtech/shorebird.git
Flutter 3.41.1 • revision 2706fb5ee5bc4fa2bfb937327192ff9fdfc14b30
Engine • revision c10186c6dcacd5a481b0a0cb6587830cf974e5bd

URL Reachability
⠏ Checking reachability of https://api.shorebird.dev... (119.1s)
```

---

## Runtime Patch Behavior

Steps:

1. Built APK using:
   ```
   shorebird release android --artifact apk
   ```
2. Installed on device.
3. Created patch using:
   ```
   shorebird patch android
   ```
4. Closed and reopened app.

Result:
- Without VPN → Patch does not download.
- With VPN enabled on device → Patch downloads and applies correctly.

This confirms the runtime patch server is not reachable without VPN.

---

### Curl test

```
curl https://api.shorebird.dev
```

Result: hangs without result

## Expected Behavior

Shorebird CLI and runtime patch downloads should work from Saudi Arabia without requiring a VPN.

---

## Question

Are there:

- Alternative CDN endpoints?
- Region-specific mirrors?
- Known regional network restrictions?

[...]

Issue #2483, opened 2024-09-12
fix: api.shorebird.dev is not always accessible in all regions of China


eseidel:
Splitting this off from https://github.com/shorebirdtech/shorebird/issues/435.

Our patch downloads are accessible from China now as part of:
https://shorebird.dev/blog/improved-patch-delivery/

However api.shorebird.dev (which is used for checking if patches are available) is hosted on Google Cloud Run and not reliably accessible.

This affects both companies outside China with users in China as well as developers inside China.

A fix would probably involve making a new patch check endpoint accessible from within China.  Probably would start by using a cloudflare worker (not doing anything specific to China, just moving off of Google Cloud).

rockingdice:
Any progress? I'm picking code-push solutions and this one really blocks it.

eseidel:
We're discussing potential solutions.  Nothing to report at this time.

aprz512:
Any progress? I'm picking code-push solutions and this one is really important.

eseidel:
> Any progress? I'm picking code-push solutions and this one is really important.

We have some plans to move api.shorebird.dev into cloudflare, but are currently focused on other issues.

I'd be very curious to hear more about other solutions you're choosing between.  contact@shorebird.dev also reaches us all.  Thanks!

Vit-boy:
Any progress? I'm picking code-push solutions and this one is really important.

wajipu:
Is there any progress? I am choosing a code push solution, this is very important, recommend several clouds (Alibaba Cloud, Tencent Cloud, etc.) (from: China) @eseidel

eseidel:
We do not have any progress to report at this time.  We have a small team, currently busy with other aspects of the product.  We will update this issue whenever we make further updates to our network structure.  Thanks!

rockingdice:
Thank you, I noticed this one is finally on track. I'm currently using Shorebird for production but with a lot of pain. I have to ask my customers to use VPN to download the patch from the mainland of China and only a small part of them could do that. China is a HUGE market, please consider improving their experience. They only care for better service than money. ;)

zhouzhuo810:
I am in mainland China, and a few days ago I was able to use Shorebird to download patches normally. However, today it suddenly stopped working. Could you please let me know what might be the issue? Is there any solution to this?
@eseidel @rockingdice @felangel @aprz512 @Vit-boy

eseidel:
[...]

Issue #1909, opened 2024-04-16
feat: api.shorebird.dev should be multi-region


eseidel:
Right now api.shorebird.dev is single-region in the US.  (Our actual patch storage is multi-region, but our api endpoint is not).  This causes needless delays for customers in other parts of the world.

aga23a:
> Right now api.shorebird.dev is single-region in the US. (Our actual patch storage is multi-region, but our api endpoint is not). This causes needless delays for customers in other parts of the world.

Some of our Chinese users are unable to upgrade the patch package. How should you deal with this problem?
If we host the server ourselves we won't have this problem.

felangel:
We can add an external load balancer to support multiple regions for api.shorebird.dev

aga23a:
> We can add an external load balancer to support multiple regions for api.shorebird.dev

You can look this https://tool.chinaz.com/dns/api.shorebird.dev
Can you handle the abnormal areas here?

aga23a:
Our products are for the world. Currently, most of our users are in China, and a small number of users cannot update the patch package, so this problem will cause us to lose users.

eseidel:
Access to Shorebird from within China maybe a separate issue from whether Shorebird is hosted multi-regional.  We make no attempt to block access to Shorebird's servers from anyone, however there are parts of the world, which do not allow access to Google Cloud, my understanding is that parts of China may do so, but I'm not an expert in such.

https://github.com/shorebirdtech/shorebird/issues/435 tracks making Shorebird usable from within China.  It's possible that we're close enough today that all that's missing is moving to a CDN: https://github.com/shorebirdtech/shorebird/issues/120.  This specific bug was filed with the intent to reduce latency to Shorebird's servers.

aga23a:
“moving to a CDN”：Is there a specific time for this plan to be implemented?

eseidel:
We just discussed it again amongst the team.  We don't have plans to move to a CDN or multi-region for at least a few months.  (We have a bunch of other applies-to-all-customers work to do first.)

aga23a:
Okay, we know a lot about your work. I also hope this problem can be solved (main work first of course).

Issue #435, opened 2023-05-03
feat: make `shorebird` easier to use by developers in China


eseidel:
Update Sept 11, 2024:  We know of at least some developers using Shorebird successfully from China today.
Users should be able to download published patches through our CDN (without needing to do anything special):
https://shorebird.dev/blog/improved-patch-delivery/
However the updater client also talks to `api.shorebird.dev` which is hosted in Google Cloud and thus may not be accessible.

Developers may also still need to find a way to reach Google Cloud for using `shorebird` tooling such as using a VPN.

=====

The directions for using Shorebird in china include setting FLUTTER_STORAGE_BASE_URL:
https://docs.flutter.dev/community/china

This environment variable is ignored by `shorebird`, since `shorebird` itself uses `FLUTTER_STORAGE_BASE_URL` to pass `download.shorebird.dev` to `flutter`.

The larger issue for developers in China is that Shorebird currently uses Google Cloud as our infrastructure provider and Google Cloud is not reliably reachable across China.

We could implement a proxy within china for `download.shorebird.dev` and either then make it possible to use FLUTTER_STORAGE_BASE_URL to point to such, or some other environment variable.

However we would still need to make access to `api.shorebird.dev` more reliable for Shorebird to work for Developers without them needing to modify their network.

Regardless, Shorebird works seems to work for some users in china (the api.shorebird.dev request sometimes fails, but if it succeeds patch installs otherwise seem to work fine in China).   Developers who cannot reach api.shorebird.dev will still have trouble actually uploading the patches.

xster:
I suspect it's not as simple as setting an alternate URL. The mirror URLs pointed to in https://docs.flutter.dev/community/china physically host the actual bytes to the Flutter engine etc. You'll need to find a partner (who can register a https://en.wikipedia.org/wiki/ICP_license) who can host the app bits being pushed to point the CLI to.

AlexV525:
Hi guys, just noticed this issue when I was browsing new stuff. I can answer the questions above since I'm also a maintainer to flutter.cn .

> I'm not sure if download.shorebird.dev is accessible from China?

Currently, it's available. If you can provide an exact artifact destination then we can test if the connection is stable enough (download.shorebird.dev returns `Unrecognized artifact path`).

> I'm also not sure how we would tell download.shorebird.dev (possibly in the headers?) what proxy to proxy too?

[...]

PR #3756, opened 2026-05-11
feat(code_push_client): add Cloudflare-fronted fallback API endpoint

nickshorebird:
## Summary

When `api.shorebird.dev` becomes unreachable for any reason (regional GCP issues, ISP-level filtering, network-level blocks), every Shorebird CLI command stops working until the upstream issue resolves. This branch fixes that by adding a second endpoint on a separate provider that the CLI falls over to per-request, automatically and w/o user intervention, so a primary outage no longer takes the whole CLI offline.

Tracked in #2483, #435, #1909, #3619.

- Adds `api.shorebird.cloud` (Cloudflare-fronted) as a fallback the CLI tries automatically when the primary fails at the transport layer. Configurable via `SHOREBIRD_HOSTED_URL_FALLBACK`. Defaults to `api.shorebird.cloud`.
- Failover is sticky w/ self-heal. Tries primary first, falls over on a transport-level failure, then promotes whichever host is working so we don't thrash. A session that fell over can heal back to primary if the fallback later fails. At most one swap happens per concurrent burst of failovers, so the post-burst preferred host is deterministic regardless of how many requests piled up.
- Scoped to API hosts only. A host-router decorator keeps GCS uploads, OAuth callbacks, and other third-party hosts on the passthrough client. The failover machinery never sees them.
- Default HTTP client gets a 3s connection timeout so an unreachable primary surfaces as a transport error in seconds instead of hanging. Only covers TCP/TLS handshake, so existing read-timeout behavior is unchanged.
- Multipart-rewrite path refuses to fail over a request w/ attached files, since `MultipartFile` streams are single-use and would silently send empty on retry. API-host multiparts are fields-only today (file uploads target pre-signed GCS URLs through the host-router passthrough), so the guard enforces an existing invariant rather than changing behavior.

## Note

You can test out this endpoint locally by calling `export SHOREBIRD_HOSTED_URL="https://api.shorebird.cloud"`. This will set your API endpoint to the new CF alternative for everything except doctor.

## Architecture

The HTTP client is composed as a chain of `http.BaseClient` decorators (header injector → host router → failover client → IO client) implementing the **Decorator** pattern for the structure and **Latching Failover** for the failover semantics. "Latching" because the working host gets latched as preferred after a successful retry and stays there until it itself fails, rather than re-trying the original primary on every request. Avoids thrash w/o sacrificing per-request recoverability.

## Testing

- Full unit coverage of `_FailoverClient`: primary success path, retry on `SocketException` / `HandshakeException` / `TimeoutException` / `http.ClientException`, sticky promotion, self-heal back to primary, no retry on 4xx, both endpoints failing surfaces the original error, multipart rewrite preserves fields
- E2E tests use the `.invalid` TLD to force fast NXDOMAIN at DNS time, w/ primary/fallback role-swapped variants
- Cross-provider E2E exercises the real Cloudflare-fronted endpoint when `SHOREBIRD_HOSTED_URL_FALLBACK` is set, skipped otherwise. CI updated to set the env var on dev/stable matrix runs
- `dart analyze --fatal-warnings lib test` clean for `shorebird_code_push_client`. Only remaining info is pre-existing from 2024 and unrelated
- `dart format` no diff
- `dart test` green: 92/92 in `shorebird_code_push_client`, affected `shorebird_cli` suites pass

## To do

- Confirm `e2e.yaml` picks up `SHOREBIRD_HOSTED_URL_FALLBACK` on the next dev/stable matrix run and the cross-provider test executes against the real CF endpoint instead of skipping
- Post-deploy: verify a forced-primary-down scenario from a restricted network actually completes via fallback (manual smoke against dev).  DNS checks were already done for reachability.

### DIFF
--- a/packages/shorebird_code_push_client/lib/src/code_push_client.dart
+++ b/packages/shorebird_code_push_client/lib/src/code_push_client.dart
@@ -188,9 +188,14 @@ class _FailoverClient extends http.BaseClient {
       e is TimeoutException ||
       e is http.ClientException;
 
-  /// Returns a copy of [original] with its host swapped to [newHost]. Only
-  /// supports request types we actually send through this client, namely
-  /// [http.Request] and [http.MultipartRequest].
+  /// Returns a copy of [original] with its host swapped to [newHost].
+  ///
+  /// Only supports request types we actually send to API hosts:
+  /// [http.Request] for JSON calls, and [http.MultipartRequest] for
+  /// field-only metadata POSTs (e.g. createPatchArtifact, which uses
+  /// multipart as a form-data envelope and carries no files). Real file
+  /// uploads target signed GCS URLs and do not reach this path because
+  /// [_HostRouter] sends them straight through.
   static http.BaseRequest _rewriteHost(
     http.BaseRequest original,
     String newHost,
--- a/packages/shorebird_code_push_client/lib/src/code_push_client.dart
+++ b/packages/shorebird_code_push_client/lib/src/code_push_client.dart
@@ -86,9 +86,9 @@ class _HeaderInjectingClient extends http.BaseClient {
   }
 }
 
-/// Routes requests by their URL host. Requests whose host is in
-/// [hostsThroughPrimary] are forwarded to [primaryClient]. All other
-/// requests are forwarded to [passthroughClient].
+/// Routes requests by their URL host. Requests whose host is in the
+/// `hostsThroughPrimary` set are forwarded to the primary client. All
+/// other requests are forwarded to the passthrough client.
 ///
 /// Used to direct API calls through a failover-aware client while letting
 /// requests aimed at unrelated hosts (signed GCS upload URLs, third-party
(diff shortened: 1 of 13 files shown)


packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -59,25 +61,208 @@ class CodePushUpgradeRequiredException extends CodePushException {
...
+  /// [http.Request] for JSON calls, and [http.MultipartRequest] for
+  /// field-only metadata POSTs (e.g. createPatchArtifact, which uses
+  /// multipart as a form-data envelope and carries no files). Real file
+  /// uploads target signed GCS URLs and do not reach this path because
+  /// [_HostRouter] sends them straight through.
+  static http.BaseRequest _rewriteHost(
+    http.BaseRequest original,
+    String newHost,
+  ) {
+    final newUri = original.url.replace(host: newHost);
```
easymac, line 219: Looks great to me!

Vanishingly tiny suggestion: maybe support ports here as well? In case we ever need those to differ. But I think we're good as-is.

easymac (approved):

packages/shorebird_code_push_client/lib/src/code_push_client.dart
nickshorebird: Thanks for the suggestion! Going to leave it as a follow-up for now, once it's feature requested or we find we need it. Keeps this PR targeted on the fallback URL change.

PR #3758, opened 2026-05-12
feat(shorebird_cli): prepare for Flutter 3.44 libapp.so strip handover

bdero:
Upstream Flutter PR https://github.com/flutter/flutter/pull/181275 (merged 2026-01-26, first shipped in 3.44) inverted the responsibility for stripping `libapp.so`. Before 3.44, Flutter stripped the AOT shared library itself. From 3.44 onward, Flutter expects AGP to strip it and emit a `libapp.so.sym` companion into the AAB's `BUNDLE-METADATA`. flutter_tools adds a post-build verification that fatal errors when the `.sym` companion is missing.

Two Shorebird flows trip the new check on 3.44:

1. **Obfuscated builds** pass `--extra-gen-snapshot-options=--strip` so gen_snapshot pre-strips `libapp.so`. On 3.44 AGP then has nothing to strip, no `.sym` is produced, build fails.
2. **Existing users** carry a legacy `packaging.jniLibs.keepDebugSymbols.add(\"**/libapp.so\")` line in `android/app/build.gradle(.kts)`. On 3.44 that blocks AGP from stripping, same failure.

This PR lands forward-compatible, version-gated behavior in shorebird_cli so users updating their CLI today are pre-loaded with the right behavior when the actual 3.44 engine bump ships. On older Flutter pins behavior is unchanged.

### shorebird_cli changes

* Add `libappStrippedByAgpConstraint` (>= 3.44) to `flutter_version_constraints.dart`.
* `ShorebirdFlutter.shouldPreStripLibappInGenSnapshot({platform, flutterRevision})`: single helper that owns the strip-gating policy (non-Android always pre-strips; Android pre-strips only below 3.44).
* `Releaser.addObfuscationMapArgs`: on Android+3.44, drop `--strip` so AGP can strip and emit the `.sym`. Other platforms still pre-strip (AGP isn't in their pipeline). The `.sym` in BUNDLE-METADATA is only visible via Play Console and is stripped before APK delivery, so obfuscation protection in user-shipped APKs is preserved. Method becomes `async` because the version check is async.
* `PatchCommand`: same gating for Android patch builds, keyed on `release.flutterRevision` so the patch's gen_snapshot behavior matches the release's.
* New `LegacyKeepDebugSymbolsValidator`: scans `android/app/build.gradle(.kts)` for the legacy `keepDebugSymbols` line on Flutter 3.44+ pins and surfaces a warning pointing at the exact file and substring to remove. Matches both `.add(...)` and `+=` forms, single and double quotes. No-op on older Flutter where the line is still appropriate.
* Registered the validator in `Doctor.androidCommandValidators` and `Doctor.initAndDoctorValidators`.

### artifact_proxy change

Flutter 3.44 also reorganized iOS USB artifact URLs from

```
ios-usb-dependencies/<artifact>/<hash>/<artifact>.zip
```

to

```
ios-usb-dependencies/arm64_x86_64/<artifact>/<hash>/<artifact>.zip
```

(see flutter/flutter#181539 and related). Upstream GCS serves both layouts indefinitely, but the `download.shorebird.dev` proxy's allowlist only recognized the pre-3.44 shape. Without this change every 3.44+ customer hits `404` on their first iOS or Android build during `flutter precache` (libimobiledevice, ios-deploy, etc.). Adds the seven `arm64_x86_64/` variants alongside the existing patterns so customers on either Flutter version resolve.

The companion Flutter-fork patches (revert of the iOS Mach-O dylib PR and demotion of the `libapp.so.sym` check from fatal to warning) are still needed as a safety net for users on truly stale `build.gradle` files; this PR shrinks the population that needs the safety net.

## Test plan

- [x] `dart test test/src/validators/legacy_keep_debug_symbols_validator_test.dart` (10 tests)
- [x] `dart test test/src/shorebird_flutter_test.dart` (full suite plus 4 new tests for `shouldPreStripLibappInGenSnapshot`)
- [x] `dart test test/src/commands/release/ test/src/commands/patch/` (453 tests)
- [x] `dart test test/src/validators/ test/src/commands/doctor_command_test.dart` (103 tests)
- [x] `packages/artifact_proxy` `dart test` (11 existing proxy tests still pass)
- [x] `dart analyze --fatal-warnings lib test` clean
- [ ] After proxy deploy: `curl -sI .../arm64_x86_64/libimobiledevice/<hash>/libimobiledevice.zip` returns 302 to storage.googleapis.com
- [ ] After proxy deploy: `curl -sI .../libimobiledevice/<hash>/libimobiledevice.zip` (no arm64_x86_64 segment) still returns 302

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -462,10 +462,43 @@ Building with Flutter $flutterVersionString to determine the release version...
         '--obfuscate',
         '--extra-gen-snapshot-options='
             '--load-obfuscation-map=${obfuscationMapFile.path}',
-        // Strip unobfuscated DWARF debug info from the compiled snapshot so
-        // it doesn't leak identifiers that obfuscation was meant to hide.
-        '--extra-gen-snapshot-options=--strip',
       ]);
+
+      // On Android with Flutter 3.44+, AGP performs the strip on
+      // libapp.so as part of `stripReleaseDebugSymbols` and produces the
+      // matching `.sym` companion that flutter_tools'
+      // post-build verification requires. Passing --strip to gen_snapshot
+      // here would pre-strip the snapshot, leaving AGP nothing to strip,
+      // and flutter_tools would fail the patch build with "libapp.so.sym
+      // or libapp.so.dbg not present". On older Flutter we keep --strip
+      // because the post-build verification does not exist there.
+      //
+      // The version check uses release.flutterRevision (not the user's
+      // currently-installed pin) because the patch's gen_snapshot
+      // behavior must match the release's. See upstream PR
+      // https://github.com/flutter/flutter/pull/181275 (merged 2026-01-26)
+      // and https://github.com/shorebirdtech/_shorebird/issues/2150.
+      final isAndroid =
+          patcher.releaseType.releasePlatform == ReleasePlatform.android;
+      var shouldPreStripInGenSnapshot = true;
+      if (isAndroid) {
+        final releaseFlutterVersion = await shorebirdFlutter
+            .resolveFlutterVersion(release.flutterRevision);
+        final agpStripsLibapp = libappStrippedByAgpConstraint.isSatisfiedBy(
+          version:
+              releaseFlutterVersion ?? libappStrippedByAgpConstraint.minVersion,
+          revision: release.flutterRevision,
+        );
+        shouldPreStripInGenSnapshot = !agpStripsLibapp;
+      }
+
+      if (shouldPreStripInGenSnapshot) {
+        // Strip unobfuscated DWARF debug info from the compiled snapshot
+        // so it doesn't leak identifiers that obfuscation was meant to
+        // hide. On Android 3.44+ this is handled by AGP instead; see the
+        // block above.
+        extraBuildArgs.add('--extra-gen-snapshot-options=--strip');
+      }
     }
     // Flutter requires --split-debug-info with --obfuscate. Auto-add it
     // if --obfuscate will be in the build args (from the user or from
--- a/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
@@ -464,33 +464,17 @@ Building with Flutter $flutterVersionString to determine the release version...
             '--load-obfuscation-map=${obfuscationMapFile.path}',
       ]);
 
-      // On Android with Flutter 3.44+, AGP performs the strip on
-      // libapp.so as part of `stripReleaseDebugSymbols` and produces the
-      // matching `.sym` companion that flutter_tools'
-      // post-build verification requires. Passing --strip to gen_snapshot
-      // here would pre-strip the snapshot, leaving AGP nothing to strip,
-      // and flutter_tools would fail the patch build with "libapp.so.sym
-      // or libapp.so.dbg not present". On older Flutter we keep --strip
-      // because the post-build verification does not exist there.
-      //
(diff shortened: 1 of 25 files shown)


packages/shorebird_cli/lib/src/validators/legacy_keep_debug_symbols_validator.dart
```diff
@@ -0,0 +1,116 @@
...
+///
+/// The legacy line tells AGP to skip stripping `libapp.so`, which prevents
+/// the `.sym` from being produced and causes the new verification to fail.
+/// On 3.44+ the line is unnecessary as well as actively harmful, so we surface
+/// a warning that points the user at the exact file and substring to remove.
+///
+/// On older Flutter versions the line is still appropriate and this validator
+/// is a no-op.
+///
+/// Tracked in https://github.com/shorebirdtech/_shorebird/issues/2150.
```
bdero: Remove ref

packages/shorebird_cli/lib/src/validators/legacy_keep_debug_symbols_validator.dart
```diff
@@ -0,0 +1,116 @@
...
+class LegacyKeepDebugSymbolsValidator extends Validator {
```
bdero, line 32: I believe vanilla Flutter users will also run into this same surprise opaque build failure, and their AI agents will also flail around for an hours investigating. So considering possibly upstreaming something to Flutter doctor.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
```diff
@@ -462,10 +462,43 @@ Building with Flutter $flutterVersionString to determine the release version...
...
+      if (isAndroid) {
+        final releaseFlutterVersion = await shorebirdFlutter
+            .resolveFlutterVersion(release.flutterRevision);
+        final agpStripsLibapp = libappStrippedByAgpConstraint.isSatisfiedBy(
+          version:
+              releaseFlutterVersion ?? libappStrippedByAgpConstraint.minVersion,
+          revision: release.flutterRevision,
+        );
+        shouldPreStripInGenSnapshot = !agpStripsLibapp;
+      }
```
eseidel: This feels like a helper.

packages/shorebird_cli/lib/src/commands/release/releaser.dart
```diff
@@ -121,18 +121,59 @@ abstract class Releaser {
...
+      final revision = shorebirdEnv.flutterRevision;
+      final version = await shorebirdFlutter.resolveFlutterVersion(revision);
+      final agpStripsLibapp = libappStrippedByAgpConstraint.isSatisfiedBy(
+        version: version ?? libappStrippedByAgpConstraint.minVersion,
+        revision: revision,
+      );
+      // On Flutter 3.44+ AGP performs the strip. Passing --strip here
+      // would pre-strip and trip flutter_tools' post-build verification.
+      shouldPreStripInGenSnapshot = !agpStripsLibapp;
+    }
```
eseidel: Here is your same helper agian.

eseidel (approved):

Gaurav-CareMonitor:
any timeline for 3.44 support, builds are failing after adding support for built in kotlin
https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers

eseidel:
We've prepared it. I expect we'll ship it today.

packages/shorebird_cli/lib/src/commands/patch/patch_command.dart
bdero: Done.

packages/shorebird_cli/lib/src/commands/release/releaser.dart
bdero: Done.

PR #3763, opened 2026-05-13
fix(shorebird_ci): path-derived job keys + has_unit_tests gate

nickshorebird:
## Summary
- Static main workflow derives YAML map keys from each package's path as `<parent_dir>_<package_name>`. Packages can share a `name:` across different directories. The `package_name:` and `package_path:` inputs to the reusable workflow carry the real package values, so codecov flags and `working-directory` are unaffected. Only the in-workflow job identifier and its GH Actions UI rendering pick up the parent-dir prefix.
- Static reusable workflows gate the test and codecov-upload steps on a `has_unit_tests` input. Packages w/o a `test/` directory skip `dart test` / `flutter test` and skip the codecov upload. `has_unit_tests` defaults to `true` so workflows that omit it run tests by default.
- Version 0.2.1.

## Testing
- 2 new test groups covering slug-collision and has_unit_tests, 150/150 passing
- `dart analyze --fatal-warnings` clean
- `dart format` clean
- cspell clean
- 100% diff coverage on changed `lib/` lines


eseidel:
Does this match what the hosted Shorebird CI product does?  I recall a similar problem when trying to run Shorebird CI on the Flutter repo (which uses a bunch of little packages with the same `_` name).  I feel like this will cause the shorebird_ci command to now ignore our non-workspace packages in an otherwise workspace repo?  We intentionally exclude published packages from our workspace, so that they're self-contained for pub for example (e.g. this package it itself excluded).  Will that cause shorebird_ci to ignore testing those excluded packages now?

nickshorebird:
> Does this match what the hosted Shorebird CI product does? I recall a similar problem when trying to run Shorebird CI on the Flutter repo (which uses a bunch of little packages with the same `_` name). I feel like this will cause the shorebird_ci command to now ignore our non-workspace packages in an otherwise workspace repo? We intentionally exclude published packages from our workspace, so that they're self-contained for pub for example (e.g. this package it itself excluded). Will that cause shorebird_ci to ignore testing those excluded packages now?

Thanks for catching this. The workspace filter was overzealous on my part. Dropped it entirely so the analyzer walks the full filesystem again. To avoid the duplicate crash at the yaml key lay I went w/ a much simpler parent prepend to the job key. LMK if you feel otherwise happy to modify.

eseidel (approved):

eseidel:
Slightly better would be just do de-dup the names if necessary and otherwise leave them short.  I wrote a thing for thsi recently for a personal project: https://github.com/eseidel/space_gen/blob/main/lib/src/naming.dart. (We don't need something that complicated of course).

nickshorebird:
> Slightly better would be just do de-dup the names if necessary and otherwise leave them short. I wrote a thing for thsi recently for a personal project: https://github.com/eseidel/space_gen/blob/main/lib/src/naming.dart. (We don't need something that complicated of course).

Thanks for the feedback. Went w/ the suggestion even though it means naming can be changed later w/ package adds as it keeps known names. Retested it to confirm working E-E.

PR #3766, opened 2026-05-14
fix(shorebird_cli): isInteractive falls back to stdioType for unsized PTYs

### LINKED ISSUE #3765: CLI: prompt gate rejects PTYs with default 0x0 winsize
## Problem

When running the `--confirm` flow through Claude Code's Bash tool (which has no parent TTY) rather than from a real terminal, the gate added in #3720 throws `InteractivePromptRequiredException` even though, by every measure other than `Stdout.hasTerminal`, the child's stdout is a real terminal. The proximate cause is a Dart quirk (`hasTerminal` checks winsize, not isatty; details below), but the choice of gate is what makes it land on Shorebird users, and the failure mode is invisible to anyone who only ever runs the CLI from an interactive shell.

## Background

I encountered this driving Fastlane via Claude Code, running [@gyuri-fluttech's `--confirm` workaround](https://github.com/shorebirdtech/shorebird/issues/3223#issuecomment-3248458151) for ["Binary Mismatch on macOS After App Store Distribution Using .pkg" #3223](https://github.com/shorebirdtech/shorebird/issues/3223). Investigating reminded me that `--confirm` doesn't help my Mac App Store distribution case at all; my real fix is the [MAS-artifact reupload workaround I documented in #3223](https://github.com/shorebirdtech/shorebird/issues/3223#issuecomment-4343839357), so `--confirm` is on its way out of my Fastfile.

## Root cause

The gate uses `stdout.hasTerminal`, which in Dart is implemented as `_getTerminalSize(fd) is List` (`sdk/lib/_internal/vm/bin/stdio_patch.dart`), calling `ioctl(TIOCGWINSZ)` and returning false when both `ws_col` and `ws_row` are zero (`runtime/bin/stdio_macos.cc`, `Stdout::GetTerminalSize`). A pty allocated by `PTY.spawn` defaults to a 0x0 winsize, so the gate throws on what is otherwise a valid pty: C's `isatty(1)`, Ruby's `STDOUT.tty?`, and Dart's own `stdioType(stdout)` all report fd 1 is a terminal.

The triggering condition is structural: a pty allocated from a parent that has no controlling tty of its own (CI runners, agent harnesses, the Fastlane wrapper Claude Code drove for me). Interactive shells don't reach it, because the terminal emulator sizes the controlling tty at startup and that winsize propagates through pty descendants the shell spawns.

## Proposed fix

When `stdout.hasTerminal` returns false, fall back to `stdioType(stdout) == StdioType.terminal` (backed by stat + `isatty()`) before declaring the context non-interactive. `hasTerminal` stays as the fast path so existing test infrastructure (the `CapturingStdout.hasTerminalOverride` in `test/src/helpers.dart`) keeps working unmodified, and behavior is unchanged for sized terminals.

## Reproduction

My Fastlane lane wraps `shorebird release macos --confirm` in `script` + `PTY.spawn` so it can answer Shorebird's prompt programmatically (and substitutes MAS `.pkg` building for codesign/notarytool/staple in the paused window). From Ghostty the lane worked; from Claude Code's Bash tool (no parent TTY), Shorebird threw at the prompt. Workaround locally: set `stdout.winsize = [40, 120]` on the pty handle before reading. Verified against Shorebird 1.6.98 / Dart 3.11.5.

lukemmtt:
## Summary

Fixes #3765. When `stdout.hasTerminal` returns false, fall back to `stdioType(stdout) == StdioType.terminal` before declaring the context non-interactive.

`Stdout.hasTerminal` is implemented in Dart as `_getTerminalSize(fd) is List`: it calls `ioctl(TIOCGWINSZ)` and returns false when both `ws_col` and `ws_row` are zero. PTYs allocated without an explicit winsize (Ruby's `PTY.spawn` and equivalents in other languages) default to 0x0 and fail that check despite `isatty(fd)` being true. The gate added in #3720 then throws `InteractivePromptRequiredException` on prompts despite stdout being a valid terminal.

`stdioType(stdout) == StdioType.terminal` is backed by stat-based handle classification plus `isatty()`, which matches the gate's intent. Adding it as a fallback means:

- Sized terminals still hit the fast `hasTerminal` path, so behavior is unchanged.
- Unsized PTYs (the bug) now resolve to the fallback and return true.
- Non-terminals (pipes, files, /dev/null) still fail both checks, so behavior is unchanged.

## Why a fallback rather than a swap

A clean swap (`isInteractive = stdioType(stdout) == StdioType.terminal && !isJsonMode`) is the architecturally cleaner shape, but would have required updates to `test/src/helpers.dart`. The existing `CapturingStdout.hasTerminalOverride` mock affects `Stdout.hasTerminal` (overridable via `IOOverrides`), but `stdioType()` is a top-level function backed by the real fd, so it can't be mocked through `IOOverrides`. Three tests in `shorebird_cli_command_runner_test.dart` would have broken because their `hasTerminal: true` overrides would no longer flow through to `isInteractive`.

Happy to do the swap and add an injection point (a `@visibleForTesting` `stdioTypeProbe` field, or a scoped-deps injection) if maintainers prefer; this PR keeps the change surgical.

## Testing

- `dart test test/src/logging/shorebird_logger_test.dart`: 21 tests pass.
- `dart test test/src/shorebird_cli_command_runner_test.dart`: 38 tests pass, including the three `interactive mode` tests that exercise `isInteractive` via the runner.

The new fallback branch (`hasTerminal=false` AND `stdioType=terminal`) isn't directly covered by these unit tests because the test infrastructure only mocks `hasTerminal` and `stdioType()` reads the real fd 1 (which `dart test` typically pipes). If explicit coverage is wanted, the path forward is the injection point mentioned above, and I'm happy to add it.

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -57,6 +57,7 @@ words:
   - ints
   - iokit
   - iphoneos
+  - isatty
   - keyalg # From .github/workflows/e2e.yaml
   - keypass # From .github/workflows/e2e.yaml
   - keysize # From .github/workflows/e2e.yaml
@@ -154,6 +155,7 @@ words:
   - VCRUNTIME
   - Verdana
   - vmcode
+  - winsize
   - worktrees
   - writeln
   - WRONGPASS
--- a/packages/shorebird_cli/lib/src/interactive_mode.dart
+++ b/packages/shorebird_cli/lib/src/interactive_mode.dart
@@ -9,9 +9,17 @@ import 'package:shorebird_cli/src/json_output.dart';
 /// (`chooseOne`/`confirm`/`prompt`).
 ///
 /// Returns `false` when any of the following is true:
-///   * stdout is not connected to a terminal (`!stdout.hasTerminal`).
+///   * stdout is not connected to a terminal.
 ///   * `--json` was passed (machine-readable output expected).
-bool get isInteractive => io.stdout.hasTerminal && !isJsonMode;
+///
+/// `Stdout.hasTerminal` is winsize-based, not isatty-based, so PTYs with
+/// default 0x0 winsize fail it. `stdioType(stdout) == StdioType.terminal`
+/// uses isatty semantics and catches them.
+bool get isInteractive {
+  if (isJsonMode) return false;
+  if (io.stdout.hasTerminal) return true;
+  return io.stdioType(io.stdout) == io.StdioType.terminal;
+}
 
 /// The default hint emitted when an interactive prompt is reached in a
 /// non-interactive context and no per-site hint was provided.


nickshorebird (approved):
LGTM - Thanks for the extremely fast fix. Fine w/ skipping the injection and keeping it surgical as I do not expect behavior changes in `stdioType` semantics and can revisit at that time or add it in my free time.

PR #3774, opened 2026-05-15
feat(shorebird_cli): warn by provider when SHOREBIRD_TOKEN is legacy

eseidel:
## Summary

Replaces the generic "legacy CI token" warning in `auth.dart` with one gated on the backing OAuth provider:

- **Shorebird-issued `CiToken`** (`authProvider == shorebird`) — no longer warns. These are part of the new auth system.
- **Google- or Microsoft-backed `CiToken`** — warns with a single shared message:

  > SHOREBIRD_TOKEN is a legacy token and will stop working in a future release.
  > Upgrade to a Shorebird API key for better control, visibility, and revocation.
  > Create one at https://console.shorebird.dev and replace SHOREBIRD_TOKEN with it.
  > Learn more: https://shorebird.dev/blog/introducing-shorebirds-upgraded-auth-service

The warning now points users at the actual fix — creating a new API key at `console.shorebird.dev` (since `shorebird logout`/`login` doesn't help when the env var is still set) — and links to the upgrade blog post.

## Blocker before landing

The blog link points at `https://shorebird.dev/blog/introducing-shorebirds-upgraded-auth-service`, which **does not exist yet**. The post needs to be published (or the slug confirmed) before this merges.

## Test plan

- [x] `dart test test/src/auth/auth_test.dart` — 48/48 pass.
- [x] Updated existing Google-CiToken test to match new wording.
- [x] New test: Microsoft-backed CiToken — warning fires with the same wording.
- [x] New test: Shorebird-issued CiToken — `verifyNever(logger.warn)`.
- [x] `dart analyze --fatal-warnings lib/src/auth/auth.dart test/src/auth/auth_test.dart` — only pre-existing `applicationConfigHome` deprecation infos.

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -360,16 +360,10 @@ class Auth {
         return;
       }
 
-      // Legacy CiToken format — still supported, but deprecated.
+      // CiToken format — Shorebird-issued tokens are still current; tokens
+      // backed by a Google or Microsoft account are legacy.
       try {
         _token = CiToken.fromBase64(trimmed);
-        logger.warn(
-          'SHOREBIRD_TOKEN contains a legacy CI token from '
-          '`shorebird login:ci`. '
-          'This format is deprecated and will stop working in a future '
-          'release. '
-          'Create an API key at https://console.shorebird.dev instead.',
-        );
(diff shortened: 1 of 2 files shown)


eseidel:
We should write a blog post with instructions on how to transition before we ship this.  We can update this to point to that blog post as well.

PR #3775, opened 2026-05-15
feat(shorebird_cli)!: hard-error on `shorebird login:ci`

eseidel:
## Summary

`shorebird login:ci` stopped issuing tokens months ago — its `run()` method has been a print-only no-op that exits 0. Make it an actual error:

- Log via `logger.err` (red) instead of `logger.info` (white).
- Exit with `ExitCode.usage.code` (64) instead of `ExitCode.success.code`.
- Replace "is deprecated … existing tokens will continue to work for now" with "has been replaced by API keys" — the command itself never produced anything long-lived; existing CI tokens are minted server-side and unaffected.
- Add a docs link alongside the console link so users can read the full migration story.

New output:

```
shorebird login:ci has been replaced by API keys.

Create an API key at https://console.shorebird.dev and set it as your SHOREBIRD_TOKEN environment variable.

Learn more: https://docs.shorebird.dev/account/api-keys/
```

## Breaking change

Scripts that called `shorebird login:ci` expecting exit 0 will now fail. Anything relying on that path needs to swap to a console-minted API key. The actual `SHOREBIRD_TOKEN` env-var contract is unchanged.

## Test plan

- [x] `dart test test/src/commands/login_ci_command_test.dart` — 4/4 pass.
- [x] Updated existing test for the new exit code, new `logger.err` channel, and the new message + docs link.

## Related

Pairs with #3774 (provider-aware warning on legacy `SHOREBIRD_TOKEN` in CI).


eseidel:
It was already deprecated, so maybe this isn't actually doing much.

easymac (approved):

PR #3781, opened 2026-05-18
feat(protocol): Add prefetch candidate release versions

### LINKED ISSUE #3755: feat: prefetch patches for server-advertised future releases
## Problem

When a device auto-updates the native binary to a newer release that already has patches published, those patches aren't on disk yet. On first launch of the new binary, the user runs unpatched code even though a fix exists. This happens because [`/api/v1/patches/check`](https://github.com/shorebirdtech/updater/blob/main/library/src/network.rs#L16) is scoped to a single [`release_version`](https://github.com/shorebirdtech/updater/blob/main/library/src/network.rs#L242); patches for any other release on the same track aren't fetched until the user is already running that release.

<p align="center">
  Image
</p>

Concrete scenario:

Consider an app that has version 9.0.0 in production, and *9.0.1 is pending App Store review*.

- A user reports a data-loss bug that is affecting the production release, 9.0.0.
- Dev releases a hotfix for the bug, as patch 9.0.0-p1. Since 9.0.1 is already pending review, rather than repackaging the release and resubmitting, they patch that release as well, with 9.0.1-p1.
- Dev tells user to restart to apply patch 9.0.0-p1; user does so and is pleased with the fix. User goes to bed.
- Meanwhile, v9.0.1 is approved by App Review and gets installed as an automatic update for the user overnight.
- User boots v9.0.1 in the morning. Since it's the first launch of that release, and the shorebird client didn't pre-download the v9.0.1-p1 patch ahead of time, it's not available to be applied on that first launch of 9.0.1. The update checker downloads it in the background, and it's ready to apply (pending restart), but the user doesn't know that.
- As a result, the user encounters the "fixed bug" again.
  - User says: "My data is gone again; I thought you said you fixed it? what do you mean the patch 'un-applied?'", etc.
  - Result: Broken trust and confusion (justifiably)

## Proposal

Pre-fetch patches for server-advertised release versions on the same track, so when the OS swaps in a new native binary, the patch can apply on first launch instead of the user booting unpatched code.

<p align="center">
  Image
</p>

Nuances:
  - Benefits existing users on every native update, in addition to the emergency-hotfix scenario.
  - Doesn't depend on solving the harder fresh-install-boots-unpatched problem ([#741 ("feat: install patches without having to restart the app")](https://github.com/shorebirdtech/shorebird/issues/741), consolidated into [#2350 ("feat: enable advanced users to implement patch-install-without-restart")](https://github.com/shorebirdtech/shorebird/issues/2350)).
- This isn't updater-only, but the updater already has most of the local primitives this needs: the engine asks the updater which patch to boot, patch state is release-scoped, and boot validation happens against the running binary.
  - The missing pieces appear to be: server/API support to advertise server-selected prefetch candidates for the same app/track/platform/architecture, public protocol support for that response field, updater storage that survives the release-version reset, and updater promotion logic that validates the prefetched delta after the matching native binary is running.
  - The engine ([`shorebird_next_boot_patch_path()`](https://github.com/shorebirdtech/engine/blob/shorebird/dev/shell/common/shorebird/shorebird.cc#L82-L86)) consumes whatever path the updater designates, so a patch designated during initialization can be picked up on launch.
  - [Strict-mode boot validation re-verifies the signature at launch](https://github.com/shorebirdtech/updater/blob/main/library/src/cache/lifecycle.rs#L80-L82) using the running binary's public key. In this proposal, the current release stores only the compressed delta for a server-advertised future release, and the future release promotes/validates it only after its native binary is actually running.
  - The disk schema is already release-scoped, but prefetched compressed deltas need a location that survives the release-version reset.
  - One concrete blocker is the [unconditional wipe-on-mismatch](https://github.com/shorebirdtech/updater/blob/main/library/src/cache/updater_state.rs#L187-L207) in `load_or_new_on_error()`, which deletes patches when `release_version` changes. The updater draft PR preserves a dedicated prefetch area across that reset and promotes matching bytes on first init of the new release.
- Default-on, with an opt-out flag for unusual cases (e.g. `prefetchUpcomingPatches: false` for metered networks or embedded devices). The cost is small and bounded; the benefit applies to every native update for every user; opt-in adoption would leave most apps vulnerable to the regression by accident.
  - Per-patch bandwidth is small: Android tens of KB to ~166 KB ([#2737 ("feat: make patch sizes smaller on Android")](https://github.com/shorebirdtech/shorebird/issues/2737)), iOS hundreds of KB to a few MB ([#811 ("feat: Make patch sizes smaller on iOS")](https://github.com/shorebirdtech/shorebird/issues/811)).
  - The simplest cache policy bounds disk usage to ~2 patches at any time: keep the patch for the current release plus the patch for the latest server-advertised release on the track.
    - Self-limits across release churn: a user on `9.5.0-p1` with `9.5.1-p1` pre-fetched will see `9.5.1-p1` quietly replaced by `9.5.2-p1` when the dev ships 9.5.2.
    - Even a user squatting on an old native binary indefinitely costs no more than one extra patch's worth of disk.
  - The regression we're avoiding is invisible to the user; they just see a previously-fixed bug come back.

## Draft PRs

- [shorebirdtech/updater#357](https://github.com/shorebirdtech/updater/pull/357): client-side draft/proof-of-concept for preserving prefetched patch bytes across release reset and promoting them when the matching release first runs.
- [shorebirdtech/shorebird#3781](https://github.com/shorebirdtech/shorebird/pull/3781): public protocol sketch for advertising prefetch candidate release versions. This is closed for now pending discussion here, but included to show the shape of the protocol piece.

Both drafts assume private server/API support to populate the prefetch candidates, or an equivalent server-owned mechanism.

## Related

- [#1980 (open, "feat: expose a way to detect newer releases")](https://github.com/shorebirdtech/shorebird/issues/1980): Complementary, dev-side; with this proposal you wouldn't need to nag users to update just to avoid this patch-continuity gap.
- [#2350 (open, "feat: enable advanced users to implement patch-install-without-restart")](https://github.com/shorebirdtech/shorebird/issues/2350): Complementary at the in-session level. This proposal addresses what happens *across native binary swaps*.
- [#3702 (merged, "feat(protocol): add `current_patch_number` to PatchCheckRequest")](https://github.com/shorebirdtech/shorebird/pull/3702): Shows the protocol is already evolving in this area.
- [#3749 (open, "feat: ShorebirdUpdater.watchForUpdates, managed polling for manual-update apps")](https://github.com/shorebirdtech/shorebird/issues/3749): Today scoped to the current release; this proposal would naturally extend it.

lukemmtt:
Part of a proposed solution to close shorebirdtech/shorebird#3755, along with shorebirdtech/updater#357.

## Summary

Adds an optional `available_release_versions` field to `PatchCheckResponse`, exposed in Dart as `availableReleaseVersions`, so the server can advertise same app/channel/platform/arch release versions that clients may prefetch before a native binary update.

This is intentionally a draft because `shorebird_code_push_protocol` is generated from the public OpenAPI schema, and the live schema at https://api.shorebird.dev/openapi.json does not expose this field yet. The final version should likely come from the server/OpenAPI source of truth and then regenerate this package.

## Why this exists

The updater-side proof of concept in shorebirdtech/updater#357 can consume server-advertised prefetch candidates, but the updater cannot safely infer future release versions on its own. Release versions are app/store strings, and the backend should own candidate selection for the same track.

## Scope

- Adds `PatchCheckResponse.availableReleaseVersions`.
- Includes the field in JSON serialization/deserialization, equality, and hash code.
- Adds tests for JSON behavior and parsing.

## Still required elsewhere

- Private API/server work to populate `available_release_versions`, or an equivalent server-owned candidate field.
- Public OpenAPI regeneration so this package is updated from the actual schema source.
- Updater-side client behavior in shorebirdtech/updater#357.

## Test plan

- `dart format --set-exit-if-changed packages/shorebird_code_push_protocol/lib/src/messages/patch_check/patch_check_response.dart packages/shorebird_code_push_protocol/test/src/messages/patch_check/patch_check_response_test.dart`
- `dart analyze packages/shorebird_code_push_protocol`
- `dart test packages/shorebird_code_push_protocol`

Local result: `dart test packages/shorebird_code_push_protocol` passes with 233 tests.

### DIFF
--- a/packages/shorebird_code_push_protocol/lib/src/messages/patch_check/patch_check_response.dart
+++ b/packages/shorebird_code_push_protocol/lib/src/messages/patch_check/patch_check_response.dart
@@ -12,6 +12,7 @@ class PatchCheckResponse {
     required this.patchAvailable,
     this.patch,
     this.rolledBackPatchNumbers,
+    this.availableReleaseVersions,
   });
 
   /// Converts a `Map<String, dynamic>` to a [PatchCheckResponse].
@@ -26,6 +27,8 @@ class PatchCheckResponse {
         ),
         rolledBackPatchNumbers: (json['rolled_back_patch_numbers'] as List?)
             ?.cast<int>(),
+        availableReleaseVersions: (json['available_release_versions'] as List?)
+            ?.cast<String>(),
       ),
     );
   }
@@ -50,12 +53,17 @@ class PatchCheckResponse {
   /// release.
   final List<int>? rolledBackPatchNumbers;
 
+  /// Server-selected release versions on the same app/channel/platform/arch
+  /// that the client may prefetch before the native binary updates.
+  final List<String>? availableReleaseVersions;
+
   /// Converts a [PatchCheckResponse] to a `Map<String, dynamic>`.
   Map<String, dynamic> toJson() {
     return {
       'patch_available': patchAvailable,
       'patch': patch?.toJson(),
       'rolled_back_patch_numbers': rolledBackPatchNumbers,
+      'available_release_versions': availableReleaseVersions,
     };
   }
 
@@ -64,6 +72,7 @@ class PatchCheckResponse {
     patchAvailable,
     patch,
     listHash(rolledBackPatchNumbers),
+    listHash(availableReleaseVersions),
   ]);
 
   @override
@@ -72,6 +81,7 @@ class PatchCheckResponse {
     return other is PatchCheckResponse &&
         patchAvailable == other.patchAvailable &&
         patch == other.patch &&
-        listsEqual(rolledBackPatchNumbers, other.rolledBackPatchNumbers);
+        listsEqual(rolledBackPatchNumbers, other.rolledBackPatchNumbers) &&
+        listsEqual(availableReleaseVersions, other.availableReleaseVersions);
   }
 }
--- a/packages/shorebird_code_push_protocol/test/src/messages/patch_check/patch_check_response_test.dart
+++ b/packages/shorebird_code_push_protocol/test/src/messages/patch_check/patch_check_response_test.dart
@@ -26,6 +26,7 @@ void main() {
         'patch_available': true,
         'patch': null,
         'rolled_back_patch_numbers': null,
+        'available_release_versions': null,
       });
     });
 
@@ -81,6 +82,7 @@ void main() {
           'hash_signature': null,
         },
         'rolled_back_patch_numbers': null,
+        'available_release_versions': null,
       });
     });
 
@@ -104,6 +106,7 @@ void main() {
             'hash_signature': 'signature',
           },
           'rolled_back_patch_numbers': null,
+          'available_release_versions': null,
         });
       });
     });
@@ -118,6 +121,7 @@ void main() {
           'patch_available': true,
           'patch': null,
           'rolled_back_patch_numbers': [1, 2, 3],
+          'available_release_versions': null,
         });
       });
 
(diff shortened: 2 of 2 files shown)


eseidel:
I think we should discuss on the issue more.  I'm not convinced this is the right solution.

eseidel:
You can obviously re-open, just doing our weekly triage and was clearing out the pulls queue.

PR #3785, opened 2026-05-19
feat(shorebird_ci): reserve required name, verify aggregator coverage

nickshorebird:
## Summary

The `required` aggregator job from #3780 is meant to be the single check listed in branch protection. Two silent-failure modes left that gate weaker than it looks:

1. A package slug that resolves to `required` would emit a duplicate YAML key and silently overwrite the aggregator. Branch protection would then gate on the per-package job that took its place, not on aggregate coverage.
2. A hand-edited workflow could add a top-level job and forget to add it to `required.needs`. The new job would run, but its status would not be part of the merge gate.

I'm closing both. `generate --required` now refuses to generate when a package slug resolves to `required`, so the duplicate key never lands. `verify` now enforces that any workflow w/ a top-level `required:` job lists every other top-level job in its `needs:`, and that every entry in `needs:` matches a real job in the same file. A `required:` key w/ no map body is reported as MALFORMED rather than silently skipped, which previously let an obviously broken aggregator pass the merge gate's lint.

The collision check is scoped to `--style static`. Dynamic mode keys jobs by `setup` / `dart_ci` / `flutter_ci` / `cspell`, never by per-package slug, so a package named `required` cannot collide w/ the aggregator there, and rejecting it would be over-broad. Slugs are computed once in `run()` and threaded through to the static builders rather than being computed twice.

Alongside: bad CLI args now print the usage message and exit 64 instead of bubbling a Dart stack trace, matching the convention the `args` package itself documents. The `--required` flag help and the `verify` description are rewritten to cover the new contract, and the `--no-update-actions` help is rewrapped at col 30 to match the other multi-line flags.

## Testing

- `dart test packages/shorebird_ci`: 180 passing, including new cases for each failure mode the checks guard against, slug collision, missing `needs`, stale `needs`, scalar-string `needs`, no `needs` key, malformed `required:` body, and collision-check no-op in dynamic mode.
- Generated `--style static --required` and `--style dynamic --required` against a multi-package fixture, then ran `verify` on the output to confirm freshly generated workflows pass the new rules.
- `shorebird_ci generate --help` rendered locally to confirm `--no-update-actions` now wraps at col 30 like its peers.
- Ran the generated dynamic workflow against a Flutter app to confirm `setup` runs end-to-end through `shorebird_ci verify` and `shorebird_ci affected_packages --all`, emitting the correct matrix payload for `flutter_ci`.


easymac (approved):
👍 LGTM.

Issue #3388, opened 2025-11-19
fix: PathNotFoundException : `shorebird release` ignores filtered architectures for Android project (`abiFilters`) 


gerken-tss:
App ID: be1fc06b-32af-4b20-88d4-3ffbfe2d60d5

**Description**

I replaced my previous build command:

```bash
flutter build appbundle \
          --release \
          --flavor "$FLAVOR_NAME" \
          --build-number="$(< $CM_BUILD_DIR/android/updates_build_number.txt)"
```

with this shorebird command:

```bash
 shorebird release android \
          --flavor "$FLAVOR_NAME" \
          --build-number "$(< $CM_BUILD_DIR/android/updated_build_number.txt)"
```

I get the following output, though:

> ✓ Built build/app/outputs/bundle/xxxRelease/app-xxx-release.aab (46.2MB)
> ⠋ Determining release version...
> ✓ Release version: 2.4.11+170 (2.2s)
> ⠋ Fetching releases...
> ✓ Fetching releases (0.2s)
> 
> 🚀 Ready to create a new release!
> 
> 📱 App: xxx (xxx) (6b2085ca-36d1-44f1-922d-e6aefd3f4399)
> 🍧 Flavor: xxx
> 📦 Release Version: 2.4.11+170
> 🕹️  Platform: android
> 🐦 Flutter Version: 3.38.1 (4686ea0143)
> 
> ⠋ Fetching releases...
> ✓ Fetching releases (0.1s)
> ⠋ Creating release...
> ✓ Creating release (1.0s)
> ⠋ Updating release status...
> ✓ Updating release status (0.1s)
> ⠋ Uploading artifacts...PathNotFoundException: Cannot open file, path = '/Users/builder/clone/build/app/intermediates/stripped_native_libs/xxxRelease/stripxxxReleaseDebugSymbols/out/lib/armeabi-v7a/libapp.so' (OS Error: No such file or directory, errno = 2)
> 
> If you aren't sure why this command failed, re-run with the --verbose flag to see more information.
> 
> You can also [file an issue](https://github.com/shorebirdtech/shorebird/issues/new/choose) if you think this is a bug. Please include the following log file in your report:
> /Users/builder/Library/Application Support/shorebird/logs/1763552052573_shorebird.log

I don't know how relevant this is, but this all runs in a Codemagic pipeline. So temporary local changes on my machine can't be the cause.

**Expected Behavior**

[...]

PR #3793, opened 2026-05-20
fix(shorebird_cli): skip missing-arch libapp.so on Android upload and patch

bdero:
## Summary

- Skip arches with no `libapp.so` on disk in both the release upload loop (`code_push_client_wrapper.dart`) and the patch diff loop (`android_patcher.dart`), and error out clearly when every arch is missing.
- Unblocks projects that filter ABIs via `ndk.abiFilters` / `splits.abi` / `jniLibs.excludes`, which were crashing with `PathNotFoundException` on the first missing arch (issue #3388).

## Why now

This was masked pre-Flutter 3.44 because `gen_snapshot --strip` caused AGP to skip its strip task, falling back to a path that contained every arch regardless of filters. PR #3758 stopped passing `--strip` on 3.44+, so AGP now runs its strip task and only emits directories for archs that pass the project's filters. Filtered projects upgrading to 3.44 hit the crash on the next release/patch.

## Test plan

- [x] Added unit test: release upload skips the missing arch and uploads the rest.
- [x] Added unit test: release upload exits with code 70 + clear message when every arch is missing.
- [x] Added unit test: patch flow skips the missing arch.
- [x] Added unit test: patch flow exits with code 70 + clear message when every arch is missing.
- [x] Full `dart test` suite passes (2053 tests).

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -497,6 +497,14 @@ Looked in:
       );
     }
 
+    // Track which arch paths AGP didn't produce, so we can surface them at
+    // the end if zero archs uploaded. AGP omits an arch directory whenever
+    // the project filters it out via `ndk.abiFilters`, `splits.abi`, or
+    // `jniLibs.excludes`. Iterating only over present files lets a filtered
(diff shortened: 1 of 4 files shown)


nickshorebird (approved):
LGTM

PR #3798, opened 2026-05-22
feat(shorebird_code_push_protocol): add version-distribution API surface

easymac:
## Motivation

This is the protocol half of an end-to-end **release version distribution** metric being built in the private monorepo — a steel thread through to a chart on the customer Insights page. The new `GET /apps/{appId}/metrics/version-distribution` endpoint needs typed request/response DTOs in the shared protocol package, which both the server and the API clients regenerate against. This PR adds that surface.

## What's in here

Regenerated from the OpenAPI spec via `space_gen` (`tool/gen.dart`):

- `GetVersionDistributionResponse` + `VersionDistributionEntry` models
- their generated round-trip tests under `test/generated/`
- `checkedKey()` in `model_helpers.dart`, emitted for the required-but-nullable `release_version` field

Top-level barrel exports added by hand, per the package convention that the barrel is not generated.

## Notes

Consumed by the private `_shorebird` PR for this feature. Landing this
unblocks pinning the `_shorebird` submodule to a merged `main` commit.


packages/shorebird_code_push_protocol/lib/model_helpers.dart
```diff
@@ -20,6 +20,18 @@ T parseFromJson<T>(
...
 }
 
+/// Reads a key that is required to be present in [json] but whose value
+/// may legitimately be null (OpenAPI 3.1 `type: [T, "null"]` combined
+/// with `required`). A plain `json[key] as T?` cast would otherwise
+/// accept a missing key as a null value, silently violating `required`.
+/// Throws [FormatException] when the key is absent.
+dynamic checkedKey(Map<String, dynamic> json, String key) {
+  if (!json.containsKey(key)) {
+    throw FormatException("Missing required key '$key'", json);
```
nickshorebird, line 30: Take it or leave it - No test covers `null` passing / what this is meant to fix. Value returns value is there hence the nit.

nickshorebird (approved):
LGTM

PR #3808, opened 2026-06-03
fix(shorebird_cli): read libapp.so from the aab instead of the strip task output

bdero:
## Problem

Shorebird needs `libapp.so` (the compiled Dart code) for each architecture when creating Android releases and patches. It previously read these files from the Android Gradle Plugin's strip task output under `build/app/intermediates/stripped_native_libs/`.

Flutter 3.44 (flutter/flutter#181275) moved `libapp.so` stripping from Flutter to the Android Gradle Plugin. In some projects the strip task no longer emits `libapp.so`, even though the library is still packaged into the final app bundle (the `.aab`). This breaks the old approach in two ways:

1. The strip output is empty, and `shorebird release android` and `shorebird patch android` fail with errors like "every requested architecture was missing".
2. Worse, the strip output can contain leftovers from a previous build. The commands then succeed but quietly use the old library. For patches this produces a patch with no changes in it: the patch installs and runs on devices, but users never see the new code.

A fresh `flutter create` app does not hit this, so the behavior is specific to certain project or plugin configurations. The fix does not depend on the exact trigger.

## Fix

Shorebird now reads `libapp.so` directly from the built `.aab`. The `.aab` is the exact artifact installed on devices, so the extracted libraries always match what ships, and a patch diffed against them always links at runtime. This sidesteps the strip task entirely, fixing both failure modes, and matches how the AAR flow already sources its libraries. The strip output is kept only as a fallback for when the `.aab` cannot be read.

The unstripped copies in `merged_native_libs` are deliberately not used: those bytes differ from what ships inside the `.aab`, and patches diffed against them would fail to install. See #3388.

## Testing

- Unit tests cover the extraction and selection logic plus the release upload and patch flows, using a real `.aab` fixture, including fallback behavior for unreadable bundles.
- Verified on builds across Flutter 3.44.0 and 3.44.1, both default and `--obfuscate`: the `.aab`'s `libapp.so` is always byte-identical (sha256) to the strip output when a valid one exists, and differs from `merged_native_libs`, confirming the `.aab` is the safe source.
- A run using the extraction path produced a release and patch that installed and ran on a physical device with no link failures.


eseidel (approved):
I think this is fine. I would probably reverse the order and read from the AAB first or always, but that could be a separate patch.

Reading from the AAB always feels like it's always correct (and matches what our AAR path does) and should be both forward and backward compatible.  I just picked this shortcut when originally writing `shorebird release` and never changed it.

bdero:
Agreed & done, flipped the checking order.

PR #3809, opened 2026-06-03
feat(shorebird_code_push_protocol): add patch-adoption API surface

easymac:
Regenerate the protocol DTOs (`space_gen`) against the local `code_push_server` `openapi.yaml` `getPatchAdoption` endpoint:

- `GetPatchAdoptionResponse`, `PatchAdoptionEntry`, `PatchAdoptionPoint`, `PatchAdoptionRange`
- inline-granularity `GetPatchAdoptionParameter3` enum
- `model_helpers` gains `maybeParseDateTime`/`maybeParseDate` for the nullable `period` field
- hand-added the new exports to the barrel (the generator leaves it untouched)
- unrelated regen drift on `private_user` reverted to keep the diff scoped

Consumed by the `_shorebird` patch-adoption stack (re-pins this submodule onto this commit). Draft — opening to unblock CI on the parent stack.


easymac:
Looks like a space_gen substring bug is causing the code coverage problem. Seems to be injecting a `maybeParseDate` instead of `maybeParseDateTime`, and then not testing it because it's not actually used.

Happy to file a patch.

eseidel:
We should also feel no obligation to use space_gen.  Or just modify the bindings with an llm, etc.  Also happy to fix space_gen or accept patches.  It's entire purpose at this point is to generate these bindings, and I built it only because all other open_api generators for dart seemed terrible.

eseidel (approved):

PR #3811, opened 2026-06-04
test(shorebird_cli): cover the aab extraction failure fallback

bdero:
Follow-up to #3808, which merged with a codecov/patch failure: the catch block in `androidArchsDirectoryFromAab` was never executed by tests.

The invalid-zip test was meant to cover it, but `ZipDecoder` yields an empty archive for unrecognized input rather than throwing, so that test was exercising the no-entries fallback instead. This adds a test using a structurally valid zip whose `libapp.so` entry has a payload that cannot be decompressed, which throws during extraction and exercises the catch. Both fallback tests now also assert their distinguishing log lines, so each branch is provably hit rather than inferred from the shared fallback result.

No production code changes.

### DIFF
--- a/packages/shorebird_cli/test/src/artifact_manager_test.dart
+++ b/packages/shorebird_cli/test/src/artifact_manager_test.dart
@@ -668,6 +668,8 @@ void main() {
 
       test('falls back to the stripped dir when the aab is not a valid '
           'zip', () async {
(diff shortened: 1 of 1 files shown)


eseidel (approved):

eseidel:
cspell is failing.

Issue #3783, opened 2026-05-19
`flutter build` completes successfully and `shorebird release` fails when using the same flutter version


SalmonPhapang:
* What went wrong:
Execution failed for task ':huawei_location:compileReleaseJavaWithJavac'.
> Could not resolve all files for configuration ':huawei_location:releaseCompileClasspath'.
   > Failed to transform flutter.jar to match attributes {artifactType=android-classes-jar, org.gradle.libraryelements=jar, org.gradle.usage=java-runtime}.
      > Execution failed for IdentityTransform: C:\Users\Owen-PC\.shorebird\bin\cache\flutter\70488d38f94268163dc4177b3ec268330766b4a3\bin\cache\artifacts\engine\android-arm\flutter.jar.
         > File/directory does not exist: C:\Users\Owen-PC\.shorebird\bin\cache\flutter\70488d38f94268163dc4177b3ec268330766b4a3\bin\cache\artifacts\engine\android-arm\flutter.jar

[1779181540199_shorebird.log](https://github.com/user-attachments/files/28004591/1779181540199_shorebird.log)

Issue #3703, opened 2026-04-21
shorebird patch fails on Flutter 3.41.6 with FlutterFire 12.2.0+ (agpMajor-gated kotlin-android apply)


Joatmon13:
## Description

`shorebird patch android` fails with a cryptic "Failed to build AAB. Exited with code 1" when the project uses any FlutterFire plugin from the line that gates `kotlin-android` plugin application on AGP version — notably `firebase_analytics 12.2.0+`, but the same conditional pattern appears in most FlutterFire plugins updated after Feb 2026.

The same project builds fine via `flutter build appbundle --release` on system Flutter 3.41.5, but fails under Shorebird's cached Flutter 3.41.6 (commit `76ca3dff01`). This matters because Shorebird patches always use the cached SDK matching the release commit — so once your release was cut on 3.41.6, every patch goes through this broken path.

## Two distinct issues

**Issue 1 — AGP version detection divergence (the actual bug).** FlutterFire plugins at `firebase_analytics 12.2.0+` (and several others) contain:

```groovy
def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
if (agpMajor < 9) {
    apply plugin: 'kotlin-android'
}
```

Under stock Flutter 3.41.5 with AGP 8.9.1, this evaluates true → `kotlin-android` applies → `:firebase_analytics:compileReleaseKotlin` runs → `FlutterFirebaseAnalyticsPlugin.class` is produced.

Under Shorebird's cached Flutter 3.41.6 with the same AGP 8.9.1 declared in `settings.gradle`, the Kotlin task is never created for `:firebase_analytics`. Verbose Gradle task list confirms: `preBuild`, `preReleaseBuild`, `prepareReleaseArtProfile` run, but no Kotlin compile task. Result: `GeneratedPluginRegistrant.java:59: cannot find symbol class FlutterFirebaseAnalyticsPlugin`.

Something in Shorebird's flutter_tools/gradle injection is causing that AGP version check to resolve to 9+ (or throw and fall through), even though `settings.gradle` pins AGP at 8.9.1.

[...]

PR #3820, opened 2026-06-15
fix(shorebird_cli): point users at plugins that reference legacy flutter.jar

bdero:
Improves diagnostics for #3783

A plugin whose `android/build.gradle` still declares a file dependency on the legacy `flutter.jar` engine artifact fails with an opaque Gradle error.

```
File/directory does not exist ... \.shorebird\...\android-arm\flutter.jar
```

`flutter.jar` was the Flutter v1 Android embedding artifact, removed in Flutter 1.12. Modern Flutter ships the Android engine as Maven artifacts, so the file never exists in a current cache no matter how clean it is. The reporter's plain `flutter build` succeeds because it builds arm64 only, while `shorebird release` builds every ABI by default (including 32-bit `android-arm`), so it hits the legacy reference. This is the "flutter build works but shorebird fails" symptom in the title.

### Change

On an Android build failure, scan the project's resolved plugins (via `.flutter-plugins-dependencies`) for a `flutter.jar` reference and prepend a targeted hint to the existing fix recommendation, naming the offending plugin(s) and telling the user to upgrade or patch them. Non-Android builds are unaffected, and the scan never throws so it cannot mask the underlying failure.

### Not addressed here

Shorebird still does not capture the Gradle output itself (the build inherits stdio for interactive UX, tracked in #3703), so the hint comes from a static plugin scan rather than from matching the error text. The actual fix for an affected project is upgrading the offending plugin.

### Test plan

- New `legacy_flutter_jar_test.dart` covering detection, `.kts` files, multiple plugins, malformed manifest, and recommendation text.
- New `artifact_builder_test.dart` case asserting the hint is prepended on AAB build failure.
- `dart analyze` and `dart format` clean.


eseidel (approved):
lgtm, I wonder if this should just be in upstream flutter?

nickshorebird:
Merging as is approved, a valuable bug fix and LGTM.

PR #3821, opened 2026-06-15
feat(shorebird_code_push_protocol): unique-users current/previous envelope + shared MetricsRange [1/5]

easymac:
## What

Restructures the generated `shorebird_code_push_protocol` unique-users types onto a **current/previous metrics envelope**, and consolidates the per-endpoint range models into one shared `MetricsRange`.

Generated from the `code_push_server` OpenAPI spec via `tool/gen.dart` (space_gen); the hand-maintained `private_user.dart` field is preserved per the package's standard post-regen restore. Three commits:

1. **Restructure onto the envelope.** `GetUniqueUsersResponse` becomes `{as_of, granularity, current, previous}`, built on a window-atom pair: `UniqueUsersWindow` (the `previous` window) and an `allOf`-extended `UniqueUsersCurrentWindow` that adds `breakdown`. `UniqueUsersRange` and `PatchAdoptionRange` are retired in favor of a shared `MetricsRange`, and `month` is dropped from both `granularity` enums.
2. **Make `previous` nullable.** `GetUniqueUsersResponse.previous` is now an optional `UniqueUsersWindow?` — the consumer drops it when the equal-length prior window reaches past the plan's metrics-history horizon or before the data floor.
3. **Regenerate.** Re-runs the generator so doc comments and round-trip tests match the latest spec (horizon-gated `time_series` docs, `month` removed from patch-adoption, non-null `maybeFromJson` assertions).

## Why

This is the protocol layer of the unique-users (MAU) metrics work: serving MAU as a period-over-period envelope (current vs. previous window), tiered by each plan's metrics-history horizon. The `code_push_server` and console changes that consume these types pin this commit.

## Compatibility

**Breaking for the unshipped unique-users endpoint only.** `MetricsRange` is `{start, end}` field-compatible with the retired `UniqueUsersRange` / `PatchAdoptionRange`, so the swap is wire-safe; no consumer outside the console references the restructured `GetUniqueUsersResponse` shape.

## Testing

Hand-written round-trip tests cover the fully-populated composed window and the envelope; generated round-trip tests assert non-null `maybeFromJson` results.


nickshorebird (approved):
LGTM

PR #3825, opened 2026-06-18
feat: add uninstall command to remove shorebird from the system

AbhishekDoshi26:
## Description

Added `shorebird uninstall` command to completely remove the shorebird installation from system along with tests.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [x] 🧪 Tests


eseidel:
I'm not opposed, but I think unless we have requests from users this is just signing ourselves up for maintenance beyond our current docs.

AbhishekDoshi26:
@eseidel I agree. But when we see at the [docs](https://docs.shorebird.dev/code-push/uninstall/), it feels that the software is not fully prepared. We just tell the user that stop using shorebird commands which feels a bit weird. Open to thoughts.

eseidel:
If you believe this is blocking sales, I'm happy to land, but I doubt that.  Most software doesn't have an install path.  "You can remove shorebird from your path by removing it from your .bashrc or .zshrc and deleting the .shorebird directory located in ~/.shorebird." seems like sufficient for someone to provide to their agent, and there is risk in our command accidentally deleting the wrong folder, etc.

AbhishekDoshi26:
> If you believe this is blocking sales, I'm happy to land, but I doubt that. Most software doesn't have an install path. "You can remove shorebird from your path by removing it from your .bashrc or .zshrc and deleting the .shorebird directory located in ~/.shorebird." seems like sufficient for someone to provide to their agent, and there is risk in our command accidentally deleting the wrong folder, etc.

That also makes sense. Will close the PR then. Thanks for the feedback 🎉

PR #3826, opened 2026-06-19
fix(shorebird_cli): report when a patch is missing libapp.so

### LINKED ISSUE #3813: fix: Android build failing: No architecture artifacts found to upload.
The Fix from the Pull Request does not seem to work for me:
https://github.com/shorebirdtech/shorebird/pull/3808

libapp.so.sym or libapp.so.dbg not present when checking final appbundle for debug symbols. Play Console will not receive Dart crash symbols for this build. To enable, remove
`packaging.jniLibs.keepDebugSymbols.add("**/libapp.so")` from android/app/build.gradle.kts.
✗ No architecture artifacts found to upload.

[1781024913966_shorebird.log](https://github.com/user-attachments/files/28764718/1781024913966_shorebird.log)

```
Shorebird 1.6.107 • git@github.com:shorebirdtech/shorebird.git
Flutter 3.44.1 • revision 26ad1cdf21201b704d7214cab0379ff8779c15ea
Engine • revision caf48e58a872271fc47e43f7dae6192bb78d60d1

URL Reachability
✓ https://api.shorebird.dev OK (0.2s)
✓ https://console.shorebird.dev OK (0.4s)
✓ https://oauth2.googleapis.com OK (76ms)
✓ https://storage.googleapis.com OK (92ms)
✓ https://cdn.shorebird.cloud OK (79ms)

✓ Shorebird is up-to-date (0.8s)
✓ AndroidManifest.xml files contain INTERNET permission (26ms)
✓ android/app/build.gradle(.kts) does not contain a legacy keepDebugSymbols line for libapp.so (0.4s)
✓ Xcode project does not override FLUTTER_ build settings (20ms)
✗ macOS app has correct entitlements (2ms)
  [✗] /Users/***/Developer/***/app/macos/Runner/Release.entitlements is missing the Allow Unsigned Executable Memory (com.apple.security.cs.allow-unsigned-executable-memory) entitlement.
✓ shorebird.yaml found in pubspec.yaml assets (6ms)
✓ Lock files are tracked in source control (46ms)
1 issue detected.
1 issue can be fixed automatically with shorebird doctor --fix.
```

I also don't have any trace of packaging.jniLibs.keepDebugSymbols.add("**/libapp.so")`  or similar in my gradle files.

```
shorebird release android --flavor=envtest
shorebird release android --flavor=envtest --target-platform=android-arm,android-arm64,android-x64
```
make no difference

bdero:
Fixes #3813

## Summary

A missing `libapp.so` goes wrong two ways on the Android path, and both point users somewhere other than the actual problem. With none in the bundle, the error blames `ndk.abiFilters` / `splits.abi` / `jniLibs.excludes`, which cost the reporter several wrong turns before they traced it to Gradle task ordering in their own build. With one for only some architectures, `shorebird patch` exits 0 while the omitted architectures go unpatched, visible only at `--verbose`.

- Release and patch both check the built `.aab` up front and fail w/ a message naming the real cause and listing what is present under `base/lib`. An `.aab` w/o `libapp.so` would not run on device, so there is nothing to fall back to
- A partial patch now warns, naming the patched and skipped ABIs, stating that the patch build and the release build disagree on which architectures to compile, and giving the rebuild as the action
- Dropped `ndk.abiFilters` and `splits.abi` from that guidance. On AGP 8.11.1 a clean build w/ `abiFilters` limited to one ABI still packages every ABI's `libapp.so`, and `splits.abi` is ignored when building a bundle, so naming either repeats the misdiagnosis this change exists to remove
- The per-arch skip stays. #3388 showed that removing it reintroduces `PathNotFoundException` for projects that legitimately exclude an ABI

## Testing

- Unit tests for the detection: missing libapp, libapp present, garbage file, empty `base/lib`. Release and patch flow tests assert the message reaches the user, and patcher tests assert the warning names both ABI sets and stays quiet at full coverage
- Ran the real flow on a throwaway app: `release android`, then `patch android` at full coverage with no warning, then `patch android` w/ a `packaging.jniLibs` excludes rule dropping two ABIs, which warned `Patched: arm64-v8a` / `Skipped: armeabi-v7a, x86_64` and published a patch covering arm64 alone
- Checked the detection against a real Gradle-built bundle rather than a fixture: healthy and partially-filtered bundles pass through, an all-`libapp.so`-stripped copy produces the diagnostic
- Confirmed the dropped guidance was wrong before dropping it: a clean AGP 8.11.1 build w/ `ndk.abiFilters` limited to one ABI still packaged all three `libapp.so`


nickshorebird:
Picking this one up as is stale, but a valid change. Assigned to self.

easymac (approved):

PR #3832, opened 2026-06-26
feat: resumable artifact uploads

eseidel:
Adds a resumable upload path for release/patch artifacts. Pairs with a server-side change (separate repo) that hands the client a resumable session and gates it on CLI version.

## Why

Clients upload artifacts directly to GCS via a URL the CodePush server returns. Today that's a single `multipart/form-data` POST — if it fails partway (common for large artifacts on flaky CI/dev networks), the whole upload restarts from zero. This adds **resumable uploads** that survive network drops and resume from the last acknowledged byte.

## What

- **Protocol** (`shorebird_code_push_protocol`): new optional `ArtifactUploadMethod` (`multipart` | `resumable`) on `CreateReleaseArtifactResponse` / `CreatePatchArtifactResponse`. Null on responses from older servers ⇒ multipart. Fully backward compatible.
- **Client** (`shorebird_code_push_client`): when the server returns `resumable`, PUT the file to the returned GCS resumable **session URI** in 8 MiB chunks with `Content-Range`, treating `308` as continue and resuming from the last acknowledged byte (querying the offset) on a transient failure. Falls back to the existing multipart POST when the field is absent/`multipart`.

No version bump here — the driver ships in the normal release cadence.

## Rollout

Safe to release independently and early: the client code is **dormant** until the server opts a client in (returns `upload_method: resumable`), which the server only does for CLIs at/above the cutoff version. So this can ship in any release with no behavior change in the field until the server activates it. Older clients always get `multipart`.

## Tests

Protocol (de)serialization incl. the new field + enum test; client tests covering the resumable PUT, `Content-Range`, 308 continue, network-failure resume, and error paths for both release and patch; existing multipart tests unchanged.

### DIFF
--- a/packages/shorebird_code_push_client/lib/src/code_push_client.dart
+++ b/packages/shorebird_code_push_client/lib/src/code_push_client.dart
@@ -160,17 +160,12 @@ class CodePushClient {
       json.decode(body) as Map<String, dynamic>,
     );
 
-    final uploadRequest = http.MultipartRequest('POST', Uri.parse(decoded.url))
-      ..files.add(file);
-
-    final uploadResponse = await _httpClient.send(uploadRequest);
-
-    if (!uploadResponse.isSuccess) {
-      throw CodePushException(
-        message:
-            '''Failed to upload artifact (${uploadResponse.reasonPhrase} '${uploadResponse.statusCode})''',
-      );
-    }
+    await _uploadArtifact(
+      artifactPath: artifactPath,
+      multipartFile: file,
+      url: decoded.url,
+      uploadMethod: decoded.uploadMethod,
+    );
   }
 
   /// Create a new artifact for a specific [releaseId].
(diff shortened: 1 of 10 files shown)


packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -219,17 +217,126 @@ class CodePushClient {
...
+          continue;
+        }
+
+        final status = response.statusCode;
+        await response.stream.drain<void>();
+
+        if (status == HttpStatus.ok || status == HttpStatus.created) return;
+        if (status == _resumeIncompleteStatus) {
+          failures = 0;
+          offset = _parseRangeEnd(response.headers['range']) ?? end;
```
nickshorebird: No range header on `308` means upload should be started from the beginning per the [checking status of resumeable uploads docs](https://docs.cloud.google.com/storage/docs/performing-resumable-uploads#status-check). 

Below suggestion to match L319 + handle infinite loop due to hard failure reset. Still could mean, in theory, number_of_retries == ceil(n_upload_bytes / 256_KiB) and/or total_time == link_timeout, but that is a larger change then below.
```dart
final next = _parseRangeEnd(response.headers['range']) ?? 0;
if (next > offset) {
  failures = 0;
} else if (++failures > maxConsecutiveFailures) {
  throw _uploadFailed(response);
}
offset = next;
```

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -219,17 +217,126 @@ class CodePushClient {
...
+        } else {
+          throw _uploadFailed(response);
```
nickshorebird, line 335: Only logic handled is on 308, not 5xx which could just be transitory and docs state it should be resumed from where it left off. Consider adding an `} else if (status >= 500) {` check with an exponential back off to handle transient server errors.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
```diff
@@ -219,17 +217,126 @@ class CodePushClient {
...
+          response = await _httpClient.send(
+            http.Request('PUT', sessionUri)
+              ..bodyBytes = chunk
+              ..headers['content-range'] = 'bytes $offset-${end - 1}/$total',
+          );
+        } on Exception {
+          // Network failure mid-chunk: ask GCS how far it got and resume.
+          if (++failures > maxConsecutiveFailures) rethrow;
+          offset = await _queryResumeOffset(sessionUri, total);
+          continue;
```
nickshorebird: Nit - Tight loop w/ a network call. Consider adding a backoff wait between calls.

nickshorebird (commented):
Comments are only concern. Otherwise LGTM.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
eseidel: Done in 4630da66. A 308 now reads the stored byte count from the `range` header and falls back to **0** (restart) when it's absent, per the status-check docs. To avoid an infinite loop on a stuck session, a 308 that doesn't advance the offset is now counted as a failure and aborts after `_maxUploadFailures` (5) consecutive non-progress responses.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
eseidel: Done in 4630da66. Added an `else if (status >= HttpStatus.internalServerError)` branch that treats 5xx as transient: it backs off and retries the same chunk (re-PUTting the same `Content-Range` is idempotent in GCS) rather than failing the upload, giving up only after 5 consecutive failures.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
eseidel: Done in 4630da66. Added exponential backoff (`_backoff`) before every retry — network failure, 5xx, and non-progress 308 — doubling a configurable base delay (`uploadRetryBaseDelay`, default 1s) per consecutive failure. Tests inject `Duration.zero` so retry paths stay instant.

packages/shorebird_code_push_client/lib/src/code_push_client.dart
eseidel: Refined in b2aee1c6 after a self-review: rather than re-PUTting the chunk, the 5xx path now queries the resume offset (`bytes */SIZE`) and resumes from where GCS left off — the same recovery the network-failure path uses, and the status-check flow the docs recommend. This avoids re-sending already-persisted bytes and removes the asymmetry between the two transient-failure paths.

nickshorebird (approved):
LGTM - Thanks for the mods.

PR #3839, opened 2026-07-06
feat(stripe_api): surface HTTP status and error code on failures

nickshorebird:
## Summary

A bare `Exception` on every failed Stripe call collapses a missing resource, a rate limit, and a server error into one indistinguishable throw, so a caller that wants to react differently to each has nothing to branch on but the message prose. I'm adding `StripeApiException`, which carries the HTTP `statusCode` and, when the body is a Stripe error object, the machine-readable `code` such as `resource_missing` or `rate_limit`. A caller can now tell a permanently-gone resource from a transient throttle w/o string-matching.

Three calls worth surfacing:

- **New subtype, not a richer message.** I threw a new type rather than enriching the string, because matching message text for control flow is exactly the brittleness this removes. It implements `Exception`, so every existing `on Exception` and catch-all handler keeps working unchanged. Throwing an unrelated type would have broken those catches.
- **Status is the floor, code is corroboration.** `code` is parsed from the JSON body when present, but a non-JSON body like an HTML gateway error never throws during construction and just leaves `code` null. The HTTP status is always set, so it stays the reliable signal rather than leaning on `code`, which Stripe leaves null on some errors.
- **One contract across all failure paths.** I applied the typed throw to every non-200 path in the client, not only the two lookups, so the error surface is uniform. The two write paths that already embedded the response body in their message now also carry a parsed code, harmless overlap.

## Testing

- New cases on `fetchCustomer` / `fetchSubscription`: a 404 body yields `statusCode` plus `code == resource_missing`, a 429 body yields `code == rate_limit`.
- `fromResponse` unit tests: a non-JSON body and a body w/o an error object both leave `code` null w/o throwing.
- Existing non-200 tests pass unchanged, confirming `implements Exception` keeps current catch sites working.

### DIFF
--- a/packages/stripe_api/lib/src/stripe_api.dart
+++ b/packages/stripe_api/lib/src/stripe_api.dart
@@ -36,7 +36,10 @@ class StripeApi {
 
     final response = await _client.get(uri, headers: _authHeaders);
     if (response.statusCode != HttpStatus.ok) {
-      throw Exception('Failed to retrieve customer with id $customerId');
+      throw StripeApiException.fromResponse(
+        response,
+        message: 'Failed to retrieve customer with id $customerId',
+      );
     }
 
     return StripeCustomer.fromJson(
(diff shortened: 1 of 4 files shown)


easymac (approved):
👍 LGTM

PR #3850, opened 2026-07-11
feat(shorebird_cli): record warm/cold + output artifact size in build trace

eseidel:
## What

Adds two privacy-safe fields to the build-trace summary (schema **v8 → v9**), as groundwork for the Flutter native build cache investigation:

- **`nativeOutputsPresentAtStart`** (bool) — snapshots whether Flutter's `build/` dir was non-empty *before* the build ran (a warm-build proxy). Captured in `prepareBuildTrace` because pre-build state isn't observable afterward. Lets field data separate cold builds (fresh CI checkout) from locally-warm ones so they don't skew the cold-build baselines a native build cache is measured against.
- **`outputArtifactBytes`** (int) — best-effort size of the primary build output (`.aab`/`.ipa`). Sizes what a remote native-shell cache would have to move over the network.

Both are computed entirely CLI-side; no trace-producer/fork changes. New fields are omitted (not nulled) when unavailable, so old consumers are unaffected.

## Notes / open question

- `outputArtifactBytes` measures the **full built app artifact**, located via a best-effort glob. On **release** this equals the uploaded artifact; on **patch** it does *not* (patch uploads a per-arch diff, this measures the full `.aab`/`.ipa`) — intentional, since the cache-sizing question is about the full native artifact, not the code-push diff. Open to switching to the exact known artifact path (`findAab` etc.) if we prefer exactness over the glob.

## Not included (follow-ups)

- Surfacing these fields in the admin build-timing dashboard (separate web repo; deferred until data accrues).
- Finer sub-measurements (signing span, iOS `xcode_subsection` split) — spans the flutter fork and touches privacy-sensitive subsection titles; separate PR.

## Test

`dart analyze` clean (no new issues), `dart format` clean, 121 tests pass (added serialization coverage for the new fields; bumped two `version` assertions to 9).

### DIFF
--- a/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
+++ b/packages/shorebird_cli/lib/src/artifact_builder/artifact_builder.dart
@@ -650,6 +650,50 @@ Reason: Exited with code $exitCode.''',
     );
     traceFile.parent.createSync(recursive: true);
     buildTraceSession.traceFile = traceFile;
+
+    // Snapshot warm-vs-cold *before* the build mutates `build/`. A fresh
+    // checkout (the common CI case) has no `build/` yet; a repeat local
+    // build does. `writeBuildTraceSummary` reads this back later.
+    buildTraceSession.nativeOutputsPresentAtStart = _nativeOutputsPresent();
+  }
+
+  /// Whether Flutter's `build/` output directory already exists and is
+  /// non-empty, a warm-build proxy. Null when the project root can't be
+  /// resolved. Called from [prepareBuildTrace] before the build runs.
+  bool? _nativeOutputsPresent() {
+    final root = shorebirdEnv.getShorebirdProjectRoot();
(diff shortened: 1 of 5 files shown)


eseidel:
Closing without landing. Both fields turned out unnecessary:

- **cold/warm** is already inferable from the v8 trace we collect today (Gradle `taskUpToDateCount`/`taskFromCacheCount` vs `taskExecutedCount` on Android; `flutterAssemble.skippedCount` on both platforms), and that measures *realized* reuse rather than the mere *potential* a `build/`-present flag captures. It also works retroactively on data already collected.
- **outputArtifactBytes** is premature (we haven't defined the cache unit, and the final artifact is not what we'd cache) and redundant (final-artifact size is already known server-side from uploads).

Better next step is to bucket the existing corpus by the inferred cold/warm signal instead of adding fields.

PR #3855, opened 2026-07-19
fix: pass --no-clear when regenerating against space_gen 1.4.0

eseidel:
## Description

`tool/gen.dart` points space_gen at this package, which holds hand-written code
beside generated output: `lib/extensions/`, the `shorebird_code_push_protocol.dart`
barrel, `test/src/`, the pubspec and the README.

From **space_gen 1.4.0** the generator empties the `-o` directory before each run.
Here that would delete every hand-written file in the package. 1.4.0 adds a
`--no-clear` flag to opt out.

This PR:

- **`tool/gen.dart`** appends `--no-clear` to the argument list, unconditionally
  (a forgotten flag would be destructive), so regeneration writes generated files
  and removes nothing. Removing stale generated output — a renamed schema's old
  file — becomes this package's own responsibility.
- **`pubspec.yaml`** requires `space_gen: ^1.4.0`. `--no-clear` exists only from
  1.4.0; an older release would throw on the unknown flag.

This replaces the original holding measure on this branch (cap `space_gen` below
1.4.0), now that 1.4.0 is published with the opt-out.

## Why not clean up stale output automatically

space_gen 1.4.0 can, via a `generatedDirs` override declaring the directories
this package generates into (`lib/src/models`, `lib/src/messages`,
`test/generated`). `--no-clear` is the simpler contract — this package manages
its own stale files — and is what was chosen. A `generatedDirs` override can come
later if the manual cleanup becomes a burden.

## Testing

- `dart pub get` resolves `space_gen 1.4.0` across the workspace.
- `dart analyze tool/gen.dart` clean.
- Verified through the resolved 1.4.0 that `--no-clear` is a real flag wired to
  `clearDirectory: false`, and that `clearDirectory: false` skips clearing
  (covered by space_gen's own test suite). A full regeneration was not run here —
  it hits the live API and is a separate change with its own large diff.

Upstream: https://github.com/eseidel/space_gen/pull/310, shipped in
[space_gen 1.4.0](https://pub.dev/packages/space_gen/versions/1.4.0).


nickshorebird (approved):
LGTM

Issue #3854, opened 2026-07-19
--dart-define-from-file values are empty at runtime on shorebird release ios, even with correct -- placement


MennatullahMohamed:
String.fromEnvironment() calls always return their defaultValue (never the value from --dart-define-from-file) in apps built via shorebird release ios. The same file works correctly with plain flutter build ios and (per other reports) with shorebird release android. This looks related to — but not fully resolved by — the flag-position fix in v1.1.9, and to flutter/flutter#142976 (--dart-define-from-file does not work with flutter build ipa), since shorebird release ios uses that same flutter build ipa path internally.

Environment: Shorebird CLI via shorebirdtech/setup-shorebird@v1 on GH Actions macOS runner; Flutter 3.35.4 (ad2bbfaadb), pinned via --flutter-version=3.35.4; iOS, macos-26 runner, Xcode 26.3.

Steps to reproduce:

secrets.json with several string key/value pairs (verified valid JSON, non-empty).
shorebird release ios --flavor runner --target lib/main.dart --export-options-plist ./ios/ExportOptions.plist -- --dart-define-from-file=secrets.json
Build succeeds, uploads fine, passes App Store Connect verification, installs via TestFlight.
String.fromEnvironment('SOME_KEY', defaultValue: '') returns '' for every key, instead of the real value.
What we already ruled out: flag position (tried before Shorebird's flags and after --, same result either way — the separator is respected, values still don't land); stale GH Actions DerivedData cache (cleared, no change); stale TestFlight build (confirmed via App Store Connect we tested the exact new build number); pre-generating ios/Flutter/Generated.xcconfig via flutter build ios --config-only --dart-define-from-file=secrets.json chained immediately before the release command (no effect — CI log shows shorebird release ios runs its own fully independent flutter pub get/config regeneration under its own isolated Flutter install at ~/.shorebird, discarding anything the prior step did); secrets.json content itself (validated programmatically before the build — correct JSON, all keys, no empty values).

Confirmed the values are genuinely never applied: added a temporary on-screen debug widget printing several String.fromEnvironment(...) values directly at launch. On a TestFlight-installed build from the command above, every value showed empty — not a logging/visibility issue, the values truly aren't compiled in.

Expected: shorebird release ios ... -- --dart-define-from-file=secrets.json should thread the file's values into the compiled Dart AOT snapshot the same way flutter build ios --dart-define-from-file=secrets.json does.

[...]

PR #3856, opened 2026-07-20
fix: dart-define-from-file values empty on release iOS

assassinaj602:
This PR fixes a path resolution issue that caused --dart-define-from-file values to be empty during iOS releases. The fix ensures the file path is absolute and read at the correct stage of the build process.


nickshorebird:
@assassinaj602 - Thanks for putting this together and sorry for the two month delayed response to this.

This looks aimed at #3854 , but I do not think it is the fix. Flutter reads `--dart-define-from-file` in process and hard exits when the file is missing. The build in issue 3854 succeed thus making the path absolute does not change anything.

Closing for the above. Feel free to comment on 3854 if this was aimed at that w/ a reproducer as it needs one. Also feel free to tag me if you do pick that up. Will review in less than two months this time.

PR #3858, opened 2026-07-20
fix: improve error handling for release command

assassinaj602:
This PR wraps the release process in a try-catch block to prevent uncaught exceptions from bubbling up as stack traces and instead logs a clear error message.


nickshorebird:
@assassinaj602 - Thanks for the PR. Going to close this as different approach was accepted in #3923 . As an aside the bare `catch` also swallows `UsageException` moving the exit code from 64 -> 70. [Nice blog ](https://ericlippert.com/2008/09/10/vexing-exceptions/)on exception design from 2008 if you are interested in reading some more on exception handling in system design.

PR #3866, opened 2026-07-28
chore(scoped_deps): document pub.dev handoff and stop publishing the vendored copy

eseidel:
Documents where `packages/scoped_deps` stands now that the pub.dev package has a new owner, and stops pretending we can still publish it.

Felix wrote this package in 2023 (originally `package:scoped`) so our packages could share one dependency-injection mechanism, and we published it to pub.dev in 2024 in case others found it useful. It hasn't needed changes since.

Morgan Hunt ([mrgnhnt96/scoped_deps](https://github.com/mrgnhnt96/scoped_deps)) asked to take over the pub.dev package and keep improving it, and we accepted. [pub.dev's `scoped_deps`](https://pub.dev/packages/scoped_deps) is now his, and has already diverged from this directory — it's at 0.2.1, we're at 0.1.0+2.

Our packages keep using this vendored copy, resolved through the Dart workspace rather than from pub.dev. The code hasn't meaningfully changed in two years, so there's nothing to gain from moving right now. The README notes that we should revisit deleting this copy in favor of the pub.dev package later.

## Changes

- `README.md`: new section explaining the history, the handoff, and why we stay vendored for now.
- `pubspec.yaml`: `publish_to: none`, and drop `homepage` now that this is not a pub.dev listing. Kept `repository`, which still correctly points at the vendored source.
- Delete `.github/workflows/publish_scoped_deps.yaml`, which would fail if a `scoped_deps-v*` tag were ever pushed. The shared `publish_dart_package.yaml` it called stays — `shorebird_ci` and `redis_client` still use it.

No code changes to the package; `dart pub get` still resolves cleanly and the lockfile is unchanged. The package's regular CI job in `shorebird_ci.yaml` is untouched, since we still build and test the vendored copy.

## Drive-by: unbreak CSpell on main

Also adds `lipo` to the inline cspell word list in `RELEASE_NOTES.md`. Unrelated to scoped_deps — the 1.6.115 notes added in 023c9ad3 mention Xcode 27's `lipo` architecture verification error, which turned CSpell red on main and on every PR branched from it.


nickshorebird (approved):
LGTM - Thanks for this.

PR #3872, opened 2026-08-03
fix(shorebird_cli): measure download time in build traces

eseidel:
## Problem

Our build-time telemetry could not see download time. A build that spent
ten minutes pulling artifacts over a slow link and a build that spent ten
minutes doing local work produced nearly the same summary. Three separate
gaps caused this:

1. **`network.ms` measured time-to-first-byte, not transfer time.**
   `TracingClient` timed `send()`, which resolves the moment response
   headers arrive. Bodies are drained later by the caller (see
   `CachedArtifact.extractArtifact`), so a 147MB `dart-sdk` fetch was
   recorded as a few hundred milliseconds.

2. **Cold-start setup was entirely untimed.** Cloning the Flutter fork,
   `flutter precache`, and refreshing Shorebird's cached artifacts are the
   three biggest downloads a cold machine performs. None had a span. All
   three run in subprocesses, so `TracingClient` cannot see them either,
   and `flutter precache` pulls engine artifacts *before* flutter_tools
   installs its own `BuildTracer`.

3. **That time was then mis-attributed.** `shorebirdOverhead` is computed
   by subtraction (total minus Flutter's reported build), so everything
   unmeasured landed in `shorebirdLocalMs` and read as local work — the
   exact opposite of the truth on a cold CI runner.

Two doc comments claimed `network` already covered "Flutter-side HTTP
(artifact downloads when the cache is cold)". It never did. That is
probably why this looked instrumented.

## Changes

**`TracingClient` now spans the whole transfer.** The span closes on the
last drained byte; `ttfbMs` and `responseBytes` are recorded alongside so
a slow origin stays distinguishable from a slow pipe.

The body is mirrored through a `StreamController` rather than an `async*`
generator. A generator only observes cancellation at its next `yield`, so
aborting a download that stalled mid-transfer would hang until the next
chunk arrived — a test in this PR covers that. Forwarding `onPause` /
`onResume` also keeps backpressure reaching the socket instead of
buffering an artifact in memory while the consumer is paused.

Non-2xx responses still close at header time: callers routinely throw on
an error response without reading it, and waiting for a drain that never
comes would silently drop the event.

**New `setup` trace category** (`SetupPhase`: `flutter_install`,
`flutter_precache`, `shorebird_cache`) surfaced as a `setup` section in
the summary. Every field is a subset of `shorebirdOverheadMs`; on a warm
machine all three are ~0.

**Summary version 8 → 9.** Additive only; existing fields are unchanged,
so nothing downstream has to move in lockstep.

## Notes for review

- `shorebirdLocalMs` keeps its current definition (`overhead - network`)
  so the existing dashboard metric does not shift underneath us. Its doc
  comment now states plainly what it does and does not exclude, and the
  new `setup.totalMs` is what you subtract for a truer local number. The
  shorebird-cache downloads appear in both, so that subtraction
  double-counts them; called out in the doc rather than papered over.
- Engine artifacts pulled lazily *during* `flutter build` (when
  `installRevision` early-returns and precache is skipped) are still
  invisible. Fixing that means installing the `BuildTracer` before
  `Cache.updateAll` in the fork's `flutter_command.dart` — a separate
  change in a separate repo.
- Test-only scope additions: `cache_test` and `shorebird_flutter_test`
  now seed `shorebirdTracerRef`, which production always provides via
  `main()`.

## Testing

- Full `shorebird_cli` suite passes (2074 tests).
- 100% diff coverage on changed lines.
- New tests cover drain-vs-header timing, cancellation, pause/resume
  backpressure, non-2xx header-time recording, response metadata
  pass-through, and setup-phase bucketing/serialization.


easymac (approved):
Love this

PR #3877, opened 2026-08-08
feat(shorebird_cli): use PKCE in the loopback login flow

eseidel:
## What

The CLI now uses PKCE ([RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636)) in the loopback login flow. Each login generates an S256 verifier/challenge pair, sends the challenge on the authorization request, and sends the verifier when exchanging the authorization code.

## Why

`shorebird_cli` is a public OAuth client. It holds no client secret, and it receives its authorization code on a `http://localhost:<port>` redirect rather than a server-to-server callback. PKCE ties the code to a secret that never leaves the CLI process until the exchange, so possessing the code alone is not enough to redeem it.

This is standard practice for native and CLI clients, and it is a prerequisite for the OAuth flows we want to support next.

## Compatibility

No coordination needed with a server rollout, in either direction:

- The auth service already accepts `code_challenge` / `code_challenge_method` on the authorization request.
- It requires a `code_verifier` **only** when a challenge was bound to the code being redeemed, and that binding happens server-side rather than being read off the token request.

So a CLI that sends a challenge gets the protection, and a CLI that does not keeps working exactly as before. Nothing in this PR makes PKCE mandatory.

## Changes

- `lib/src/auth/pkce.dart` — new. Generates the pair from `Random.secure()`, 32 bytes of entropy, unpadded base64url as the RFC requires.
- `lib/src/auth/shorebird_oauth.dart` — sends `code_challenge` + `code_challenge_method=S256` on the login URL, and `code_verifier` in the token exchange.

## Tests

- `test/src/auth/pkce_test.dart` — challenge is `base64url(sha256(verifier))`, verifier is within the RFC's 43-128 character bounds, both values stay inside the unreserved character set, and each call produces a distinct verifier.
- `test/src/auth/shorebird_oauth_test.dart` — end-to-end through the loopback flow, asserting the challenge on the login URL and that the verifier sent at the exchange is the one that challenge was derived from. Asserting the relationship rather than mere presence: if the two ever drift apart, every login fails at the token endpoint.

Full package suite passes (2072 tests). `dart analyze --fatal-warnings .` clean, `dart format` clean, and every line added here is covered.

### DIFF
--- a/cspell.config.yaml
+++ b/cspell.config.yaml
@@ -98,6 +98,7 @@ words:
   - pbxproj
   - pinact # Third-party action-pinning tool referenced in shorebird_ci
   - Perfetto # Chrome trace viewer at ui.perfetto.dev
+  - pkce # Proof Key for Code Exchange (RFC 7636)
   - pkcs
   - podfile
   - podspec
--- a/packages/shorebird_cli/lib/src/auth/pkce.dart
+++ b/packages/shorebird_cli/lib/src/auth/pkce.dart
@@ -1,0 +1,44 @@
+import 'dart:convert';
+import 'dart:math';
+
+import 'package:crypto/crypto.dart';
+import 'package:meta/meta.dart';
+
+/// A PKCE (RFC 7636) verifier and the S256 challenge derived from it.
+///
+/// The CLI is a public OAuth client: it holds no secret, and it receives its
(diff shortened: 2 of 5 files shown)


eseidel:
These should be safe to land in either order.  Both the server and the client are OK with the other not yet supporting PKCE for now.

PR #3894, opened 2026-08-25
feat(stripe_api): subscription list, create, and cancel

easymac:
Adds the subscription operations the code push server needs to bill annual overages on a second, monthly metered subscription: `listSubscriptions` (paginated `GET /v1/subscriptions`), plus the package's first write support — `createSubscription` (price, metadata, payment method, tax, currency, collection method with `days_until_due`, `Idempotency-Key` header), `cancelSubscription` with an immediate final invoice (`invoice_now`/`prorate`) that returns the canceled subscription as Stripe reports it, and `fetchPaymentMethod`, exposing a payment method's card country.

`StripeSubscription` gains read-only `metadata`, `currency`, `default_payment_method`, `automatic_tax`, and `collection_method` fields. The server copies these from a customer's existing subscription onto the new one, so it has to be able to read them. A null currency or payment method is left out of the create request so Stripe falls back to the customer's own defaults. `collection_method` is a `StripeCollectionMethod` enum on the model and on `createSubscription`, so a misspelt method fails to compile instead of failing at Stripe.

`StripePrice` gains `interval` and `intervalCount` from `recurring`. The console currently guesses a plan's billing cadence from the length of its current period; reading it off the price the customer bought removes the guess.

Also fixes a small pre-existing bug in `_fetchAllPages`: it added its paging parameters to the caller's own query map instead of a copy. No caller was actually affected; fixed while in the file.

The four existing `fetchActiveSubscriptions` tests stand unmodified. New tests cover the list query string and pagination, the create parameters and idempotency header (`collection_method`/`days_until_due` sent only when asked), the cancel request and the subscription it returns, the payment-method fetch, the new subscription fields, and the price interval fields including a one-time price with no `recurring` block.

### DIFF
--- a/packages/stripe_api/lib/src/models/stripe_subscription.dart
+++ b/packages/stripe_api/lib/src/models/stripe_subscription.dart
@@ -62,6 +62,7 @@ class StripeSubscription {
     this.currency,
     this.defaultPaymentMethod,
     this.automaticTaxEnabled = false,
+    this.collectionMethod,
   });
 
   /// Converts a `Map<String, dynamic>` to a [StripeSubscription].
@@ -145,6 +146,10 @@ class StripeSubscription {
   /// invoices.
   @JsonKey(name: 'automatic_tax', fromJson: _automaticTaxEnabledFromJson)
   final bool automaticTaxEnabled;
+
+  /// How this subscription's invoices collect payment:
+  /// `charge_automatically` or `send_invoice`.
+  final String? collectionMethod;
 
   /// Whether this subscription is in an active or trialing state.
   bool get isActiveOrTrial =>
--- a/packages/stripe_api/lib/src/models/stripe_subscription.dart
+++ b/packages/stripe_api/lib/src/models/stripe_subscription.dart
@@ -36,6 +36,23 @@ enum StripeSubscriptionStatus {
   paused,
 }
 
+/// How a [StripeSubscription]'s invoices collect payment.
+///
+/// See https://docs.stripe.com/api/subscriptions/object#subscription_object-collection_method.
+@JsonEnum(fieldRename: FieldRename.snake)
+enum StripeCollectionMethod {
+  /// Stripe charges the payment method on file when an invoice finalizes.
+  chargeAutomatically('charge_automatically'),
+
+  /// Stripe emails a hosted invoice and waits for the customer to pay it.
+  sendInvoice('send_invoice');
+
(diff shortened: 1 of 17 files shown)


packages/stripe_api/lib/src/models/stripe_subscription.dart
```diff
@@ -124,6 +129,28 @@ class StripeSubscription {
...
+  final String? defaultPaymentMethod;
+
+  /// Whether Stripe calculates tax automatically on this subscription's
+  /// invoices.
+  @JsonKey(name: 'automatic_tax', fromJson: _automaticTaxEnabledFromJson)
+  final bool automaticTaxEnabled;
+
+  /// How this subscription's invoices collect payment:
+  /// `charge_automatically` or `send_invoice`.
+  final String? collectionMethod;
```
nickshorebird: Suggest using `ENUM` here rather than `String?`. Typo for strings is a runtime bug while enum typo is a compile time bug.

packages/stripe_api/lib/src/stripe_api.dart
```diff
@@ -69,6 +69,126 @@ class StripeApi {
...
+    );
+  }
+
+  /// Immediately cancels the subscription with the given [subscriptionId].
+  ///
+  /// [invoiceNow] and [prorate] control whether pending metered usage is
+  /// invoiced on cancellation; Stripe does neither by default.
+  ///
+  /// See https://docs.stripe.com/api/subscriptions/cancel.
+  Future<void> cancelSubscription({
```
nickshorebird: Consider returning the response body rather than void. The returned stripe object is being dropped here as the return type is `void`. This means passed args are no-ops when users could have actionable flows based on cancellation.

nickshorebird (approved):
LGTM otherwise - Leave the two comments to your discretion.

packages/stripe_api/lib/src/models/stripe_subscription.dart
easymac: Done in b94e2279: `StripeCollectionMethod` enum on the model and on `createSubscription`.

packages/stripe_api/lib/src/stripe_api.dart
easymac: Done in b94e2279: `cancelSubscription` now returns the canceled `StripeSubscription` Stripe sends back.

PR #3901, opened 2026-08-28
fix(shorebird_cli): precache and resume the requested Flutter install

nickshorebird:
## Summary

`installRevision` precached with the active revision's binary rather than the
one it had just cloned, so the new install got no artifacts and the guard from
#3789 only ever covered a revision that was not at risk. It also returned early
on any existing directory, so an interrupted run left a state every later run
read as finished, recoverable only by `shorebird cache clean`, which wipes every
installed revision to repair one.

- Precache runs in a `runScoped` override pinned to the target revision.
  `useVendedFlutter` stays on, so the storage base URL, `--verbose`, and local
  engine arguments are unchanged.
- Clone and checkout stage in a private directory published by rename, w/ an
  untracked `.shorebird_precache` stamp, so an interrupted run resumes rather
  than re-cloning.
- A launcher-less directory is moved aside, not deleted. The shell bootstrap's
  in-place `git clone --no-checkout` looks identical while it runs, and no lock
  this side can take separates them.
- Reclaimable directories carry their creation time in the name, so the sweep
  never reads an mtime that an unrelated write can bump and a rename leaves
  untouched.
- `CacheCorruptedException` takes an optional remedy, so precache failures say
  to retry and only callers w/ nothing narrower still say `cache clean`.

## Testing

- Full `shorebird_cli` suite passes, 2093 tests. Diff coverage 100%, which
  `.github/codecov.yml` requires to merge.
- Every new guard mutation-checked. 10 mutations, all killed.
- Repaired two `verifyNever` calls whose named arguments could never match the
  real call, leaving the already-installed fast path vacuously covered.

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_env.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_env.dart
@@ -13,21 +13,29 @@ import 'package:shorebird_cli/src/shorebird_cli_command_runner.dart';
 import 'package:shorebird_code_push_client/shorebird_code_push_client.dart';
 
 /// Exception thrown when the Shorebird cache appears to be corrupted.
-///
-/// Surfaces a user-actionable message directing the user to run
-/// `shorebird cache clean` and retry.
 class CacheCorruptedException implements Exception {
   /// Creates a [CacheCorruptedException] explaining why the cache is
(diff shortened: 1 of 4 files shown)


easymac (approved):
👍 Way cool

PR #3905, opened 2026-08-31
feat(shorebird_cli): add app and channel management commands

nickshorebird:
## Summary

App and channel lifecycle has been console-only, so an agent or CI job that
needs to rename an app or retire a channel drops out of the CLI and calls the
API by hand. The endpoints already exist, only the CLI surface was missing. I
gated both deletes on `--confirm-name` rather than a prompt or a `--force`
flag, since a prompt cannot run under `--json` and would put deletes out of
reach of the callers that need them most.

- `shorebird apps list|rename|delete|transfer` and
  `shorebird channels list|create|delete`, all w/ `--json`
- `channels` sits top-level alongside `releases` and `patches` rather than
  under `apps`
- `apps transfer` checks org membership before the write, so a wrong
  `--org-id` names the organization it could not find
- `apps list` mirrors `shorebird account apps`, putting the display name that
  `apps delete` requires next to the command that needs it
- The `account apps` command class is now `AccountAppsCommand`, freeing the
  name for the new group. `shorebird account apps` is unchanged

## Testing

- Diff coverage is 100% on both packages, 330/330 lines in `shorebird_cli`
  and 34/34 in `shorebird_code_push_client`
- Ran all six commands against a live account in both a TTY and a pipe,
  including every refusal path. `--json` parsed on every invocation


packages/shorebird_cli/lib/src/commands/confirm_name_argument.dart
```diff
@@ -0,0 +1,47 @@
...
+  /// reporting why it may not. [resourceDescription] names the thing being
+  /// deleted, for the human-readable message.
+  int? checkConfirmName({
+    required String expected,
+    required String resourceDescription,
+  }) {
+    final provided = results[confirmNameArgName] as String?;
+    if (provided == expected) return null;
+
+    final hint = 'Re-run with --$confirmNameArgName="$expected".';
```
easymac, line 29: Neat idea here.

easymac (approved):
👍

PR #3907, opened 2026-09-01
fix(e2e): make nightly failures diagnosable, and stop leaking test apps

eseidel:
## Why

The nightly e2e has been failing 2–16 `patch` legs a night at a rotating set of Flutter versions. The job logs blame `adb: device offline`, which is wrong — that's just the last thing the script prints. Every failing leg actually died on an API request, and the API's answer was unparseable, so the CLI printed a bare `An unknown error occurred.` and the script kept going until `adb shell monkey` tripped over the adb daemon it had killed earlier.

Three separate things conspired to hide that.

## 1. The client threw away the status code

`_parseErrorResponse` collapses every non-2xx response whose body isn't a JSON `ErrorResponse` into one indistinguishable string. A rate-limited 429, a load balancer 502, and a malformed 400 all read identically — to a nightly log and to a user.

Now: `An unknown error occurred. (HTTP 429)`, with the whitespace-collapsed, 200-char-capped body as `details`.

## 2. `patch_e2e.sh` failed in the wrong place

Both `shorebird preview` calls feed a `while read` loop through a process substitution, so their exit status is discarded. When preview failed the loop simply ended and the script carried on. Each loop now checks it actually saw its marker and fails there, naming what went wrong.

Also: `adb wait-for-device` after `adb kill-server`, so later adb commands don't race the restarted daemon; and a `timeout` on the final `adb logcat` tail, so a patch that never boots fails in two minutes instead of burning the job's 30-minute budget.

## 3. Both suites leak an app on every run

`shorebird init` mints a new app per leg and nothing deleted it, so the `shorebird.tester` account has accumulated **16,367 apps**. `shorebird release` fetches the account's entire app list up front — that response is now 7.7 MB and takes up to 10s under load. Thirty legs pulling it concurrently is what overloaded the API in the first place. It compounds: ~30 more every night.

- `patch_e2e.sh` deletes its app from an EXIT trap, so a failed leg cleans up too. There's no `shorebird apps delete` command, so it calls the API directly with the token already in the job env.
- The CLI integration test already deleted its app, but as the last statement of the test body — any earlier failure leaked one. Moved to `addTearDown` right after the app is created. The explicit delete stays, so the assertion that deletion works is still exercised.

Backfilling the already-leaked apps and the server-side capacity fix are in shorebirdtech/_shorebird#2891.

## Testing

`shorebird_code_push_client`: 106 tests pass, including three new cases covering a non-JSON error body, a long body, and an empty one. `dart analyze --fatal-warnings lib test` clean on both touched packages.

https://claude.ai/code/session_017sMaEhRtREivtfCHBjySQ5


scripts/patch_e2e.sh
```diff
@@ -16,6 +16,36 @@
 
 FLUTTER_VERSION=$1
 
+# The environment this script exercises. `shorebird release` and `shorebird
+# patch` pick this up from shorebird.yaml below; the cleanup trap calls the API
+# directly and needs it too.
+API_BASE_URL="https://api-dev.shorebird.dev"
```
easymac: Might prefer `SHOREBIRD_HOSTED_URL` here; I don't think it matters

easymac (approved):

scripts/patch_e2e.sh
eseidel: Good catch — it matters a little more than it looks. `SHOREBIRD_HOSTED_URL` is the name the CLI reads directly (`ShorebirdEnv.hostedUri`, which prefers it over `shorebird.yaml`'s `base_url`), and the `patch` job in `e2e.yaml` already sets it to `https://api-dev.shorebird.dev`. So `API_BASE_URL` was a second spelling of a value already in the environment, and the two could drift.

Switched to `export SHOREBIRD_HOSTED_URL="${SHOREBIRD_HOSTED_URL:-https://api-dev.shorebird.dev}"` — the workflow's value wins in CI (no behavior change there), the default keeps local runs pointed at dev, and the cleanup `curl` now hits whatever host the CLI is actually talking to.

---

PR #3910, opened 2026-09-06
fix(shorebird_cli): re-login when stored credentials have expired

eseidel:
## Problem

`shorebird login` short-circuits on `auth.isAuthenticated`, which is purely local state — it's true whenever `credentials.json` parses and yields an email, regardless of whether the session behind it is still alive. So a user whose refresh token has expired or been revoked gets:

```
You are already logged in as <me@example.com>.
Run shorebird logout to log out and try again.
```

...even though every other command is failing with `Failed to refresh credentials`. The only way forward is to notice that hint and run `shorebird logout` first.

## Fix

`shorebird login` now verifies the stored credentials before short-circuiting. If they're dead, it says so and logs you in — no second command to run:

```
✓ Checking existing credentials
Your credentials have expired. Logging you in again...

The Shorebird CLI needs your authorization to ...
```

- `Auth.hasValidCredentials()` — API keys and CI tokens report valid (they're validated server-side on every request); stored OAuth credentials are verified by actually refreshing them against the auth service. A successful refresh is persisted, so the check doubles as a refresh. Failure is logged at `detail` level and returns false.
- The provider-branching refresh was extracted out of `AuthenticatedClient._refreshForProvider` into a top-level `_refreshCredentialsForProvider`, so it can be reused without that client's "Failed to refresh credentials" error + `ProcessExit` handling.
- `_clearCredentials` is now public `clearCredentials()` — discards local credentials without the pointless `/api/logout` round trip on a token we already know is dead.

Any refresh failure is treated as "needs re-login", which is exactly what `AuthenticatedClient` already tells the user to do today. That does mean an offline `shorebird login` will start a fresh login rather than reporting the cached session, but a login is not going to work offline either way.

## While in here: the already-logged-in message

Logging in when you already are is a successful, informational outcome, but the message read like a failed instruction. It now states the account and mentions logout only as the way to switch users:

```
You are already logged in as <me@example.com>.
Run shorebird logout to log in as a different user.
```

The `SHOREBIRD_TOKEN` case got split off, because the shared second line was actively wrong there — env-var auth takes precedence over stored credentials, so `shorebird logout` would not change who that machine is authenticated as:

```
You are already authenticated via the SHOREBIRD_TOKEN environment variable.
Unset SHOREBIRD_TOKEN to log in as a different user.
```

## Not doing: pre-selecting the provider on re-login

Considered having the CLI skip the provider picker based on the previous login. Three reasons not to:

1. The CLI doesn't know the provider — for shorebird-issued tokens `iss` is always `https://auth.shorebird.dev`, and the upstream IdP isn't in the JWT. Only legacy Google/Microsoft credentials carry it, and those are on the way out.
2. `web/apps/auth` already does this better: the `last_auth_method` cookie has a one-year `maxAge` specifically so the hint survives a session expiry, and `/login` badges that provider "Last used". It's per-browser (the right scope) and knows about `saml`, which the CLI's `AuthProvider` enum doesn't model.
3. Deep-linking `/auth/google` would break the loopback — its loader just redirects to `/login`, and the `continue` destination lives in a cookie only `/login` sets.

## Test plan

New tests in `auth_test.dart` (no credentials, API key, legacy CI token, malformed JWT, refresh succeeds, refresh fails) and `login_command_test.dart` (expired credentials clear and re-login, plus the reworded messages). Full `shorebird_cli` suite passes (2243 tests); `dart analyze --fatal-warnings` and `dart format` clean.

https://claude.ai/code/session_01GL7TxHn6RWyweYJoT16edB

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/auth.dart
+++ b/packages/shorebird_cli/lib/src/auth/auth.dart
@@ -64,6 +64,35 @@ typedef ObtainCredentialsViaLoopbackLogin =
 typedef OnRefreshCredentials =
     void Function(oauth2.AccessCredentials credentials);
 
+/// Refreshes [credentials] using the flow appropriate for [authProvider].
+///
+/// Throws if the refresh fails, e.g. because the session has expired or been
+/// revoked.
+Future<oauth2.AccessCredentials> _refreshCredentialsForProvider({
+  required AuthProvider authProvider,
+  required oauth2.AccessCredentials credentials,
+  required http.Client httpClient,
+  required Uri authServiceUri,
+  required RefreshCredentials refreshCredentials,
+}) async {
+  // Shorebird uses its own refresh flow; Google and Microsoft use the
+  // standard OAuth refresh. This branching can be removed once the
+  // Google/Microsoft providers are fully removed from the CLI.
+  if (authProvider == AuthProvider.shorebird) {
+    return shorebird_oauth.refreshShorebirdCredentials(
+      credentials,
+      httpClient,
+      authBaseUrl: authServiceUri,
+    );
+  }
+  return refreshCredentials(
+    authProvider.clientId,
+    credentials,
+    httpClient,
+    authEndpoints: authProvider.authEndpoints,
+  );
+}
+
 /// A client that automatically refreshes OAuth 2.0 credentials.
 class AuthenticatedClient extends http.BaseClient {
   /// Creates a new [AuthenticatedClient] with the given [httpClient] and
@@ -156,21 +185,12 @@ class AuthenticatedClient extends http.BaseClient {
     oauth2.AccessCredentials credentials,
   ) async {
     try {
-      // Shorebird uses its own refresh flow; Google and Microsoft use the
-      // standard OAuth refresh. This branching can be removed once the
-      // Google/Microsoft providers are fully removed from the CLI.
-      if (authProvider == AuthProvider.shorebird) {
-        return await shorebird_oauth.refreshShorebirdCredentials(
-          credentials,
-          _baseClient,
-          authBaseUrl: _authServiceUri,
-        );
-      }
-      return await _refreshCredentials(
-        authProvider.clientId,
-        credentials,
-        _baseClient,
-        authEndpoints: authProvider.authEndpoints,
+      return await _refreshCredentialsForProvider(
+        authProvider: authProvider,
+        credentials: credentials,
+        httpClient: _baseClient,
+        authServiceUri: _authServiceUri,
+        refreshCredentials: _refreshCredentials,
       );
     } on Exception catch (e, s) {
       logger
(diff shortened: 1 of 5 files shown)


packages/shorebird_cli/lib/src/auth/auth.dart
```diff
@@ -268,6 +291,38 @@ class Auth {
...
+        credentials: credentials,
+        httpClient: _httpClient,
+        authServiceUri: _authServiceUri,
+        refreshCredentials: _refreshCredentials,
+      );
+      _credentials = refreshed;
+      _email = refreshed.email ?? _email;
+      _flushCredentials(refreshed);
+      return true;
+    } on Exception catch (error) {
```
easymac, line 320: This catches a transient network error (e.g., the user submits `shorebird login` while not connected to wifi,) logs the user out, and prints `Stored credentials are no longer valid`. Consider returning false only for 400/401, and letting 5xx surface with a "try again" message.

packages/shorebird_cli/lib/src/commands/login_command.dart
```diff
@@ -17,17 +17,36 @@ class LoginCommand extends ShorebirdCommand {
...
-              ? 'You are already logged in as <$emailDisplay>.'
-              : 'You are already authenticated via API key.',
-        )
-        ..info(
-          'Run ${lightCyan.wrap('shorebird logout')} to log out and try again.',
-        );
-      return ExitCode.success.code;
+      final progress = logger.progress('Checking existing credentials');
+      final hasValidCredentials = await auth.hasValidCredentials();
+      progress.complete();
```
easymac, line 22: This will print "✅ Checking existing credentials" regardless of whether it succeeds, and then immediately print "Your credentials have expired". Acceptable, but consider `progress.fail()` if `hasValidCrednetials` fails.

easymac (approved):
👍 LGTM. Left two UX nits

packages/shorebird_cli/lib/src/auth/auth.dart
eseidel: Right — and worse than printing the wrong line: the command goes on to `auth.clearCredentials()`, so a dropped wifi connection discarded working credentials. Fixed in 60648534.

Only a refusal of the credentials themselves answers `false` now:

- `ShorebirdAuthException` with a 4xx — it now carries the status it saw, rather than only embedding it in the message
- `AccessDeniedException` (the Google/Microsoft path's spelling of a refused grant)
- `ServerRequestFailedException` with a 4xx — an expired or revoked refresh token comes back `400 invalid_grant`

A 5xx or a `SocketException` rethrows, and `shorebird login` catches it, fails the spinner with "Could not reach the Shorebird auth service", and exits `tempFail` with "Check your network connection and try again" — credentials untouched.

One case I kept as `false`: a stored id token too malformed for `Jwt.parse`. That's a fact about the credentials, not about the network, and it already had a test.

Four new tests — false on access-denied and on a 400, rethrow on a 502 and on a `SocketException` — plus one asserting the command keeps the credentials and never calls `login` on a network failure.

---

packages/shorebird_cli/lib/src/commands/login_command.dart
eseidel: Taken — `progress.complete()` only runs on the branch that found valid credentials. The expiry path is now `progress.fail('Your credentials have expired.')` followed by `Logging you in again...`, and the new network-failure path fails it with `Could not reach the Shorebird auth service.` A test asserts `verifyNever(() => progress.complete(any()))` on the expiry path so the checkmark can't come back. 60648534.

---

PR #3911, opened 2026-09-06
fix(shorebird_cli): keep a pubspec comment with its key on init

eseidel:
## Description

`shorebird init` moves the comment block above the first key in a project's `flutter:` section onto the `assets:` block it inserts, so the comment ends up describing something it has nothing to do with.

`addShorebirdYamlToPubspecAssets` asks `yaml_edit` to add the `assets` key, and `yaml_edit` inserts a new key at the **front** of a mapping that holds exactly one entry rather than after it. With two or more entries the same call appends correctly — which is why this went unnoticed. It only misfires on a `flutter` section with a single key, and that is exactly what `flutter create` emits:

```yaml
flutter:
  # The following line ensures that the Material Icons font is
  # included with your application, so that you can use the icons in
  # the material Icons class.
  uses-material-design: true
```

becomes

```yaml
flutter:
  # The following line ensures that the Material Icons font is
  # included with your application, so that you can use the icons in
  # the material Icons class.
  assets:
    - shorebird.yaml
  uses-material-design: true
```

So nearly every first `shorebird init` on a default project rewrites a comment onto the wrong line.

### The fix

Splice the block in after the last entry of the `flutter` section, which is where `yaml_edit` puts it once it has more than one entry to reason about. A comment is not part of the YAML tree, so anything choosing a position from the tree alone is free to land between a comment and its key; choosing the position from the source text is what keeps the two together.

Indentation is read from the document rather than assumed, so a pubspec written with a different indent does not come back reindented in one section only.

After:

```yaml
flutter:
  # The following line ensures that the Material Icons font is
  # included with your application, so that you can use the icons in
  # the material Icons class.
  uses-material-design: true
  assets:
    - shorebird.yaml
```

### Tests

The existing `non-empty flutter` test had pinned the old placement, so its expectation is corrected here. A new regression test covers the `flutter create` shape with its comments, including the trailing "To add assets to your application" block.

Verified against a single key with and without a blank line after `flutter:`, two or more keys, a nested block as the last entry, an existing `assets` list, and a file with no trailing newline.

`pubspec_editor` (7), `init_command` and `shorebird_yaml_asset_validator` (52) all pass.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

https://claude.ai/code/session_01ChE18sC3pmpxxvoZxt3jAT

### DIFF
--- a/packages/shorebird_cli/lib/src/pubspec_editor.dart
+++ b/packages/shorebird_cli/lib/src/pubspec_editor.dart
@@ -77,7 +77,7 @@ class PubspecEditor {
     final lastEntry = flutter.nodes[flutter.keys.last]!;
 
     // A scalar's span stops at the value itself; a nested block's already
-    // includes the newline that closed it. Normalise both to the position just
+    // includes the newline that closed it. Land both on the position just
     // before that newline, or a nested block gains a blank line above the
     // insertion.
     var end = lastEntry.span.end.offset;
@@ -90,11 +90,10 @@ class PubspecEditor {
     }
 
     // Taken from the document rather than assumed, the way `yaml_edit` takes
-    // it, so a pubspec written with a different indent does not come back
-    // reindented in one section only. The keys of `flutter` sit exactly one
-    // step in, which makes that column the step.
-    final step =
-        (flutter.nodes.keys.first as YamlScalar).span.start.column;
+    // it, so a pubspec written with a different indent keeps that indent
+    // here instead of gaining a section in some other one. The keys of
+    // `flutter` sit exactly one step in, which makes that column the step.
+    final step = (flutter.nodes.keys.first as YamlScalar).span.start.column;
     final key = ' ' * step;
     final item = ' ' * (step * 2);
 
--- a/packages/shorebird_cli/lib/src/pubspec_editor.dart
+++ b/packages/shorebird_cli/lib/src/pubspec_editor.dart
@@ -38,17 +38,67 @@ class PubspecEditor {
       );
     } else {
       if (!(yaml['flutter'] as Map).containsKey('assets')) {
-        editor.update(['flutter', 'assets'], ['shorebird.yaml']);
-      } else {
-        final assets = (yaml['flutter'] as Map)['assets'] as List;
-        if (!assets.contains('shorebird.yaml')) {
-          editor.update(['flutter', 'assets'], [...assets, 'shorebird.yaml']);
-        }
+        // Not `editor.update`, which chooses where a new key goes and chooses
+        // wrong when the map has exactly one key: it inserts *before* that
+        // key rather than after it, which separates the key from the comment
+        // block above it. That is the shape `flutter create` emits --
+        // `flutter:` holding only `uses-material-design: true`, with three
+        // lines of comment above describing it -- so almost every first
+        // `shorebird init` rewrote a comment onto the wrong line. With two or
+        // more keys the same call appends correctly, which is why this went
+        // unnoticed: it only misfires on the default project.
+        pubspecFile.writeAsStringSync(
+          _appendAssetsToFlutterSection(pubspecContents, yaml as YamlMap),
+        );
+        return;
+      }
+
+      final assets = (yaml['flutter'] as Map)['assets'] as List;
+      if (!assets.contains('shorebird.yaml')) {
+        editor.update(['flutter', 'assets'], [...assets, 'shorebird.yaml']);
       }
     }
 
     if (editor.edits.isEmpty) return;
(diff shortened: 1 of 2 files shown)


packages/shorebird_cli/lib/src/pubspec_editor.dart
```diff
@@ -38,17 +38,66 @@ class PubspecEditor {
...
+    // Taken from the document rather than assumed, the way `yaml_edit` takes
+    // it, so a pubspec written with a different indent keeps that indent
+    // here instead of gaining a section in some other one. The keys of
+    // `flutter` sit exactly one step in, which makes that column the step.
+    final step = (flutter.nodes.keys.first as YamlScalar).span.start.column;
+    final key = ' ' * step;
+    final item = ' ' * (step * 2);
+
+    final block = '\n${key}assets:\n$item- shorebird.yaml';
+    return contents.substring(0, end) + block + contents.substring(end);
```
easymac: Given `flutter: {users-material-design: true}` this will output
```yaml
fluter: {uses-material-design: true}
    assets:
        - shorebird.yaml
```

packages/shorebird_cli/lib/src/pubspec_editor.dart
```diff
@@ -38,17 +38,66 @@ class PubspecEditor {
...
+  /// `flutter` section, which is where `yaml_edit` puts it whenever it has
+  /// more than one entry to reason about.
+  ///
+  /// Placing it by hand rather than by key is what keeps every comment
+  /// attached to the line it documents: a comment is not part of the YAML
+  /// tree, so anything choosing a position from the tree alone is free to land
+  /// between a comment and its key.
+  String _appendAssetsToFlutterSection(String contents, YamlMap yaml) {
+    final flutter = yaml.nodes['flutter']! as YamlMap;
+    final lastEntry = flutter.nodes[flutter.keys.last]!;
```
easymac: `flutter: {}` will reach `flutter.keys.last` and brash with `Bad state: No element`.

This and the other comment in the method would fix by using the old `editor.update` when the section is a block map

packages/shorebird_cli/lib/src/pubspec_editor.dart
```diff
@@ -38,17 +38,66 @@ class PubspecEditor {
...
+    // before that newline, or a nested block gains a blank line above the
+    // insertion.
+    var end = lastEntry.span.end.offset;
+    if (end > 0 && contents[end - 1] == '\n') {
+      end -= 1;
+    } else {
+      while (end < contents.length && contents[end] != '\n') {
+        end++;
+      }
+    }
```
easymac: This block sweep includes comments, so the `assets` line would appear under the comments.

i.e.,

```yaml
flutter:
    uses-material-design: true
    fonts:
        - {...}

# Some unrelated comment
    assets:
        - shorebird.yaml
```

easymac (approved):
Approving because I don't know the impact of the gaps Claude identified :)

packages/shorebird_cli/lib/src/pubspec_editor.dart
eseidel: Right, and fixed in 8acba9ce — this exact case now has a test. I had caught it in self-review just before your comment landed, so we found it independently.

Your suggested fix is what I went with: the append only runs for a non-empty **block** map, and everything else falls back to `editor.update`. Neither an empty nor a flow map can hit the misplacement the append exists to avoid, so there is no reason for them to go through it.

packages/shorebird_cli/lib/src/pubspec_editor.dart
eseidel: Also right, and fixed by the same guard in 8acba9ce. Worse than the empty-map case, because it is silent — the file still writes, it just no longer parses.

Covered by a test that asserts the result parses and still has both entries, rather than pinning a particular formatting.

packages/shorebird_cli/lib/src/pubspec_editor.dart
eseidel: This one I had missed, and it is worse than your example shows — for a block collection as the last entry, the span runs to the **end of the document**, not just past the next comment. So `assets:` landed below the comment exactly as you drew it.

Fixed in 5a49aa9d by walking down to the deepest scalar and appending after that, since only a scalar span is tight. I considered scanning back over comment lines instead, but that walks over text inside a block scalar that merely looks like a comment; a block scalar is a leaf, so recursing gets that right without a special case. Regression test added with your example.

PR #3912, opened 2026-09-06
feat(shorebird_cli): show plan level in `account whoami`

eseidel:
`shorebird account whoami` reported `Plan: paid` or `Plan: free`, derived from `has_active_subscription`. That is a boolean, so it cannot tell pro from enterprise — and "which plan am I on" is almost always really "can I use this plan-gated capability", like scoped API keys or the `developer` role.

`GET /api/v1/plan` already returns `level`. This reads it.

### Before

```
ID:             42
Email:          user@example.com
Display name:   Example User
Plan:           paid
Overage limit:  10000
```

### After

```
ID:             42
Email:          user@example.com
Display name:   Example User
Plan:           enterprise
Subscription:   active
Overage limit:  10000
```

The paid/free bit moves to its own line rather than being dropped. That is not just for continuity: a subscription whose price carries no `plan_level` metadata resolves to level `free` while still being an active subscription, and seeing `Plan: free` next to `Subscription: active` is how you'd notice.

`Plan` shows `unknown` if the server reports no level.

### Notes

- Only `level` is read, not the whole plan. The plan is a server-local model whose billing fields are deliberately not in the protocol package, so porting it would drag that boundary into the public submodule for one string.
- `--json` keeps `plan` as `paid`/`free` so existing scripts keep working, and adds `plan_level` next to it.

### Testing

New tests cover the client method (success, no `level` in the response, request failure), the wrapper method (success and failure), and the command (plan level in human and `--json` output, `unknown` when null, `Subscription: none` on a free account, and the plan fetch failing).

https://claude.ai/code/session_01JuXTMzbvJ977hMnkZMNfVN

### DIFF
--- a/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
+++ b/packages/shorebird_cli/lib/src/code_push_client_wrapper.dart
@@ -133,6 +133,20 @@ class CodePushClientWrapper {
     return user;
   }
 
+  /// Fetches the plan level for the current user, e.g. `free`, `pro`,
+  /// `business` or `enterprise`. Null when the server does not report one.
+  Future<String?> getPlanLevel() async {
+    final progress = logger.progress('Fetching plan');
+    final String? level;
+    try {
+      level = await codePushClient.getPlanLevel();
+      progress.complete();
+    } catch (error) {
+      _handleErrorAndExit(error, progress: progress);
+    }
+    return level;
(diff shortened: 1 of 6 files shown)


easymac (approved):
👍  LGTM -- though this only grabs the user's own plan, so only an org's owner will be learning about their entitlements

PR #3913, opened 2026-09-06
feat(shorebird_cli): manage API keys from `shorebird account api-keys`

eseidel:
Adds `shorebird account api-keys list | create | revoke`, so setting up CI no longer means leaving the terminal for the console.

```
$ shorebird account api-keys create --name "Production CI" --scope release-and-patch --expires-in-days 365

sb_api_…

This is the only time the key will be shown. Store it now — in CI, as the SHOREBIRD_TOKEN environment variable.

$ shorebird account api-keys list
7  Production CI  release-and-patch  2026-01-04  2026-09-06
9  Local scripts  full-access        2026-02-11  never
```

## Blocked on a server change

**Draft** until the auth service accepts a scope by name. `create` sends `{"scope": "release_and_patch"}`; today's server only understands an expanded `permissions` array and silently defaults to full access when it sees neither.

The CLI does not paper over that. It compares the scope the server reports back against the one requested and fails if they differ, telling you to revoke the key it just made. Better a loud failure than quietly handing someone a full-access credential they asked to be narrow. That check is also what makes this safe to land ahead of the deploy — an old server produces an error, not a wrong key.

The alternative was for the CLI to send the expanded permission list itself, which would have put a third copy of those presets in the tree, to be kept in sync with `api_key_permissions.dart` and `api-key-permissions.ts` forever. Sending a name and letting the server expand it leaves the CLI with no copy at all.

## Session-only, deliberately

None of these commands work in CI, and that is the intent rather than a gap.

The auth service authenticates key management by looking the bearer token up in `session_`. An API key lives in `api_key_` and a legacy CI token is a third-party JWT, so neither hashes to anything in that table — both already get a 401. A leaked CI credential able to mint fresh credentials would defeat revocation entirely: you would revoke the key you knew about while the attacker used the one they had already made.

`_sendApiKeyRequest` therefore reads the stored refresh token directly instead of going through `client`, and raises `ApiKeySessionRequiredException` when there is none, so CI gets

> API key management requires an interactive login. Run `shorebird login` on a machine with a browser.

rather than an unexplained 401. `gh` and `fly` land in the same place; key management is something a person does at a keyboard.

## Notes for review

- `--scope` is **mandatory**. The console defaults to Full access, which is how most existing keys ended up far wider than their job needs. Making the CLI ask seemed better than inheriting that default.
- A scope this CLI doesn't recognize displays as `unknown` rather than being guessed at, so a preset added server-side later degrades to a cosmetic gap instead of a wrong answer.
- `revoke` takes `--id` from `list` and has no confirmation prompt. Worth deciding whether it should have one; it is immediate and irreversible.
- 26 new tests, `dart analyze` clean, full `test/src/auth` and `test/src/commands/account` suites pass (147).

https://claude.ai/code/session_01JuXTMzbvJ977hMnkZMNfVN

### DIFF
--- a/packages/shorebird_cli/lib/src/auth/api_key.dart
+++ b/packages/shorebird_cli/lib/src/auth/api_key.dart
@@ -1,36 +1,16 @@
 import 'package:equatable/equatable.dart';
+import 'package:shorebird_code_push_protocol/shorebird_code_push_protocol.dart';
 
-/// The permission preset an API key is minted with.
+/// The `--scope` spelling of an [ApiKeyScope].
 ///
-/// These are the only scopes the auth service accepts. Arbitrary permission
-/// combinations are deliberately not offered — a key is one of these two
-/// shapes so that what a key can do is legible from its scope alone.
-enum ApiKeyScope {
-  /// Everything the minting user can do.
-  fullAccess('full-access', wireName: 'full_access'),
-
-  /// Create and publish releases and patches, and read insights. No deletes,
-  /// no member management, no billing.
-  releaseAndPatch('release-and-patch', wireName: 'release_and_patch');
-
-  const ApiKeyScope(this.flagName, {required this.wireName});
-
+/// Command-line spelling is a CLI concern, so it lives here rather than on
+/// the protocol enum, which carries only the wire value.
+extension ApiKeyScopeFlagName on ApiKeyScope {
   /// The value accepted by `--scope` on the command line.
-  final String flagName;
-
-  /// The value the auth service uses for this scope.
-  final String wireName;
-
-  /// The scope whose [wireName] is [value], or null if none matches.
-  ///
-  /// Returns null for a scope this CLI does not know about, which is what a
-  /// newer server introducing a preset looks like from here.
-  static ApiKeyScope? fromWireName(String value) {
-    for (final scope in ApiKeyScope.values) {
-      if (scope.wireName == value) return scope;
-    }
-    return null;
-  }
+  String get flagName => switch (this) {
+    ApiKeyScope.fullAccess => 'full-access',
+    ApiKeyScope.releaseAndPatch => 'release-and-patch',
+  };
 }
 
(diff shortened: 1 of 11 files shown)


eseidel:
Two updates since the description was written.

## `ApiKeyScope` moved to `shorebird_code_push_protocol`

It was in `shorebird_cli`, which was the wrong layer. The auth service mints a key against one of these presets and the CodePush server is what enforces the resulting permissions on every request — so the vocabulary spans server and client, which is what the protocol package is for.

It lands in `lib/extensions/` rather than `lib/src/`, since `lib/src/` is generated from the CodePush OpenAPI spec and API keys are minted by the auth service, which that spec does not cover. `AuthProvider` sits there for the same reason. Only the wire value moved; `--scope`'s hyphenated spelling is a command-line concern and stays here as an extension.

The HTTP calls stay in `Auth` deliberately. `shorebird_code_push_client` has no auth-service surface at all, while `shorebird_oauth.dart` and `Auth._revokeSession` already talk to that service from here — so this follows the existing seam rather than cutting a new one.

## Output now composes

`create` wrote the secret and its guidance to the same stream, so piping it picked up the prose. The key is content and the rest is diagnostic, so they are split: the secret alone on stdout, guidance on stderr.

```
shorebird account api-keys create --name CI --scope release-and-patch | pbcopy
shorebird account api-keys create --name CI --scope release-and-patch --json | jq -r .data.api_key
```

Same shape as `gh auth token` and `kubectl create token`, and the split the logger already applies to progress ("progress is diagnostic, never content"). `list` follows it too — "No API keys." moved to stderr so an empty account writes nothing a pipe can count.

## Known: the plan gate is not enforced server-side

Scoped keys are pro+ via `PlanFeature.apiKeyPermissions`, but that check lives in the console's form validation, not in the API. This PR does not widen the gap; it does make it reachable from the default client. Filed as #2983 rather than fixed here, because the gate arguably points the wrong way and the fix location is an open design question.

PR #3915, opened 2026-09-10
fix(shorebird_cli): ignore asset catalog toolchain versions when diffing

eseidel:
Upgrading Xcode between a release and its patch reports every `Assets.car` in the archive as changed, even when no asset was touched.

`assetutil --info` emits the versions of the tools that compiled the catalog in the same JSON we hash to compare two builds. Xcode 26.3 moves `CoreUIVersion` from 972 to 973 and rewrites `Authoring Tool` and `MainVersion` to match, so the hash changes and the differ reports asset changes across the app.

That warning is the only signal a developer gets that patched Dart may reference assets the installed app does not have, and `--allow-asset-diffs` is all-or-nothing. Firing it on an Xcode upgrade costs the warning its credibility for the case where the change is real.

## Change

Drop `AssetStorageVersion`, `Authoring Tool`, `CoreUIVersion`, `DumpToolVersion` and `MainVersion` before hashing, alongside the `Timestamp` and layered-icon exclusions already there.

`PlatformVersion` and `SchemaVersion` are deliberately kept. They describe the catalog that was asked for rather than the tool that built it, so a change in either is worth reporting.

A re-encode that actually alters a rendition is still reported: `SizeOnDisk`, `Encoding`, `Compression`, both dimensions and `SHA1Digest` remain in the hash. The added test covers that case by changing `SizeOnDisk` under a bumped `CoreUIVersion`.

## Tests

Three added to `sanitizeCarJson`:
- an unchanged catalog hashes equal across an Xcode toolchain bump
- a rendition edit made under a new toolchain is still detected
- `PlatformVersion` and `SchemaVersion` changes are still detected

Full file passes (32 tests). The single remaining analyzer info on line 1 is pre-existing on `main`.

https://claude.ai/code/session_019EGM1ZAgHB2ahUiSHq5Hmf

### DIFF
--- a/packages/shorebird_cli/lib/src/archive_analysis/apple_archive_differ.dart
+++ b/packages/shorebird_cli/lib/src/archive_analysis/apple_archive_differ.dart
@@ -169,6 +169,22 @@ class AppleArchiveDiffer extends ArchiveDiffer {
     return _hash(sanitizeCarJson(jsonFile.readAsStringSync()).codeUnits);
   }
 
+  /// Keys in the `assetutil --info` header block that describe the toolchain
+  /// that compiled the asset catalog rather than its contents. Xcode 26.3
+  /// bumped `CoreUIVersion` from 972 to 973, for example, which changes every
+  /// one of these strings without changing a single asset.
+  ///
+  /// `PlatformVersion` and `SchemaVersion` are deliberately not here. They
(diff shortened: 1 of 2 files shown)


nickshorebird (approved):
LGTM

PR #3916, opened 2026-09-11
feat(shorebird_cli): --json envelopes for release and patch

eseidel:
Twenty commands answer `--json` with a success envelope. The two that do the job did not: `shorebird release` and `shorebird patch` printed their human progress lines and ended, so a caller that asked for JSON got no release id, no patch number, and nothing to branch on short of scraping log text.

## What changed

**`shorebird release --json`**

```json
{"status":"success","data":{"app_id":"…","platforms":["android","ios"],"releases":[{"id":123,"app_id":"…","version":"1.2.3+4","flutter_revision":"…","flutter_version":"3.44.0","platform_statuses":{"android":"active","ios":"active"},…}]},"meta":{"version":"1.6.120","command":"release"}}
```

Releases are collected by id, since one Release carries several platforms. Each is refetched after finalize so its status reads `active` rather than the `draft` the command held in hand; that fetch only happens when there's an envelope to fill.

**`shorebird patch --json`**

```json
{"status":"success","data":{"app_id":"…","track":"stable","patches":[{"platform":"android","release_id":123,"release_version":"1.2.3+4","patch_id":456,"patch_number":7,"link_percentage":97.5}]},"meta":{…}}
```

**`--dry-run --json`** on either emits what would have been published (`dry_run: true` plus the version, platform, and for patch the link percentage) instead of exiting with success and nothing on stdout.

**stdout is the envelope and nothing else.** In JSON mode the runner now redirects everything else bound for stdout (human log lines, `logger.info`/`success`, subprocess output) to stderr for the length of the command, and binds the envelope's sink to the real stdout from outside that redirect. Before this, `--json` on a long command meant a JSON line at the bottom of a page of prose, which is not JSON to anything reading it. Commands that already returned before logging are unaffected.

Smoke test on a real invocation: stdout is exactly one JSON line, the human message is on stderr.

## Plumbing

- `CodePushClientWrapper.publishPatch` and `Patcher.uploadPatchArtifacts` return the `Patch` they create rather than `void`.
- `createRelease` returns the `Release`; `createPatch` returns a `PublishedPatch` (platform, release, patch, link percentage).
- `isJsonMode` reads `false` where nothing has set it, so code that consults it can run under a scope that never mentions it. `JsonResult.write` falls back to the zone's stdout when no sink is bound, which is what tests capture.

## Testing

- `release`: envelope shape with the refetched release; only the envelope reaches stdout; dry-run envelope.
- `patch`: envelope shape; dry-run envelope.
- Runner: a command that writes a human line to stdout then an envelope produces stdout = envelope only, stderr = the human line, under `--json`; stdout untouched without it.
- Full `shorebird_cli` suite passes (2251). Changed lib lines are covered.

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/release/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_command.dart
@@ -426,11 +426,23 @@ of the iOS app that is using this module. (aar and ios-framework only)''',
         // this platform still reads draft. A caller reading the envelope
         // should see the release as it is now, which costs one fetch and is
         // only paid when there is an envelope to fill.
+        //
+        // Best-effort, because by this point the release is published: a
+        // refresh that fails must not turn a run that succeeded into a
+        // failed one. getRelease would do exactly that -- it exits the
+        // process, and does it with a message about publishing patches,
+        // which is not what the user just ran. Falling back to the release
+        // in hand costs the envelope a stale status, no more.
         if (!isJsonMode) return release;
-        return codePushClientWrapper.getRelease(
-          appId: appId,
-          releaseVersion: release.version,
-        );
+        try {
+          final refreshed = await codePushClientWrapper.maybeGetRelease(
+            appId: appId,
+            releaseVersion: release.version,
+          );
+          return refreshed ?? release;
+        } on Exception {
+          return release;
+        }
       },
       values: {shorebirdEnvRef.overrideWith(() => releaseFlutterShorebirdEnv)},
     );
--- a/packages/shorebird_cli/lib/src/commands/release/release_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/release/release_command.dart
@@ -13,6 +13,7 @@ import 'package:shorebird_cli/src/common_arguments.dart';
 import 'package:shorebird_cli/src/config/config.dart';
 import 'package:shorebird_cli/src/extensions/arg_results.dart';
 import 'package:shorebird_cli/src/extensions/string.dart';
+import 'package:shorebird_cli/src/json_output.dart';
 import 'package:shorebird_cli/src/logging/logging.dart';
 import 'package:shorebird_cli/src/metadata/metadata.dart';
 import 'package:shorebird_cli/src/platform.dart';
@@ -193,12 +194,23 @@ of the iOS app that is using this module. (aar and ios-framework only)''',
       return ExitCode.usage.code;
     }
 
-    final releaserFutures = results.releaseTypes
-        .map(_resolveReleaser)
-        .map(createRelease);
+    // One Release can carry several platforms, so releases are collected by
+    // id: each platform's pass refetches the release, and the last fetch is
+    // the one whose platform statuses are complete.
+    final releases = <int, Release>{};
+    for (final releaseType in results.releaseTypes) {
+      final release = await createRelease(_resolveReleaser(releaseType));
+      releases[release.id] = release;
+    }
 
-    for (final future in releaserFutures) {
-      await future;
+    if (isJsonMode) {
+      emitJsonSuccess({
+        'app_id': appId,
+        'platforms': [
+          for (final type in results.releaseTypes) type.releasePlatform.name,
+        ],
+        'releases': [for (final release in releases.values) release.toJson()],
+      });
     }
 
     return ExitCode.success.code;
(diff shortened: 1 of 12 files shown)


packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
```diff
@@ -148,6 +149,13 @@ class ShorebirdCliCommandRunner extends CompletionCommandRunner<int> {
       final shorebirdArtifacts = engineConfig.localEngineSrcPath != null
           ? const ShorebirdLocalEngineArtifacts()
           : const ShorebirdCachedArtifacts();
+      // In JSON mode stdout carries the envelope and nothing else. Both
+      // sinks are bound here, outside the redirect below: the envelope goes
+      // to the real stdout, and everything else that would have reached
+      // stdout (human log lines, progress, subprocess output) goes to stderr,
+      // where a caller parsing stdout as JSON will not trip over it.
+      final jsonSink = io.stdout;
+      final humanSink = io.stderr;
```
easymac, line 162: `ShorebirdProcess.stream` starts children with `ProcessStartMode.inheritStdio`, so `IOOverrides` cannot reach it. Reproduced directly; Gradle and Flutter build log into `out.json` ahead of the envelope. The comment on 155 says they go to stderr instead.

packages/shorebird_cli/lib/src/commands/release/release_command.dart
```diff
@@ -396,6 +421,16 @@ of the iOS app that is using this module. (aar and ios-framework only)''',
...
+
+        // The Release in hand predates finalizeRelease, so its status for
+        // this platform still reads draft. A caller reading the envelope
+        // should see the release as it is now, which costs one fetch and is
+        // only paid when there is an envelope to fill.
+        if (!isJsonMode) return release;
+        return codePushClientWrapper.getRelease(
+          appId: appId,
+          releaseVersion: release.version,
+        );
```
easymac: If this fetch fails, the whole release is reported as failed even if it succeeded.

I believe it also prints "Patches can only be published for existing releases. Please create a release using `shorebird release`" even though the user is invoking `shorebird release`.

easymac (approved):
👍 LGTM. Will be a strong improvement. Approving with two changes; not certain whether the stderr change is a blocker.

packages/shorebird_cli/lib/src/shorebird_cli_command_runner.dart
eseidel: You're right and the comment was wrong. `inheritStdio` hands the child our real fd 1 — the one place `IOOverrides` cannot follow — so gradle and `flutter build` wrote straight into the stream the caller parses as JSON. Thanks for reproducing it rather than just flagging the smell.

Fixed at the source rather than in the comment. `ShorebirdProcess.stream` now picks its mode:

```dart
mode: isJsonMode
    ? ProcessStartMode.normal
    : ProcessStartMode.inheritStdio,
```

and in JSON mode forwards both of the child's streams to stderr itself. The reason `inheritStdio` exists — the child sees a terminal and renders its spinner and ANSI the way users expect — doesn't apply to a caller reading stdout as JSON, so the tradeoff the class doc describes goes the other way here. (`add` rather than `addStream`, since two concurrent `addStream` calls on one sink is a `StateError`.) Bytes, not lines, so a partial UTF-8 sequence can't throw.

Two tests: one that `ProcessStartMode.inheritStdio` is never used in JSON mode, one that both of the child's streams land on stderr. Mutation-checked — putting `inheritStdio` back turns both red.

The comment on 152 is updated to name the second path rather than imply `IOOverrides` covers it. 2fe8f81a.

---

packages/shorebird_cli/lib/src/commands/release/release_command.dart
eseidel: Both true. `getRelease` → `maybeGetRelease` → `getReleases` → `_handleErrorAndExit`, so any failure exits the process; and when the lookup merely comes back empty it prints "Patches can only be published for existing releases. Please create a release using `shorebird release`" — during `shorebird release`.

The release is already published by the time this runs, so a refetch that fails must not turn a successful run into a failed one. Made it best-effort in 2fe8f81a:

```dart
if (!isJsonMode) return release;
try {
  final refreshed = await codePushClientWrapper.maybeGetRelease(
    appId: appId,
    releaseVersion: release.version,
  );
  return refreshed ?? release;
} on Exception {
  return release;
}
```

The cost of falling back is a stale `platform_statuses` in the envelope — which is exactly what the caller got before this refetch existed — against reporting a successful release as a failure, so it's not close.

New test asserts a throwing refetch still exits 0 and emits the release in hand. Mutation-checked: dropping the try/catch turns it red.

---

Issue #3919, opened 2026-09-11
shorebird_cli: every error names the fix


eseidel:
The CLI has 163 \`logger.err\` call sites and 107 \`throw ProcessExit\` sites. Some errors say what to do next (\`It looks like you have an existing android release for version X. Please bump your version number and try again.\`); many state only what went wrong (\`Failed to determine release version from Info.plist: <error>\`, \`Obfuscation was enabled but the obfuscation map was not found\`, \`Patch 3 is already in channel stable\`).

#3918 is the pattern: \`Invalid argument(s): Invalid platform: rollback\` became

\`\`\`
Invalid platform: "rollback".
Valid platforms: aar, android, ios, ios-framework, linux, macos, windows
Did you mean "shorebird patches rollback"?
\`\`\`

## Task

Pass over every user-facing error in \`packages/shorebird_cli/lib\` and make each one carry the next step: the flag to pass, the command to run, the value that is valid, or the doc section that explains it. Where the failure is a usage mistake, throw a \`UsageException\` (exit 64, command usage printed, \`usage_error\` in \`--json\`) rather than \`logger.err\` + \`ProcessExit(70)\`, so the generic "re-run with --verbose / file an issue" text stops appearing for things that are not bugs.

Concentrations by file: \`preview_command.dart\` (22), \`init_command.dart\` (12), \`shorebird_validator.dart\` (8), \`extensions/arg_results.dart\` (8), \`patch_command.dart\` (8), \`shorebird_cli_command_runner.dart\` (7), \`ios_patcher.dart\` (7).

Also in scope: the unknown-option error on \`patch\` (\`Could not find an option named "--force"\`) currently only explains the \`--\` pass-through; when the option looks like a safety override it should name \`--allow-native-diffs\` / \`--allow-asset-diffs\`.

## Why

Third-party agent skill files (#3917) wrap the CLI to translate its errors into actions. An error that names its fix removes the need for the wrapper.

https://claude.ai/code/session_01LG6NjeENXsouDyHrt1YPB7

PR #3920, opened 2026-09-11
fix(shorebird_cli): name the next step in every preview error

eseidel:
Part of #3919. Same pattern as #3918: usage mistakes throw `UsageException`; runtime failures keep `logger.err` but say what to do next.

## Before / after

| Situation | Before | After |
|---|---|---|
| `--staging` | `The --staging flag is deprecated...` (exit 64, no usage) | UsageException: `The --staging flag has been removed. Use --track=staging instead.` |
| `--ks` without `--ks-pass` | `You must provide a keystore password.` | UsageException: `--ks requires --ks-pass (e.g. --ks-pass=pass:<password> or --ks-pass=file:<path>).` (same for `--ks-key-alias`, `--ks-pass`/`--ks-key-pass` prefix) |
| No previewable releases | `No previewable releases found for this app` | `No previewable releases found for app <id>.` + `Create one with shorebird release <platform>, or pass --app-id=<app-id> to preview a different app.` |
| `--platform=ios` but only Android is previewable | `No previewable iOS releases found` | `...for app <id>.` + `Previewable releases exist for: android. Pass --platform=<one of those> or omit --platform to choose interactively.` |
| Unknown `--release-version` | `No previewable releases found for version X` | `No previewable release found for version X.` + `Previewable versions: 1.0.0+1, 1.0.1+2. Pass one with --release-version=<version> or omit it to choose interactively.` |
| iOS-only release on Windows | `This release can only be previewed on platforms that support iOS` | `Release X only has iOS artifacts, which cannot be previewed from windows.` + `Platforms previewable from this machine: Android, Linux, macOS, Windows. iOS releases can only be previewed from macOS.` |
| `--platform` names a non-previewable artifact | `The iOS artifact for this release is not previewable.` | same + `iOS releases built with --no-codesign and add-to-app releases (aar, ios-framework) cannot be previewed. Create a new release with shorebird release ios, or pick another with --release-version.` |
| Download / extract fails | `Exception: ...` | `Failed to download the <arch> artifact: <error>` + `Re-run the command to try the download again.` |
| `getReleaseArtifact` fails | progress fail + `Error getting release artifact: Instance of 'ProcessExit'` | progress fail only |

## Behavior changes beyond text

- `getReleaseArtifact` already logs its failure and throws `ProcessExit` (which `implements Exception`), so the `on Exception` wrapper around it printed a second, meaningless error. The wrapper is removed at all five call sites; the `ProcessExit` propagates to the runner as it does elsewhere.
- A failed download used to leave the (empty or partial) artifact directory/file behind, and the next run skipped the download because the path existed. The new `downloadAndExtractIfNeeded` helper (and the Android `.aab` path) delete the partial artifact on failure. Tests assert the preview directory is empty afterwards.
- The four identical download-and-extract blocks (Linux, Windows, macOS, iOS) are one helper; macOS passes its `ditto` extractor.

## Tests

`dart test` in `packages/shorebird_cli`: 2244 pass. Coverage on `preview_command.dart` unchanged from main (only the `description` getter line is uncovered, as before).

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/preview_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/preview_command.dart
@@ -8,6 +8,7 @@ import 'dart:isolate';
 import 'package:archive/archive_io.dart';
 import 'package:collection/collection.dart';
 import 'package:mason_logger/mason_logger.dart';
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 import 'package:shorebird_cli/src/artifact_manager.dart';
 import 'package:shorebird_cli/src/cache.dart';
@@ -134,10 +135,9 @@ This is only applicable when previewing Android releases.''',
     }
 
     if (results.wasParsed('staging')) {
-      logger.err(
-        '''The --staging flag is deprecated and will be removed in a future release. Use --track=staging instead.''',
+      usageException(
+        'The --staging flag has been removed. Use --track=staging instead.',
       );
-      return ExitCode.usage.code;
     }
 
     final shorebirdYaml = shorebirdEnv.getShorebirdYaml();
@@ -202,10 +202,28 @@ This is only applicable when previewing Android releases.''',
     if (platformReleases.isEmpty) {
       if (maybePlatform != null) {
         logger.err(
-          '''No previewable ${maybePlatform.displayName} releases found''',
+          '''No previewable ${maybePlatform.displayName} releases found for app $appId.''',
         );
+        if (sideloadableReleases.isNotEmpty) {
+          final otherPlatforms = sideloadableReleases
+              .expand((r) => r.activePlatforms)
+              .toSet()
+              .map((p) => p.name)
+              .join(', ');
+          logger.info(
+            '''Previewable releases exist for: $otherPlatforms. Pass --platform=<one of those> or omit --platform to choose interactively.''',
+          );
+        } else {
+          logger.info(
+            '''Create one with ${lightCyan.wrap('shorebird release ${maybePlatform.name}')}.''',
+          );
+        }
       } else {
-        logger.err('No previewable releases found for this app');
+        logger
+          ..err('No previewable releases found for app $appId.')
+          ..info(
+            '''Create one with ${lightCyan.wrap('shorebird release <platform>')}, or pass --app-id=<app-id> to preview a different app.''',
+          );
       }
       return ExitCode.usage.code;
     }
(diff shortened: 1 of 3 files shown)


packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -231,9 +254,16 @@ This is only applicable when previewing Android releases.''',
...
+      final supportedString = supportedReleasePlatforms
+          .map((p) => p.displayName)
+          .join(', ');
+      logger
+        ..err(
+          '''Release $releaseVersion only has $activePlatformsString artifacts, which cannot be previewed from ${platform.operatingSystem}.''',
+        )
+        ..info(
+          '''Platforms previewable from this machine: $supportedString. iOS releases can only be previewed from macOS.''',
+        );
```
easymac, line 274: `supportedReleasePlatforms` only filters out iOS when not running on MacOS, meaning this string will include MacOS when invoked on Windows

packages/shorebird_cli/lib/src/commands/preview_command.dart
```diff
@@ -342,21 +372,14 @@ This is only applicable when previewing Android releases.''',
   }) async {
     const platform = ReleasePlatform.linux;
     late Directory appDirectory;
```
easymac: Since we're removing the try/catch, these can be `final` (4 occurrences)

easymac (approved):
👍 LGTM. Left a minor problem and a separate nit

packages/shorebird_cli/lib/src/commands/preview_command.dart
eseidel: Confirmed, and the test I wrote had the wrong list baked into it — it asserted "Android, Linux, macOS, Windows" from a Windows host and I read that as correct.

It's not only the string. `supportedReleasePlatforms` feeds the `availablePlatforms` filter a few lines up, so a macOS-only release passed the guard on Windows and went on to `installAndLaunchMacos` there. Same for Linux.

Rewrote it so each desktop platform answers for its own OS, with Android universal since `adb` talks to the device from any host:

```dart
static bool _isPreviewableHere(ReleasePlatform releasePlatform) =>
    switch (releasePlatform) {
      ReleasePlatform.android => true,
      ReleasePlatform.ios || ReleasePlatform.macos => platform.isMacOS,
      ReleasePlatform.linux => platform.isLinux,
      ReleasePlatform.windows => platform.isWindows,
    };
```

The existing test now expects "Android, Windows", plus a new one asserting a macOS-only release is refused from Windows rather than attempted. eebc8741.

---

packages/shorebird_cli/lib/src/commands/preview_command.dart
eseidel: Done in eebc8741 — all four `late Directory` declarations (three `appDirectory`, one `runnerDirectory`) are gone and the assignment is now the `final` declaration.

---

PR #3921, opened 2026-09-11
fix(shorebird_cli): name the next step in every init error

eseidel:
Part of #3919. Same pattern as #3918 and #3920: usage mistakes throw `UsageException`; runtime failures keep `logger.err` but say what to do next.

## Before / after

| Situation | Before | After |
|---|---|---|
| `--organization-id=abc` | `Invalid organization ID: "abc"` | UsageException: `Invalid --organization-id "abc": expected a number.` + `Available organizations:` list |
| `--organization-id=999` (not a member) | `Organization with ID "999" not found.` + list | UsageException: `You are not a member of an organization with id 999.` + list |
| Multiple orgs, non-interactive | `Multiple organizations found. Use --organization-id to specify one:` + list | UsageException: `You belong to multiple organizations. Pass --organization-id=<id> to choose one.` + list |
| `--display-name` empty / >128 | `App display name must be between 1 and 128 characters.` (exit 64) | UsageException: `--display-name must be between 1 and 128 characters (got 129).` |
| `./gradlew --status` fails | `Unable to initialize gradlew.` | `Unable to initialize gradlew: <error incl. gradle output>` + `Fix the Gradle error above (run ./gradlew --status in the android directory to reproduce it), then re-run "shorebird init".` |
| daemon start fails | `Unable to start gradle daemon.` | `Unable to start the gradle daemon: <error>` + reproduce hint |
| flavor detection fails | `Unable to extract product flavors.\n<gradle output>` | `Unable to detect product flavors.\n<gradle output>` + `Fix the build error above (run ./gradlew app:tasks --all ...), then re-run "shorebird init".` |
| shorebird.yaml `app_id` no longer exists (adding flavors) | `Could not find app with id...` then `Failed to get existing app info: Instance of 'ProcessExit'` | `Could not find app with id...` + `Fix the app_id in shorebird.yaml, or run shorebird init --force to create a new app.` |
| createApp throws | `<error>` | `Failed to create the app: <error>` + `Re-run "shorebird init" once the error above is resolved. If it persists, contact us...` |
| pubspec.yaml unparseable | `Error parsing "pubspec.yaml": <error>` | same + `Fix the YAML error above and re-run "shorebird init".` |

## Behavior changes beyond text

- The `--display-name` length check sat inside the `try { ... } on Exception` around `createApp`, so a `UsageException` thrown there would have been swallowed. It now runs before the try.
- `getApp` throws `ProcessExit` (which `implements Exception`) after logging; init caught it as a generic Exception and printed `Instance of 'ProcessExit'`. It now catches `ProcessExit` specifically, adds the next step, and rethrows.
- `gradlew.isDaemonAvailable` included no output in its exception; it now includes gradle's stdout/stderr so the init error names the real failure.

## Tests

`dart test` in `packages/shorebird_cli`: 2244 pass. `init_command.dart` coverage unchanged from main (only the `description` getter line is uncovered, as before). `mocks.dart` gains the same `usageRunner()` helper as #3918/#3920 (identical hunk; merges clean).

### DIFF
--- a/packages/shorebird_cli/lib/src/commands/init_command.dart
+++ b/packages/shorebird_cli/lib/src/commands/init_command.dart
@@ -14,6 +14,7 @@ import 'package:shorebird_cli/src/shorebird_command.dart';
 import 'package:shorebird_cli/src/shorebird_documentation.dart';
 import 'package:shorebird_cli/src/shorebird_env.dart';
 import 'package:shorebird_cli/src/shorebird_validator.dart';
+import 'package:shorebird_cli/src/third_party/flutter_tools/lib/flutter_tools.dart';
 import 'package:shorebird_code_push_client/shorebird_code_push_client.dart';
 import 'package:yaml_edit/yaml_edit.dart';
 
@@ -68,7 +69,9 @@ Please make sure you are running "shorebird init" from within your Flutter proje
         return ExitCode.noInput.code;
       }
     } on Exception catch (error) {
-      logger.err('Error parsing "pubspec.yaml": $error');
+      logger
+        ..err('Error parsing "pubspec.yaml": $error')
+        ..info('Fix the YAML error above and re-run "shorebird init".');
       return ExitCode.software.code;
     }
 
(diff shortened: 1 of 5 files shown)


nickshorebird (approved):
LGTM - Thanks for this

packages/shorebird_cli/lib/src/commands/init_command.dart
```diff
@@ -204,9 +221,13 @@ Please make sure you are running "shorebird init" from within your Flutter proje
...
-      } on Exception catch (e) {
-        updateShorebirdYamlProgress.fail('Failed to get existing app info: $e');
-        return ExitCode.software.code;
+      } on ProcessExit {
+        // getApp has already explained which app is missing.
+        updateShorebirdYamlProgress.fail();
+        logger.info(
+          '''Fix the app_id in shorebird.yaml, or run ${lightCyan.wrap('shorebird init --force')} to create a new app.''',
+        );
+        rethrow;
```
easymac: `getApps` uses `_handleErrorAndExit` for every error -- so a network failure, expired token, or `CodePushUpgradeRequiredException` will all print this line. `maybeGetApp` might be better here

easymac (approved):
👍 LGTM. Left one remark

packages/shorebird_cli/lib/src/commands/init_command.dart
eseidel: You're right, and it's worse than the message being imprecise — `getApp` calls `maybeGetApp` calls `getApps`, so a `CodePushUpgradeRequiredException` reached the same `on ProcessExit` as a genuinely missing app and got told to edit `app_id`.

Switched to `maybeGetApp` in e5c5ca71. `null` is the missing app and keeps both lines (the "could not find app" explanation `getApp` used to print, plus the remedy); anything else exits from the fetch with its own message and no advice about `app_id`.

Two tests now instead of one: the missing-app case asserts the remedy is printed, and a new one asserts `verifyNever` on it when the lookup itself throws. Dropping `getApp` also left the `flutter_tools` import unused — `ProcessExit` was the only thing it provided here — so that's gone too, which `--fatal-warnings` would have caught in CI.

---

PR #3923, opened 2026-09-11
fix(shorebird_cli): signing-flag mistakes are usage errors that name the flag

eseidel:
Part of #3919. Covers the signing-flag checks that lived in `extensions/arg_results.dart` (8 sites) plus the two call sites in `patch` and `release`.

## Before / after

| Situation | Before (err + exit 64, no usage, `--json` → `process_exit`) | After (UsageException: exit 64, usage printed, `--json` → `usage_error`) |
|---|---|---|
| `--public-key-path` + `--public-key-cmd` | `Cannot specify both --public-key-path and --public-key-cmd.` | `Pass either --public-key-path or --public-key-cmd, not both.` |
| `--private-key-path` + `--sign-cmd` | `Cannot specify both --private-key-path and --sign-cmd.` | `Pass either --private-key-path or --sign-cmd, not both.` |
| `--sign-cmd` alone | `--sign-cmd requires a public key (--public-key-path or --public-key-cmd).` | `--sign-cmd requires a public key: add --public-key-path=<path> or --public-key-cmd=<command>.` |
| Only one of the key-pair files | `Both public and private keys must be provided.` | `--public-key-path and --private-key-path must be passed together (missing --private-key-path).` |
| Key file missing | `No file found at /x/private.pem` | `--private-key-path: no file found at /x/private.pem.` |
| `--public-key-cmd` fails / bad output (exit 70, unchanged) | `Failed to run --public-key-cmd: ...` | same + `Run the command in your shell to see why it fails; it must print a PEM-encoded public key to stdout and exit 0.` |

## Design

The checks were extension methods on `ArgResults`, which has no `Command` to call `usageException` on. They are now `ShorebirdCommand.assertSigningArgsValid()` (patch) and `assertPublicKeyArgsValid()` (release), so they raise usage errors the same way every other command does. Only `patch` and `release` ever called them. Logic is unchanged; the five nested `assertAbsentOrValid*` methods collapse into two plus a private file-exists check.

`extensions/file.dart` (`File.assertExists`) had no other callers and is removed; the check moved to where the flag name is known.

## Tests

New `test/src/shorebird_command_test.dart` covers both methods (moved from `arg_results_test.dart`, plus missing-key and missing-file naming). `dart test` in `packages/shorebird_cli`: 2247 pass. `mocks.dart` gains the same `usageRunner()` helper as #3918/#3920/#3921 (identical hunk).


nickshorebird (approved):
LGTM - Thanks for this whole friction set fix.

PR #3930, opened 2026-09-11
ci: run required checks on the merge queue

nickshorebird:
## Summary

Merge queue cannot be turned on today. Branch protection requires the `ci` and `required` contexts, neither workflow triggers on `merge_group`, so a queued PR produces no check runs and gets ejected at the queue timeout. Turning the queue on also changes what a green `ci` means, from something a human reads before clicking merge into the thing that authorizes an automated one, and `ci` had two ways to report success over work that never ran. I passed on scoping the cancellation check to queue runs only, since `!cancelled()` is GitHub's documented alternative to `always()` and closes it on every event instead.

- `merge_group` trigger on both workflows and on both templates `shorebird_ci generate` emits, so a regenerate does not silently drop it
- `ci` fails when a job ends `cancelled`, matching the `required` job it sits beside
- `ci` depends on `changes`, so a detection failure fails the gate instead of skipping every build into a green result
- `ci` guards on `!cancelled()` rather than `always()`, so a push that supersedes a run does not report red on the stale commit
- `dorny/paths-filter` pinned to v4 in the static template, released as `shorebird_ci` 0.2.5 so other repos actually receive it

## Testing

- The `merge_group` path cannot be exercised before the queue is enabled, so the event-specific behavior was verified against upstream sources rather than a run
- `dorny/paths-filter` v4 takes its base from `merge_group.base_sha` and fetches that SHA itself, so the shallow checkout in `main.yaml`'s `changes` job still diffs correctly. v3 has no `merge_group` case, which is why the template pin moved
- `cspell-action` falls back to a full scan on an unsupported event and the VGV semantic-pull-request workflow already guards its step on `github.event_name`, so neither turns red in the queue
- `shorebird_ci verify` passes, 184 package tests pass

## Risk

`ci` can now fail where it previously passed, on a cancelled job or a failed `changes` step. That is the intent, and it means a cancellation that used to slide through now blocks a merge. Rollback is the two workflow files.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -254,7 +254,11 @@ jobs:
         build_redis,
         verify_packages,
       ]
-    if: ${{ always() }}
+    # Not always(): that keeps this job alive through a run cancellation,
+    # so a superseded push would report ci red on the stale commit. A
+    # skipped required check reports success, which is the right answer
+    # for a run nothing is waiting on.
+    if: ${{ !cancelled() }}
 
     runs-on: ubuntu-latest
 
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -246,6 +246,7 @@ jobs:
   ci:
     needs:
       [
+        changes,
         license_check,
         semantic_pull_request,
         build_cross_platform_dart_packages,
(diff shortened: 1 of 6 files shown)


eseidel (approved):
Love it.

PR #3932, opened 2026-09-12
feat(shorebird_cli): support --flutter-version=fvm and =system

### LINKED ISSUE #1385: feat: support fvm

eseidel:
## Description

Users who manage Flutter with [fvm](https://fvm.app) (or who just want Shorebird to follow the `flutter` on their `PATH`) had to look up and repeat the version number on every release. `--flutter-version` now accepts two aliases:

- **`system`** asks the `flutter` on `PATH` which version it is.
- **`fvm`** asks fvm which version the project is pinned to, so we don't have to parse `.fvmrc` ourselves.

The reported version is then resolved to a revision like any other version, so we still build with Shorebird's fork at that version rather than the user's Flutter install, and an unsupported version fails with the same "open an issue" error as an explicitly requested one.

Fixes #1385

### How `fvm` resolves

`fvm api project --path <root>` reads `.fvmrc` and prints it as JSON without touching the Flutter install, so a project pinned to a version number answers immediately and can never trigger a Flutter download.

A project can also pin a channel (`stable`) or a fork ref, and only the Flutter that resolves to knows its version number, so those fall back to `fvm flutter --version` under a progress message naming what we're waiting on (that call is what can install Flutter, silently, for minutes). The same fallback covers fvm versions with no `api` command.

### Relationship to #3238

This revives the approach from #3238 (closed as stale, not rejected) on current `main`. It does **not** carry forward two bugs from that branch:

1. **Returned a version where a revision was required.** `flutterVersionFromFVMFlutter` returned `"3.32.4"` straight out of `resolveTargetFlutterRevision`, whose result feeds `installRevision(revision:)` → `git for-each-ref --contains 3.32.4`, which is not a git object. The tests mocked the call away, so it passed CI while being broken in practice. Here the reported version goes through the existing version→revision lookup instead.
2. **Probed the wrong Flutter.** It called `process.runSync('flutter', ...)` with the default `useVendedFlutter: true`, and `_resolveExecutable` rewrites `flutter` → `shorebirdEnv.flutterBinaryFile.path`. `--flutter-version=system` would have reported Shorebird's own pinned version, making the flag a no-op.

It's also much smaller, because `main` already has a correct `getSystemVersion()` that was dead code. `system` is now nearly free, and `getFvmVersion()` is its sibling.

One change beyond the original scope: `assertArgsAreValid` resolves the arg too, so the `minimumFlutterVersion` check actually applies to `fvm`/`system` instead of silently skipping.

Docs for this are on a matching branch in `shorebirdtech/docs` and should land only after this ships.

## Type of Change

- [ ] 🧹 Code refactor
- [ ] 📝 Documentation
- [ ] 🗑️ Chore
- [ ] 🧪 Tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SQV4ay9pqj2NUBhw72dLW8

### DIFF
--- a/packages/shorebird_cli/lib/src/shorebird_flutter.dart
+++ b/packages/shorebird_cli/lib/src/shorebird_flutter.dart
@@ -33,6 +33,10 @@ class ShorebirdFlutter {
   /// The executable name.
   static const executable = 'flutter';
 
+  /// The fvm executable name, used to ask fvm which Flutter version a project
+  /// is pinned to.
+  static const fvmExecutable = 'fvm';
+
   /// The Shorebird Flutter fork git URL.
   static const String flutterGitUrl =
       'https://github.com/shorebirdtech/flutter.git';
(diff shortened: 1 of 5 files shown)


eseidel:
Felix was historically against this (the long term dependency on fvm's file format), but it's one of our most upvoted issues and seems pretty simple to do/maintain.  Thoughts @nickshorebird @wrpeck @easymac?

eseidel:
Eh. I'm confident this is low-risk for us and good for users. We should just land it. I don't think anything beyond normal review is required.

packages/shorebird_cli/lib/src/shorebird_flutter.dart
```diff
@@ -383,11 +387,55 @@ class ShorebirdFlutter {
...
+      throw ProcessException(
+        fvmExecutable,
+        args,
+        '${result.stderr}',
+        result.exitCode,
+      );
+    }
+
+    return _parseVersion(result.stdout.toString());
+  }
```
easymac, line 423: `fvm flutter --version` can trigger a clone of Flutter and we're producing no output here -- might appear as hanging to the user.

Consider `fvm api project --path <root>` if you're only looking to check the version, or piping the `fvm` output

easymac (approved):
👍 Looks good

packages/shorebird_cli/lib/src/shorebird_flutter.dart
eseidel: Good catch — fixed in d4e4869.

`getFvmVersion` now asks `fvm api project --path <root>` first. That reads `.fvmrc` and prints it as JSON without touching the Flutter install, so a project pinned to a version number (the common case) answers instantly and can never trigger a download. We read `project.pinnedVersion.name`, falling back to `project.config.flutter`.

We can't use it exclusively, though: a project can pin a channel (`stable`) or a fork ref, and only the Flutter that resolves to knows its version number — which is what we need to map to a Shorebird revision. So those still fall back to `fvm flutter --version`, and that fallback now runs under a progress message that says what we're waiting on:

```
Asking fvm which Flutter version "stable" resolves to (fvm may need to install it first)
```

The same fallback covers fvm versions old enough not to have `api`, and anything unparseable coming back from it.

---

Issue #1956, opened 2024-04-22
fix: CFBundleShortVersionString & CFBundleVersion Sync while Build Generation for iOS


tirth-multipl:
**Description**

CFBundleShortVersionString & CFBundleVersion are not sync when generating iOS build with shorebird. These are soft-warnings. ⚠️ 

**Steps To Reproduce**

```bash
shorebird release ios --flavor prod  -- --dart-define=BASE_URL=<url>
```

**Expected Behavior**

CFBundleShortVersionString & CFBundleVersion should sync when generating iOS build with shorebird.

**Warnings**

```error
CFBundleShortVersionString Mismatch. The CFBundleShortVersionString value '1.15.32' of extension 'Runner.app/PlugIns/NotificationService.appex' does not match the CFBundleShortVersionString value '1.15.67' of its containing iOS application 'Runner.app'.

CFBundleVersion Mismatch. The CFBundleVersion value '2142' of extension 'Runner.app/PlugIns/NotificationService.appex' does not match the CFBundleVersion value '250' of its containing iOS application 'Runner.app'.
```

bryanoltman:
If I'm understanding correctly, the issue is that your app extension version and your app version are not the same. Shorebird does not manipulate your app version or the versions of your app extensions. Are you suggesting that it should, or am I misunderstanding the problem?

tirth-multipl:
yeah you're correct but when we do normal build generation w/o shorebird, this issue doesn't happen.

bryanoltman:
When you say "normal build generation", what do you mean exactly? How do you normally perform your builds?

tirth-multipl:
> When you say "normal build generation", what do you mean exactly? How do you normally perform your builds?

Without Shorebird I meant.

```
flutter build ipa --flavor prod --dart-define=BASE_URL=<url>
```

eseidel:
I would be useful to understand which versions the Runner.app and Runner.appx end up with after a normal flutter build vs. a shorebird build.  It's not clear to me which version is correct.  I presume 1.15.67 is the "correct" version and the appx is just lagging behind for some reason?  I also wonder if `flutter` also isn't updating the version of the appx but swallowing the log?

`shorebird` just uses `flutter build` to do the underlying build, so it shouldn't affect this aspect.  Although it's possible that we're showing logs from `flutter build` that `flutter build` is hiding by default?

bryanoltman:
[...]

PR #3933, opened 2026-09-12
fix(shorebird_cli): warn when app extension versions do not match the app

eseidel:
## Description

Refs #1956. This warns about the mismatch. It does not sync the versions, and does not close the issue.

### Why it does not sync them

#2620 removed our generated `ExportOptions.plist` and closed this. That fixed plain `shorebird release ios`, since Flutter's own export options let Xcode harmonize embedded bundle versions. It was reopened in July 2025 by users on fastlane-plugin-shorebird, users supplying their own `ExportOptions.plist`, and users archiving in Xcode by hand.

Those remaining cases are structural. `assertValidExportOptionsPlist` rejects `manageAppVersionAndBuildNumber: true`, and `ios_releaser.dart` tells `--no-codesign` users to uncheck "Manage Version and Build Number". We need Xcode's version management off so the version we record is the version that ships, and that setting is what would otherwise rewrite extension versions.

Syncing them requires one of:

1. Rewriting `Info.plist` inside each `.appex` after export. Extensions are signed individually, so this invalidates their signatures.
2. Setting the extension targets' build settings before the build, which means mutating the user's Xcode project from a release command.
3. @eseidel's option 3 from the thread: teach `flutter build ipa` to set extension versions the way it already sets the app's. Still the right long-term answer, and upstream.

### What this does

After `flutter build ipa`, reads the exported `Runner.app/Info.plist` and each `PlugIns/*.appex/Info.plist`, and warns on any `CFBundleShortVersionString` / `CFBundleVersion` disagreement:

```
These app extensions have versions that do not match the app:

  NotificationService.appex: CFBundleShortVersionString is "1.15.32", the app is "1.15.67"

App Store Connect rejects uploads with mismatched extension versions
(ITMS-90473).

Shorebird builds with Xcode's "Manage Version and Build Number" disabled, so
Xcode does not update extension versions for you. Set each extension target's
MARKETING_VERSION and CURRENT_PROJECT_VERSION to match the app, or to
$(MARKETING_VERSION) and $(CURRENT_PROJECT_VERSION).
```

Today the failure arrives as an Apple email hours or days after a release that appeared to succeed, with nothing pointing back at Shorebird. Five people in the thread each rediscovered the cause over two years. Every reporter's extension (`NotificationService.appex`, `CleverTapNotificationServiceExtension.appex`, `OneSignalNotificationService.appex`) would have produced this warning at build time.

It warns and continues, and never fails the release. A missing or malformed extension plist is skipped.

Scoped to `shorebird release ios`. Not covered: macOS bundles (`Contents/PlugIns`), extensions nested inside a bundled watchOS app, and any ExtensionKit `Extensions/` directory. Worth follow-ups if this approach is wanted.

## Type of Change

## Verification

Installed a Dart SDK in the session container and ran:

- `dart analyze` on the package: 90 issues, identical to the pre-change baseline, none in the new files.
- `dart format --set-exit-if-changed`: clean across all touched files.
- `dart test test/src/platform/apple/app_extension_versions_test.dart`: 12/12. Covers both keys, multiple extensions, non-`.appex` entries in `PlugIns`, missing and malformed app and extension plists, and extensions that do not declare a key.
- `ios_releaser_test.dart` is `testOn: 'mac-os'`. Ran it on Linux with the guard temporarily lifted: 36/36 pass, including the new case. Guard restored; that change is not committed.
- Mutation tests. Removing the `logger.warn` call fails the releaser test. Returning `[]` from the detector fails 3 unit tests. Dropping the `.appex` filter fails 1. Checking only `CFBundleShortVersionString` fails 2.
- `cspell --config cspell.config.yaml`: clean. `appex` and `ITMS` went into the global config rather than inline, since three files need them.
- CI green on macos-latest, ubuntu-latest and windows-latest.

Not tested: a real iOS build. The `.app` layouts in the tests are constructed, since verifying against an actual `.xcarchive` needs macOS and Xcode.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015fnEF9Ch8k5VQnwmBGD1nn

### DIFF
--- a/packages/shorebird_cli/lib/src/platform/apple/app_extension_versions.dart
+++ b/packages/shorebird_cli/lib/src/platform/apple/app_extension_versions.dart
@@ -34,18 +34,16 @@ class AppExtensionVersionMismatch {
 /// the app extensions it embeds.
 ///
 /// Apple rejects uploads whose extensions disagree with their containing app
-/// about `CFBundleShortVersionString` or `CFBundleVersion` (ITMS-90473). Xcode
-/// normally papers over this by rewriting every embedded bundle's version at
-/// export time, but that behavior is driven by
-/// `manageAppVersionAndBuildNumber`, which Shorebird requires to be off so that
-/// the version it records for a release is the version that actually ships.
-/// The consequence is that extension versions are left as their targets built
-/// them, and a mismatch surfaces as an App Store Connect rejection rather than
-/// as a build failure.
+/// about `CFBundleShortVersionString` or `CFBundleVersion` (ITMS-90473).
+///
+/// Xcode rewrites embedded bundle versions at export time when
+/// `manageAppVersionAndBuildNumber` is set. Shorebird requires it off so the
+/// version it records for a release is the version that ships, so extension
+/// versions stay as their targets built them and a mismatch surfaces as an App
+/// Store Connect rejection.
 ///
 /// An extension is skipped if its Info.plist is missing, cannot be parsed, or
-/// does not declare the version keys. This is a diagnostic aid, and a
-/// malformed extension plist is not something to fail or warn a release over.
+/// does not declare the version keys.
 List<AppExtensionVersionMismatch> findAppExtensionVersionMismatches({
   required Directory appDirectory,
 }) {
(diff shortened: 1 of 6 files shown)


packages/shorebird_cli/lib/src/platform/apple/app_extension_versions.dart
```diff
@@ -0,0 +1,130 @@
...
+/// does not declare the version keys.
+List<AppExtensionVersionMismatch> findAppExtensionVersionMismatches({
+  required Directory appDirectory,
+}) {
+  final appPlistFile = File(p.join(appDirectory.path, 'Info.plist'));
+  if (!appPlistFile.existsSync()) return [];
+
+  final Plist appPlist;
+  try {
+    appPlist = Plist(file: appPlistFile);
```
easymac, line 55: Had Claude run this. On iOS, the joined `appDirectory.path, 'Info.plist'` is a binary, not XML, so it throws a `FileSystemException: Failed to decode data using encoding 'utf-8'`.

It may be worth fixing in `Plist` instead of here.

easymac (approved):
👍 Approach looks great; left one note

